Issue #29, opened 2015-03-30
Support magical toJson(), fromJson() methods


swankjesse:
Classes can declare the following methods:
```
 <any access modifier> Object toJson();
 <any access modifier> public static Object fromJson(Object json);
```
If they do, Moshi honors ‘em. The returned objects are then themselves converted to JSON. The idea is that if you’re implementing a library class (say OkHttp’s Request) it can define it’s own JSON form without a compile-time dependency on Moshi. If this works I’ll talk to Tatu (Jackson) and Inder/Joel (gson) to try to make it a universally-supported pattern.

inder123:
Shouldn't the returned object be String?

swankjesse:
This returns an object that can be converted to JSON, not a JSON object itself. Example:

```
class Point {
  double x;
  double y;

  public Object toJson() {
    return Arrays.asList(x, y);
  }
  public static Object fromJson(Object json) {
    List list = (List) json;
    return new Point((Double) list.get(0), (Double) list.get(1));
  }
}
```

inder123:
and what is the advantage of that over returning a String?

swankjesse:
The advantage is in how the data looks in the finished JSON document. Imagine we have this:

```
  {
    "path": [[1,2],[3,4],[3,2],[1,4]]
  }
```

You could marshall this with this code, without registering any type adapters:

```
  class Shape {
    List<Point> path;
  }
```

iNoles:
even 
```java
class Shape {
    Point[] path;
}
```
?

swankjesse:
Yup; that too.

inder123:
I am dense, need more explanation. How is the same not achieved if toJson() returned a String instead.

inder123:
toJson() is meant to customize JSON output to be different from the default representation based on the fields. I presume you meant that Shape class would not need to define toJson() because List<Point> and Point[] will automatically get serialized correctly. Well, that is no different from the default behavior based on fields.

pforhan:
Maybe the names could be a bit clearer?  I don't have a good recommendation, though.

swankjesse:
It’s in, and the final version looks a lot like `@Provider` methods.

```
  static class PointAsListOfIntegersJsonAdapter {
    @ToJson List<Integer> pointToJson(Point point) {
      return Arrays.asList(point.x, point.y);
    }

    @FromJson Point pointFromJson(List<Integer> o) throws Exception {
[...]

PR #32, opened 2015-04-11
EnumJsonAdapter

dandc87:
Enums from Strings

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi;
+
+import java.io.IOException;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Type;
+
+/**
+ * Convert Enum types to JSON string values
+ *
+ * @param <T> Enum class to adapt
+ */
+final class EnumJsonAdapter<T extends Enum<T>> extends JsonAdapter<T> {
+  public static final Factory FACTORY = new Factory() {
+    @Override public JsonAdapter<? extends Enum> create(Type type, AnnotatedElement annotations,
+        Moshi moshi) {
+      Class<?> rawType = Types.getRawType(type);
+      if (Enum.class.isAssignableFrom(rawType)) {
+        //noinspection unchecked
+        Class<? extends Enum> enumType = (Class<? extends Enum>) rawType;
+        return new EnumJsonAdapter<>(enumType.getEnumConstants()).nullSafe();
+      }
+      return null;
+    }
+  };
+
+  private final T[] values;
+
+  private EnumJsonAdapter(T[] values) {
+    this.values = values;
+  }
+
+  @Override
+  public T fromJson(JsonReader reader) throws IOException {
+    String name = reader.nextString();
+    for (T value : values) {
+      if (value.name().equals(name)) {
+        return value;
+      }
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+  public EnumJsonAdapter(Class<T> clazz) {
+    this.clazz = clazz;
+  }
+
+  @Override
+  public T fromJson(JsonReader reader) throws IOException {
+    String name = reader.nextString();
+    try {
+      //noinspection unchecked
+      T[] values = (T[]) clazz.getMethod("values").invoke(null);
```
JakeWharton: You don't need reflection for this. `class.getEnumConstants()` will give you a `T[]`. This should also only be done once, as it's expensive.

dandc87: Good call, I hadn't seen that.

moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
```diff
@@ -0,0 +1,48 @@
+package com.squareup.moshi;
+
+import java.io.IOException;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Type;
+
+/**
+ * Convert Enum types to JSON string values. This uses the enum class's
+ * <pre>toString()</pre> to allow for customization.
```
JakeWharton: Hmm... I'm not sure `toString()` is the method we'd want to use for this (vs. just `name()`). #29 and/or registering a custom `JsonAdapter` would allow for customization.

swankjesse: Yeah, we should use `name()` exclusively, particularly because that's easiest to decode.

dandc87:
Sorry about the rebase; I'm used to doing so internally.

moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
```diff
@@ -0,0 +1,47 @@
+package com.squareup.moshi;
```
swankjesse: Square copyright header please.

moshi/src/test/java/com/squareup/moshi/EnumJsonAdapterTest.java
```diff
@@ -0,0 +1,73 @@
+package com.squareup.moshi;
```
swankjesse: Same.

moshi/src/test/java/com/squareup/moshi/EnumJsonAdapterTest.java
```diff
@@ -0,0 +1,73 @@
...
+    assertThat(adapter.fromJson("null")).isNull();
+    assertThat(adapter.toJson(null)).isEqualTo("null");
+  }
+
+  static class Game {
+    Roshambo player1;
+    Roshambo player2;
+  }
+
+  @Test public void enumInObject() throws Exception {
```
swankjesse: I don't think this test is necessary; we already test composition in other tests.

swankjesse:
Nice.

Please add Square copyright headers, delete the unnecessary tests, and squash the commits. See http://rebaseandsqua.sh for details on that. 

Your CLA is valid. Thanks!

moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public T fromJson(JsonReader reader) throws IOException {
+    String name = reader.nextString();
+    for (T value : values) {
+      if (value.name().equals(name)) {
+        return value;
+      }
+    }
+    return null;
```
swankjesse: I think we want to throw an exception in this case. Silently discarding data is dangerous, right?

moshi/src/main/java/com/squareup/moshi/EnumJsonAdapter.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
```
swankjesse: 2015

moshi/src/test/java/com/squareup/moshi/EnumJsonAdapterTest.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
```
swankjesse: https://jira.corp.squareup.com/browse/FRANKLIN-5036

swankjesse:
Sorry for sitting on this! Code looks great!

PR #33, opened 2015-04-12
Add Okio overloads to JsonAdapter.

JakeWharton:


JakeWharton:
Merged into https://github.com/square/moshi/pull/35 since it depends on it.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -26,12 +28,25 @@
 public abstract class JsonAdapter<T> {
   public abstract T fromJson(JsonReader reader) throws IOException;
 
+  public final T fromJson(Source source) throws IOException {
```
swankjesse: Does this lose data?

```
Buffer b = new Buffer().writeUtf8("{}{}");
adapter.fromJson(b);
adapter.fromJson(b);
```

When we create a buffer, we risk reading more than we use.
.

swankjesse:
LGTM

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
swankjesse: https://github.com/square/moshi/issues/38

PR #35, opened 2015-04-12
Expose Sink as constructor type.

### LINKED ISSUE #38: Confirm we don't buffer more than we read
Might need to change our APIs to let the user provide a BufferedSource, since right now we might consume more data than we should.

https://github.com/square/moshi/pull/33

JakeWharton:
Closes #38.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -19,6 +19,8 @@ import java.io.IOException;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Type;
 import okio.Buffer;
+import okio.BufferedSource;
+import okio.Sink;
 
 /**
  * Converts Java values to JSON, and JSON values to Java.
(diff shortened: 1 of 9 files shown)


moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -180,7 +181,11 @@
   /**
    * Creates a new instance that writes a JSON-encoded stream to {@code sink}.
    */
-  public JsonWriter(BufferedSink sink) {
+  public JsonWriter(Sink sink) {
```
swankjesse: Do we offer the necessary APIs to flush or emit this?

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -26,12 +28,25 @@
...
   public final T fromJson(String string) throws IOException {
     return fromJson(new JsonReader(string));
   }
 
   public abstract void toJson(JsonWriter writer, T value) throws IOException;
 
+  public final void toJson(Sink sink, T value) throws IOException {
+    JsonWriter writer = new JsonWriter(sink);
+    toJson(writer, value);
+    writer.close();
```
swankjesse: Closing is an odd choice. What if you're writing JSON to a file that contains mixed data?

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
JakeWharton: Yes. `JsonWriter` is a `Flushable` which flushes its `BufferedSink` wrapper.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
JakeWharton: Switched to `.flush()`.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
JakeWharton: I can add Javadoc to that effect if you think it's likely to be a point of user confusion / problems.

JakeWharton:
Updated to address #38 with test.

swankjesse:
Dig.

PR #36, opened 2015-04-12
Make token a nested type of the reader.

JakeWharton:
This is the only class which uses these tokens and it does not deserve a top-level type.


swankjesse:
That's better.

Issue #40, opened 2015-04-12
Replace substring with Okio-native substring writes.


JakeWharton:
This is a tracking issue for the slight improvement that can be made whenever Okio 1.4 is released.

`JsonWriter`'s `string` method has two calls to `sink.writeUtf8(string.substring(#, #))` which will be able to be replaced by `sink.writeUtf8(string, #, #)`. This avoids the `String` & backing `char[]` allocation as well as the data copy to only read it back out again for encoding.

PR #43, opened 2015-05-25
Save an allocation by using Okio instead of substring.

swankjesse:


swankjesse:
Closes https://github.com/square/moshi/issues/40

Issue #30, opened 2015-03-30
Enum JsonAdapter


swankjesse:

swankjesse:
Fixed. Thanks @dandc87 !

PR #44, opened 2015-05-25
Enum adapter.

swankjesse:


swankjesse:
Closes https://github.com/square/moshi/issues/30

PR #48, opened 2015-06-01
Support @Nullable in Adapter Methods.

swankjesse:
When the annotation is present, the adapter method is used. When
it isn't, it isn't.


swankjesse:
Absent is support for `@Nullable` on the methods that take a `JsonReader` or `JsonWriter`. These are more awkward because, at least in the `JsonReader` case, there’s no value to put the `@Nullable` on to. So I'm just not going to support it.

PR #50, opened 2015-06-07
Throw JsonDataException consistently.

swankjesse:
We throw IllegalStateException when the application code is
internally inconsistent, such as when it closes and then writes
or similar.

We throw an IOException when the source JSON is structurally
invalid, such as when strings are unquoted in strict mode.

And we throw JsonDataException when the application code is
correct, and the JSON is well-formed, but the application code
and JSON disagree with one another.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -311,8 +311,8 @@ public final class JsonReader implements Closeable {
   }
 
   /**
-   * Consumes the next token from the JSON stream and asserts that it is the
-   * beginning of a new array.
+   * Consumes the next token from the JSON stream and asserts that it is the beginning of a new
+   * array.
    */
   public void beginArray() throws IOException {
     int p = peeked;
@@ -324,7 +324,7 @@ public final class JsonReader implements Closeable {
       pathIndices[stackSize - 1] = 0;
       peeked = PEEKED_NONE;
     } else {
-      throw new IllegalStateException("Expected BEGIN_ARRAY but was " + peek()
+      throw new JsonDataException("Expected BEGIN_ARRAY but was " + peek()
           + " at path " + getPath());
     }
   }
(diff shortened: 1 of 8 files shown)


moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -884,14 +872,19 @@ public double nextDouble() throws IOException {
...
-          + " at path " + getPath());
+      throw new JsonDataException("Expected a double but was " + peek() + " at path " + getPath());
     }
 
     peeked = PEEKED_BUFFERED;
-    double result = Double.parseDouble(peekedString); // don't catch this NumberFormatException.
+    double result;
+    try {
+      result = Double.parseDouble(peekedString);
+    } catch (NumberFormatException e) {
```
JakeWharton: +1. I disliked this behavior in Gson.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -353,12 +353,12 @@ public JsonWriter name(String name) throws IOException {
...
       throw new NullPointerException("name == null");
     }
-    if (deferredName != null) {
-      throw new IllegalStateException();
-    }
     if (stackSize == 0) {
       throw new IllegalStateException("JsonWriter is closed.");
     }
+    if (deferredName != null) {
+      throw new IllegalStateException();
```
JakeWharton: Is this what happens when you call `.name("foo").name("bar")`? We should follow up with a good message if so.

swankjesse: Yep. Good idea.

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -216,16 +216,16 @@
     }
   }
 
-  @Test public void emptyString() {
+  @Test public void emptyString() throws Exception {
```
JakeWharton: Seems not needed.

PR #54, opened 2015-06-16
Initial round of Moshi docs.

swankjesse:


README.md
```diff
@@ -1,14 +1,184 @@
...
+
+String json = jsonAdapter.toJson(blackjackHand);
+System.out.println(json);
+```
+
+### Built-in Type Adapters
+
+Moshi has built-in support for reading and writing Java’s core data types:
+
+ * Primitives (int, float, char...) and their boxed counterparts (Integer, Float, Character, ...).
```
JakeWharton: uber nit: you ellipsize differently in each list

README.md
```diff
@@ -1,14 +1,184 @@
...
+
+Moshi uses the same streaming and binding mechanisms as [Gson][gson]. If you’re a Gson user you’ll
+find Moshi works similarly. If you try Moshi and don’t love it, you can even migrate to Gson without
+much violence!
+
+But the two libraries have a few important differences:
+
+ * **Moshi has fewer built-in type adapters.** For example, you need to configure your own date
+   adapter. Most binding libraries will encode whatever you throw at them. Moshi refuses to serialize
+   platform types (`java.*`, `javax.*, and `android.*`) without a user-provided type adapter. This is
```
JakeWharton: you missed a closing backtick on `javax.*`

README.md
```diff
@@ -1,14 +1,184 @@
...
+   encoded in HTML without additional escaping. Moshi encodes it naturally (as `=`) and assumes that
+   the HTML encoder – if there is one – will do its job.
 
 
 
 Download
 --------
 
-**Moshi is currently under development.** The API is not stable and neither is the feature set. It should not be used.
+**Moshi is currently under development.** The API is not stable and neither is the feature set.
```
JakeWharton: "not stable" -> "mostly stable"
"neither is" -> "so is"

We're a bit more committed I think to make these claims.

examples/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+    <version>0.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>moshi-examples</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.squareup.okio</groupId>
+      <artifactId>okio</artifactId>
+    </dependency>
```
JakeWharton: you don't actually need this dependency. it will come transitively.

JakeWharton: and we definitely don't want to encourage people to explicitly specify it. okhttp, moshi, and the wire's of the world should let it come transitively and for dependency resolution semantics to choose the newest version out of those.

examples/src/main/java/com/squareup/moshi/recipes/Suit.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi.recipes;
+
+public enum Suit {
+  CLUBS, DIAMONDS, HEARTS, SPADES;
+
+  @Override public String toString() {
+    return name().substring(0, 1);
```
JakeWharton: The unicode consortium is frowning on this single-character representation.

JakeWharton:
LGTM

README.md
swankjesse: Fixed.

README.md
swankjesse: Fixed.

README.md
swankjesse: New text is simpler: `**Moshi is under development.** The API is not final.`

examples/pom.xml
swankjesse: Fixed.

examples/src/main/java/com/squareup/moshi/recipes/Suit.java
swankjesse: Was tempted to do ♣ but I thought it would be distracting.

Issue #59, opened 2015-06-17
Better error message for missing half of an adapter


JakeWharton:
Given
```java
class AnInterfaceAdapter {
  @ToJson String to(AnInterface ai) {
    return "an_interface";
  }
}
```
```java
Moshi moshi = new Moshi.Builder()
    .add(new AnInterfaceAdapter())
    .build();
moshi.adapter(AnInterface.class);
```
An error message is thrown saying no adapter could be found for `AnInterface`. This is not strictly true and misleading, the problem is that only a "to" adapter could be found but no "from" adapter. We should report a better error message.

PR #62, opened 2015-06-21
Nicer failure messages on partial adapters.

swankjesse:
Closes https://github.com/square/moshi/issues/59


JakeWharton:
LGTM. Two existing failing tests because this throws differently:
```
Expecting message:
 <"No JsonAdapter for class java.lang.String annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]">
but was:
 <"No @ToJson adapter for class java.lang.String annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]">
```

PR #66, opened 2015-08-03
New APIs to reject unknown values.

swankjesse:


swankjesse:
I'm a little uncertain about where this API is slotted in.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -103,6 +103,29 @@ public final String toJson(T value) {
     };
   }
 
+  /**
+   * Returns a JSON adapter equal to this, but that throws a {@link JsonDataException} when {@link
```
JakeWharton: nit: linkplain

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -214,6 +214,9 @@
   /** True to accept non-spec compliant JSON */
   private boolean lenient = false;
 
+  /** True to throw a {@link JsonDataException} on any attempt to call {@link #skipValue()}. */
+  private boolean skipForbidden = false;
```
JakeWharton: When I first read this, I thought it was a boolean to control skipping something classified as a forbidden value. Not forbidding the act of skipping values.

JakeWharton: Should this be `skipAllowed`? I usually find booleans that assert in the positive direction to be easier to reason about.

Think fast: what does `setSkipForbidden(false)` do?

JakeWharton:
@SimonVT this change is for you!

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -191,6 +190,42 @@
...
+    reader.setSkipForbidden(true);
+    reader.beginObject();
+    assertThat(reader.nextName()).isEqualTo("a");
+    try {
+      reader.skipValue();
+      fail();
+    } catch (JsonDataException expected) {
+      assertThat(expected).hasMessage("Cannot skip unexpected NUMBER at $.a");
+    }
+    // Confirm that the reader is left in a consistent state after the exception.
```
JakeWharton: How considerate!

JakeWharton:
LGTM

moshi/src/main/java/com/squareup/moshi/JsonReader.java
swankjesse: How do you feel about this: `JsonReader.setFailOnUnknown(true)` ?

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
swankjesse: Yep. Though this behavior might be the first to go if it harms performance.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
swankjesse: ... that's what I'm going with.

JakeWharton: :+1:

SimonVT:
@JakeWharton haha, awesome! But now I have to look into using moshi...

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -103,6 +103,30 @@ public final String toJson(T value) {
...
   }
 
+  /**
+   * Returns a JSON adapter equal to this, but that throws a {@link JsonDataException} when
+   * {@linkplain JsonReader#setFailOnUnknown(boolean) unknown values} are encountered. This
+   * constraint applies to both the top-level message handled by this type adapter as well as to
+   * nested messages.
+   */
+  public final JsonAdapter<T> failOnUnknown() {
+    final JsonAdapter<T> delegate = this;
```
inder123: Instead of failing, you should consider using a callback for the programmer to handle it.

swankjesse: Interesting idea. So the parser could log a warning but not crash?

inder123: How do you log without passing an explicit logger to Moshi?

Just allow the programmer to set a callback to receive such events and handle them in some way.

PR #67, opened 2015-08-03
New APIs to reject unknown values.

swankjesse:


swankjesse:
I'm a little uncertain about where this API is slotted in. It's

PR #68, opened 2015-08-03
New @Json annotation to customize a field's name.

swankjesse:


moshi/src/main/java/com/squareup/moshi/Json.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/** Customizes how a field is encoded as JSON. */
+@Target(FIELD)
+@Retention(RUNTIME)
+@Documented
+public @interface Json {
+  String name();
```
JakeWharton: Not `value()`?

moshi/src/test/java/com/squareup/moshi/ClassJsonAdapterTest.java
```diff
@@ -389,6 +405,35 @@ public ExtendsPlatformClassWithProtectedField() {
     assertThat(fromJson.toByteArray()).contains((byte) 5, (byte) 6);
   }
 
+  static class NamedFields {
+    @Json(name = "#") String phoneNumber;
+    @Json(name = "@") String emailAddress;
+    @Json(name = "zip code") String zipCode;
```
JakeWharton: Worth testing a non-"primitive" (i.e., nested object)?

moshi/src/main/java/com/squareup/moshi/Json.java
JakeWharton: It's stupidly confusing, but is it worth forcing everyone to type `name = ` everywhere?

JakeWharton:
LGTM

moshi/src/main/java/com/squareup/moshi/Json.java
swankjesse: I was thinking we might eventually offer other options, like `required = true` or `serializeNull = true` or whatever. When that happens it's weird to have this:

```
  @Json(value = "phone number", required = true);
  String phoneNumber;
```

... since it's not a value.

swankjesse: I don't feel strongly!

JakeWharton: Would have been nice if Java let us choose! I'm fine with `name` based on that.

swankjesse: After a few minutes of deliberation, I think I'll leave this as-is. This is an exotic feature anyway, so it doesn't need to be terse; just obvious.

inder123:
Based on my experience with SerializedName annotation, I think you should allow multiple values to be mapped to a single name. May be even support regular expressions.
This is especially useful for enums where an unknown (or recently introduced) value otherwise results in an exception.

PR #71, opened 2015-08-08
Throw a better message attempting to adapt a platform type.

JakeWharton:


moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -706,6 +707,16 @@
       assertThat(expected).hasMessage("Cannot skip unexpected STRING at $.crust");
     }
   }
+  
+  @Test public void platformTypeThrows() throws IOException {
+    Moshi moshi = new Moshi.Builder().build();
+    try {
+      moshi.adapter(File.class);
```
artem-zinnatullin: Missing `fail()` in the `try` block?

swankjesse:
LGTM

Issue #73, opened 2015-08-20
Convert List<MyModel> to JSON


CorradoDev:
I would like to have the ability to convert something like
List<MyModel> myModelList = new ArrayList();
To JSON.  The easiest way I can find and the GSON kind of way to do it is:
moshi.adapter(Types.newParameterizedType(List.class, MyModel.class);

The problem is the Types package is private.  Am I missing something or can we make this package or method public so I can easily convert a list to JSON.

kedar1607:
@JakeWharton do have solution for this yet? This seems to be pretty common use case when using room with `TypeConverters` for `ArrayList`

PR #74, opened 2015-08-21
Make four methods on Types public.

swankjesse:
Not Moshi's responsibility, but pragmatic.

Closes https://github.com/square/moshi/issues/73


kedar1607:
@JakeWharton do have solution for this yet? This seems to be pretty common use case  when using room with `TypeConverters` for `ArrayList`

PR #75, opened 2015-09-05
Implement UUID adapter

Turbo87:
This PR adds an UUID adapter to the `StandardJsonAdapters` class that converts between the `java.util.UUID` class and JSON strings.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
+++ b/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 
 final class StandardJsonAdapters {
   public static final JsonAdapter.Factory FACTORY = new JsonAdapter.Factory() {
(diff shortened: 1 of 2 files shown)


swankjesse:
We were planning to deliberately _not_ include type adapters for everything in the JDK, and just let applications embed it themselves.  The code to do it is really quite trivial with our adapter methods API:

```
        .add(new Object() {
          @ToJson String toJson(UUID uuid) {
            return uuid.toString();
          }
          @FromJson UUID fromJson(String s) {
            return UUID.fromString(s);
          }
        })
        .build();
```

swankjesse:
Agree/disagree @JakeWharton ?

Turbo87:
@swankjesse Thanks for the quick feedback. To be honest I was a bit tired of implementing the same basic adapter over and over again, and I thought that the `UUID` class was probably basic enough to be integrated into moshi directly. Have you considered publishing additional artifacts like e.g. `moshi-adapter-uuid`?

JakeWharton:
I agree with the policy.

I also sympathize as a consumer. While something like UUID seems obvious,
we'll eventually get to types which aren't clear like Date, File, etc. This
puts us in a precarious position since we have no clear definition of what
gets added and what does not.

swankjesse:
We probably want a _examples_ project like OkHttp, which can include this, plus documentation to explain it!

PR #80, opened 2015-09-16
ObjectJsonAdapter should not fail on null values.

### LINKED ISSUE #79: Parsing fails on NULL field values
This JSON
{
		"id":"2",
		"name":"glpi",
		"firstname":null
}

fails: Expected a value but was NULL at path $.firstname

class ObjectJsonAdapter  in  file StandardJsonAdapters.java
has switch (reader.peek()) {}
there are BEGIN_ARRAY, BEGIN_OBJECT, STRING, NUMBER, BOOLEAN
and noooooo  NULL value  among  cases !!!

serj-lotutovici:
Fixes #79

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/ObjectAdapterTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/ObjectAdapterTest.java
@@ -84,6 +84,17 @@ public final class ObjectAdapterTest {
     assertThat(adapter.fromJson("[0, 1]")).isEqualTo(Arrays.asList(0d, 1d));
   }
 
+  @Test public void fromJsonDoesNotFailOnNullValues() throws Exception {
+    Map<Object, Object> emptyDelivery = new LinkedHashMap<>();
+    emptyDelivery.put("address", null);
+    emptyDelivery.put("items", null);
+
+    Moshi moshi = new Moshi.Builder().build();
+    JsonAdapter<Object> adapter = moshi.adapter(Object.class);
+    assertThat(adapter.fromJson("{\"address\":null, \"items\":null}"))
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/ObjectAdapterTest.java
```diff
@@ -84,6 +85,17 @@
     assertThat(adapter.fromJson("[0, 1]")).isEqualTo(Arrays.asList(0d, 1d));
   }
 
+  @Test public void fromJsonDoesNotFailOnNullValues() throws Exception {
+    Moshi moshi = new Moshi.Builder().build();
+    JsonAdapter<Object> adapter = moshi.adapter(Object.class);
+
+    try {
+      adapter.fromJson("{\"address\":null, \"items\":null}");
```
swankjesse: nit: the try/catch isn't really necessary. Better would be to confirm that the result matches a map with two keys whose values are null

swankjesse:
Nice! Three things:

1. Please complete our contributor's agreement. https://github.com/square/moshi/blob/master/CONTRIBUTING.md
2. Please adjust the test to confirm that the null values are retained.
3. Please use http://rebaseandsqua.sh when you're done that so we get a single commit in our history. Then push the change again.

Thanks!

serj-lotutovici:
> Please adjust the test to confirm that the null values are retained.

Done

> Please complete our contributor's agreement. https://github.com/square/moshi/blob/master/CONTRIBUTING.md

Signed the CLA a few minutes before submitting the pr.

swankjesse:
Awesome!

PR #82, opened 2015-09-22
Add read json array recipe

serj-lotutovici:
Since many people are asking how to parse a json array into a list, I think it's a good idea to have it in recipes.

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/ReadJsonList.java
+++ b/examples/src/main/java/com/squareup/moshi/recipes/ReadJsonList.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


swankjesse:
Brilliant!

examples/src/main/java/com/squareup/moshi/recipes/ReadJsonList.java
```diff
@@ -0,0 +1,53 @@
...
+        + "  },\n"
+        + "  {\n"
+        + "    \"rank\": \"J\",\n"
+        + "    \"suit\": \"SPADES\"\n"
+        + "  }\n"
+        + "]";
+
+    Moshi moshi = new Moshi.Builder().build();
+
+    Type listOfCardsType = Types.newParameterizedType(List.class, Card.class);
```
amilcar-andrade: @serj-lotutovici Thanks!

PR #83, opened 2015-09-22
Cache json adapters in memory

serj-lotutovici:
Implement simple in memory caching for `JsonAdapter` classes. Not sure if this is the right way to do this, and if this is the most efficient solution. Will be glad to rework this if necessary.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -20,7 +20,9 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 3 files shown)


swankjesse:
This implementation is dangerous & likely does the wrong thing on a hash collision. What's your motivation? We recommend caching the adapter in the application layer if you can.

serj-lotutovici:
I agree, probably a more smarter cache would make more sense.

regarding motivation: we have a few large json models that we receive form our api. It's not 5m, but reusing created adapters speeds up the parsing. Call this micro optimising, but stuff like that adds up really quick.

Also, I wouldn't want to cache the adapters on application level, because then I ether have to write similar code on top of moshi or analyse which objects require that, and make sure that their adapters are cached. I don't like the second, because it makes it look hacky and someone will defiantly forget about that in the future.

Thirdly, in my opinion such things should be included in parsing libraries, at least on a basic (semi-primitive) level.

An last, but not least, I saw a TODO in the code base :)

swankjesse:
I made a different attempt at this, inspired by your pull request:
https://github.com/square/moshi/pull/86

serj-lotutovici:
Thank you @swankjesse. Will this be part of 1.0.0 release?

swankjesse:
Yep.

PR #86, opened 2015-09-26
Adapter caching, plus other ergonomic features.

swankjesse:
This adds a toString() to most adapters. The format isn't perfect, but
it should make step-debugging easier.

Define the precedence order of adapter factories.

Forbid registering adapters with annotation literals if those annotations
need values to be provided.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
@@ -34,7 +34,7 @@ final class AdapterMethodsFactory implements JsonAdapter.Factory {
   }
 
   @Override public JsonAdapter<?> create(
-      Type type, Set<? extends Annotation> annotations, final Moshi moshi) {
+      final Type type, final Set<? extends Annotation> annotations, final Moshi moshi) {
     final AdapterMethod toAdapter = get(toAdapters, type, annotations);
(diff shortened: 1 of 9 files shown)


moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
```diff
@@ -87,6 +87,10 @@ public AdapterMethodsFactory(List<AdapterMethod> toAdapters, List<AdapterMethod>
           }
         }
       }
+
+      @Override public String toString() {
+        return "JsonAdapter" + annotations + "(" + type + ")";
```
JakeWharton: What does this look like when rendered? Shouldn't the annotations be inside the parens?

JakeWharton:
LGTM

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
JakeWharton: we use `type + " annotated " + annotations` in other places like exception messages

swankjesse: It's a bit giant: `JsonAdapter[@com.squareup.moshi.JsonQualifiersTest$FooPrefix(), @com.squareup.moshi.JsonQualifiersTest$BazSuffix()](class java.lang.String)`

PR #95, opened 2015-10-03
Doc fix for JsonWriter.serializeNulls

edwardaa:
Is the default value of `JsonWriter.serializeNulls` `false`?


JakeWharton:
Yes it is!

PR #97, opened 2015-10-06
Add example for custom JsonAdapter.

DavidMihola:
Since this came up in one of my projects today, I wrote a `JsonAdapter` that combines two JSON values into one Java variable. And while there is a complete example in the JavaDoc for `JsonReader` I thought it might be good to have a working `JsonAdapter` in the recipes section.


swankjesse:
This is neat, but I think it's a little too much code for a recipe. Is there a simpler example?

examples/src/main/java/com/squareup/moshi/recipes/CustomJsonAdapter.java
```diff
@@ -0,0 +1,136 @@
...
+import java.io.IOException;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public final class CustomJsonAdapter {
+
+    private static final DateFormat FORMAT_DATE_AND_TIME = new SimpleDateFormat("yyyyMMddHH:mm");
+    private static final DateFormat FORMAT_DATE = new SimpleDateFormat("yyyyMMdd");
```
swankjesse: SimpleDateFormat is not thread safe, so this code is broken. Also timezones.

DavidMihola:
Thanks a lot for your reply!

From my point of view the main point was changing the structure from JSON to Java - combining two JSON values to one Java variable.

So, what if I:

1.) remove the `title` and everything related to it
2.) remove `fooToken` and everything related to it
3.) just keep the values as strings and simply concatenate them instead of parsing them to a `Date`?

Should I just make a simpler version and push it for review?

Also, thanks for your remarks with regard to `Date` and `DateFormat` - to be honest I was actually using `LocalDateTime` (which doesn't care about timezones) and `DateTimeFormatter` (which is thread safe) in my own code. But I didn't want to depend on these JSR-310 classes in my example here.

But, of course, sorry for sending in broken code!

DavidMihola:
As a first step I did points 2 and 3 I suggested above - I kept `title` since I believe it makes the example feel more intuitive and illustrates that we are keeping some of the JSON structure while changing other parts.

swankjesse:
Yeah, I like this a lot better! Maybe rename from CustomJsonAdapter to something that expresses this specific situation? Not exactly sure what that is.

I think there's a more concise way to solve this problem with Moshi. Here goes:


```
static class Event {
  ... // as above
}

static class EventJson {
  String title;
  String beginDate;
  String beginTime;
}

static class EventJsonAdapter {
  @ToJson EventJson eventToJson(Event event) {
    EventJson json = new EventJson();
    json.title = event.title;
    json.beginDate = event.beginDateAndTime.substring(0, 8);
    json.beginTime = event.beginDateAndTime.substring(9, 14);
    return json; 
  }
  @FromJson Event eventFromJson(EventJson json) {
    Event event = new Event();
    event.title = json.title;
    event.beginDateAndTime = json.beginDate + json.beginTime;
    return event
  }
}
```
What do you think about that?!

DavidMihola:
I think this is amazing - I have of course used `@FromJson` and `@ToJson` in simpler cases (i. e. parsing a `String` to an Android `Uri`), but I was under the assumption that the `@FromJson` method had to take a String and the `@ToJson` method had to return a String. (Even though your Wiki does not actually say so: " A type adapter is any class that has methods annotated @ToJson and @FromJson.")

So, to make sure I understand this correctly - using your code above I did:

        .add(new EventJsonAdapter())
        .build();
    JsonAdapter<Event> jsonAdapter = moshi.adapter(Event.class);

In the last line I am telling Moshi: "Give me an adapter that produces an `Event` from the JSON I will parse".
Moshi then thinks: "OK, I could try to just parse the JSON to an `Event` directly... But I also know how to produce an `Event` from an `EventJson` (using the `EventJsonAdapter`) - so I'll use that and just try to parse the JSON to an `EventJson` as an intermediate step."

(I also now tried to introduce `EventJson2` (that's identical to `EventJson`) as a further intermediate step and also had an `EventJson2Adaper` that converted between `EventJson` and `EventJson2`. As a result Moshi coverted the JSON -> `EventJson2` -> `EventJson` -> `Event`.)

So, without looking at the code in much detail, I assume that Moshi looks for a chain of `@FromJson` methods and when that chain ends it tries to produce the argument type of the last annotated method from the input JSON - more or less correct?

I had thought about parsing to an intermediate object before but thought that I had to have access to a `Moshi` to create that intermediate object. But your way is so much easier!

So, this kind of makes my initial recipe obsolete - but I think *this* should be documented: That `@FromJson` does not have to be based on a String as input.

swankjesse:
Yes, definitely needs to be shouted from the rooftops! The blackjack example is cute but insufficient!

adriancole:
Oh yes. This indeed needs a megaphone!

DavidMihola:
I changed the example from using a `JsonAdapter<Event>` to using @swankjesse's `EventJsonAdapter`. I also took the liberty of adding the example to the appropriate section in README.md.

English is not my native language so some formulations may need some cleanup.

swankjesse:
Do you mind rewriting the example to put the problem first? Ie. you want to show that your JSON doesn't match your model. Then you introduce a helper class (EventJson) that acts as the bridge.

DavidMihola:
Yeah, sure - I hope it's OK now!

PR #98, opened 2015-10-11
Fix description

ming13:


swankjesse:
This is good! But I'm already working on a new section to describe the `@Json` annotation and its motivation.

ming13:
All right!

Issue #101, opened 2015-10-19
Support for Moshi.Builder(Moshi.Builder) constructor or add method.


serj-lotutovici:
This may sound like a stupid request, and no other (know to me) json library allows such possibilities, but there is a case when this is very handy. I have the following setup:
* One api client that uses Moshi for json parsing.
* The api client is shipped with predefined pojos, some of which require custom adapter factories (set by the api client).
* The api client is used by multiple teams (not only onsite).
* Other teams consume other "private" apis. Which result in -> new pojos and eventually requirement to set new adapters/adapter factories.

The point is that the client could expose in it's builder methods that will just mirror the ones in `Moshi.Builder` (currently 4 of them), but a more cleaner solution, would be just have one method `addAll(builder)` or `addFrom(builder)`.

swankjesse:
What about a mechanism to recover the factories from a `Moshi` instance?

serj-lotutovici:
That was actually my initial idea, but I was afraid that it would sound a little unreasonable. Since `Moshi` also instantiates a `Map` and a `ThreadLocal`, and in my use case this is needed to initialise a bigger `Moshi` object, so some performance geeks would argue that this is redundant memory allocation :smile:.

But it works for me. If you agree on `Moshi.Builder#addFrom(Moshi)` I will come up with a PR later today.

PR #102, opened 2015-10-20
Possibility to create new `Moshi.Builder` from `Moshi` instance.

serj-lotutovici:
Added `Moshi#newBuilder` method, which creates a `Moshi.Builder` with all custom factories.

Discused in #101

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -35,14 +36,14 @@
   private final Map<Object, JsonAdapter<?>> adapterCache = new LinkedHashMap<>();
 
   private Moshi(Builder builder) {
-    List<JsonAdapter.Factory> factories = new ArrayList<>();
+    Set<JsonAdapter.Factory> factories = new LinkedHashSet<>();
```
swankjesse: This won't work, because the order will be wrong for factories like `StandardJsonAdapters.FACTORY`.

swankjesse: Probably best to have a constant like `BUILT_IN_FACTORIES` and not add those ones twice.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -171,6 +172,12 @@ public Builder add(Object adapter) {
       return add(AdapterMethodsFactory.get(adapter));
     }
 
+    /** Add all factories enclosed in the provided moshi instance. */
+    public Builder add(Moshi moshi) {
```
swankjesse: I think we probably want `newBuilder()` instead of this. As-is you can compose a `Moshi` from multiple other `Moshi` instances, and that doesn't make much sense.

swankjesse: ... the drawback of `newBuilder()` is it limits where you can add new factories. Order matters here.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -162,8 +177,8 @@ private Object cacheKey(Type type, Set<? extends Annotation> annotations) {
       });
     }
 
-    public Builder add(JsonAdapter.Factory jsonAdapter) {
-      factories.add(jsonAdapter);
+    public Builder add(JsonAdapter.Factory factory) {
```
serj-lotutovici: Squeezing in a typo fix :)

moshi/src/main/java/com/squareup/moshi/Moshi.java
serj-lotutovici: should this be something mentioned in the java doc of the method?

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -116,6 +123,14 @@ private Moshi(Builder builder) {
         + type + " annotated " + annotations);
   }
 
+  /** Returns a new builder containing all custom factories used by the current instance. */
+  public Moshi.Builder newBuilder() {
```
swankjesse: This is awesome

moshi/src/main/java/com/squareup/moshi/Moshi.java
swankjesse: Yes!

swankjesse:
Awesome. Thankyou!

PR #105, opened 2015-10-25
Hook up osstrich, a tool to publish Javadoc to GitHub pages.

swankjesse:


swankjesse:
I didn't wire up `deploy_snapshot`, mostly ’cause that means we won't publish Javadoc for new releases until some unrelated change after the fact. Also (though minor) it means we’ll `git clone` the `gh-pages` repo more than strictly necessary.

...though I don't feel strongly here!

swankjesse:
(once we figure out the right strategy here, I'll repeat for similar repos)

JakeWharton:
Isn't it true that we already have to wait to publish since it uses the search API which is lagged up to 4 hours? Maybe a service isn't such a bad idea after all.

JakeWharton:
LGTM though

inder123:
Neat! We should set this up for Gson too.

PR #111, opened 2015-12-08
Add EnumMap support to MapJsonAdapter.

serj-lotutovici:
Currently when trying to pars a field like `EnumMap<SomeEnum, Value>` Moshi throws an exception saying that it requires a custom json adapter. Since `EnumMap` is a java type I think the adapters place is in the default adapter list.

This is basically a copy/paste of `MapJsonAdapter` with slight alterations (Though that it won't be good to keep everything inside one class).

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
@@ -18,13 +18,16 @@ package com.squareup.moshi;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.EnumMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 /**
  * Converts maps with string keys to JSON objects.
  *
- * TODO: support maps with other key types and convert to/from strings.
+ * @author Jesse Wilson
  */
 final class MapJsonAdapter<K, V> extends JsonAdapter<Map<K, V>> {
   public static final Factory FACTORY = new Factory() {
@@ -32,16 +35,28 @@ final class MapJsonAdapter<K, V> extends JsonAdapter<Map<K, V>> {
         Type type, Set<? extends Annotation> annotations, Moshi moshi) {
       if (!annotations.isEmpty()) return null;
       Class<?> rawType = Types.getRawType(type);
-      if (rawType != Map.class) return null;
+      if (!SUPPORTED_MAPS.contains(rawType)) return null;
       Type[] keyAndValue = Types.mapKeyAndValueTypes(type, rawType);
-      return new MapJsonAdapter<>(moshi, keyAndValue[0], keyAndValue[1]).nullSafe();
+      return new MapJsonAdapter<>(moshi, rawType, keyAndValue[0], keyAndValue[1]).nullSafe();
     }
   };
 
+  /** List of supported raw map types. */
+  private static final List<Class<?>> SUPPORTED_MAPS = new ArrayList<>(2);
+
+  static {
+    SUPPORTED_MAPS.add(Map.class);
+    SUPPORTED_MAPS.add(EnumMap.class);
+  }
+
+  private final Class<?> rawType;
+  private final Class<?> keyRawType;
   private final JsonAdapter<K> keyAdapter;
   private final JsonAdapter<V> valueAdapter;
 
-  public MapJsonAdapter(Moshi moshi, Type keyType, Type valueType) {
+  public MapJsonAdapter(Moshi moshi, Class<?> rawType, Type keyType, Type valueType) {
+    this.rawType = rawType;
+    this.keyRawType = Types.getRawType(keyType);
     this.keyAdapter = moshi.adapter(keyType);
     this.valueAdapter = moshi.adapter(valueType);
   }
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/EnumMapJsonAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+    for (Map.Entry<K, V> entry : map.entrySet()) {
+      writer.promoteNameToValue();
+      keyAdapter.toJson(writer, entry.getKey());
+      valueAdapter.toJson(writer, entry.getValue());
+    }
+    writer.endObject();
+  }
+
+  @Override public String toString() {
+    return "EnumMapJsonAdapter(" + keyAdapter + "=" + valueAdapter + ")";
```
serj-lotutovici: Named this EnumMapJsonAdapter, to avoid key collision with MapJsonAdapter.

moshi/src/test/java/com/squareup/moshi/EnumMapJsonAdapterTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    jsonAdapter.toJson(jsonWriter, null);
+    assertThat(buffer.readUtf8()).isEqualTo("null");
+
+    JsonReader jsonReader = newReader("null");
+    jsonReader.setLenient(true);
+    assertThat(jsonAdapter.fromJson(jsonReader)).isEqualTo(null);
+  }
+
+  @Test public void orderIsRetained() throws Exception {
```
swankjesse: This test is insincere. Order is only maintained because you are using the declaration order.

swankjesse:
If we want to support this, we should probably just give the existing MapAdapter the ability to build EnumMaps.

serj-lotutovici:
@swankjesse Can do that... I just fought that you wouldn't want to over inflate the `MapJsonAdapter`.

> If we want to support this, we should probably just give the existing MapAdapter the ability to build EnumMaps.

serj-lotutovici:
@swankjesse I'm facing a small issue, can you give me a hint? `EnumMap` forces the key to be of type `<K extends Enum<K>>`. What would be the best approach to create an instance of `EnumMap` not forcing K generic in `MapJsonAdapter` to extend from enum?

swankjesse:
@serj-lotutovici you'll probably need some casting! Not much better options there unfortunately.

moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
```diff
@@ -32,16 +35,28 @@
         Type type, Set<? extends Annotation> annotations, Moshi moshi) {
       if (!annotations.isEmpty()) return null;
       Class<?> rawType = Types.getRawType(type);
-      if (rawType != Map.class) return null;
+      if (!SUPPORTED_MAPS.contains(rawType)) return null;
```
serj-lotutovici: It's ether this or `if (!(rawType == Map.class || rawType == EnumMap.class))`. But this saves some effort if other map types should be supported in the future.

moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
```diff
@@ -18,13 +18,16 @@
...
+import java.util.EnumMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 /**
  * Converts maps with string keys to JSON objects.
  *
- * TODO: support maps with other key types and convert to/from strings.
+ * @author Jesse Wilson
```
swankjesse: Please no

moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
```diff
@@ -60,7 +75,19 @@ public MapJsonAdapter(Moshi moshi, Type keyType, Type valueType) {
   }
 
   @Override public Map<K, V> fromJson(JsonReader reader) throws IOException {
-    LinkedHashTreeMap<K, V> result = new LinkedHashTreeMap<>();
+    Map<K, V> result;
+    // EnumMap maintains the natural order of the declared enum type, and in some cases
+    // may be more efficient. So try to use EnumMap if the key is an enum.
+    if (rawType == EnumMap.class || keyRawType.getSuperclass() == Enum.class) {
```
swankjesse: I don't like the 2nd case. This breaks our ability to preserve iteration order.

moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
```diff
@@ -60,7 +75,19 @@ public MapJsonAdapter(Moshi moshi, Type keyType, Type valueType) {
...
+      // Since the compiler will block the caller of declaring an EnumMap with a non enum key,
+      // we we don't have to worry about type safety.
+      //noinspection unchecked
+      result = new EnumMap(keyRawType);
```
swankjesse: Neat.

swankjesse:
I’m still not convinced that this is worth doing. The benefit of EnumMaps as a declaration type are so limited.

serj-lotutovici:
Honestly me too... In this case I went with the flow, since some of my users requested it. I'm closing this, since I can't find arguments other than the ones you already know... really sorry for waisting your time on this.
> I’m still not convinced that this is worth doing. The benefit of EnumMaps as a declaration type are so limited.

PR #119, opened 2016-01-07
Enable Json annotation for enum values.

### LINKED ISSUE #115: @Json annotation on enum values
Another difference between Gson and Moshi is, that Gson allows its `@SerializedName` annotation on enums:

```
enum CustomEnum {
  @SerializedName("not_a") a,
  b
}

@Test
public void testSerializedName() throws Exception {
  CustomEnum a = gson.getAdapter(CustomEnum.class).fromJson("not_a");
  assertThat(a).isEqualTo(CustomEnum.a);
}

@Test
public void testJson() throws Exception {
  CustomEnum a = moshi.adapter(CustomEnum.class).fromJson("not_a");
  assertThat(a).isEqualTo(CustomEnum.a);
}
```

This will work with Gson. Unfortunately, Moshi does not process its `@Json` annotation on enums.

Is this a desired behaviour is this just a missed edge case?

serj-lotutovici:
Attempt to fix #115. Inspired by Gson's implementation. No regex support at this point though.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
+++ b/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -52,7 +53,7 @@ final class StandardJsonAdapters {
       Class<?> rawType = Types.getRawType(type);
       if (rawType.isEnum()) {
         //noinspection unchecked
-        return enumAdapter((Class<? extends Enum>) rawType).nullSafe();
+        return new EnumJsonAdapter<>((Class<? extends Enum>) rawType).nullSafe();
       }
       return null;
     }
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -212,27 +213,46 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
+      String name = reader.nextString();
+      T constant = nameConstantMap.get(name);
+      if (constant != null) return constant;
+      throw new JsonDataException("Expected one of "
+          + nameConstantMap.keySet() + " but was " + name + " at path "
+          + reader.getPath());
+    }
+
+    @Override public void toJson(JsonWriter writer, T value) throws IOException {
+      for (Map.Entry<String, T> entry : nameConstantMap.entrySet()) {
```
serj-lotutovici: Not sure about this past. Gson keeps two maps <Name to Value> and <Value to Name>. Thought, that this would be a useless waist of memory, although this is slower.

serj-lotutovici:
Ping? @swankjesse do you have a different idea for this?

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -212,27 +213,46 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
+        if (entry.getValue() == value) {
+          writer.value(entry.getKey());
+          break;
```
JakeWharton: Just for sanity, I would change this to `return` and `throw new AssertionError` so future programming bugs (which we never make, of course!) don't cause silent failures.

swankjesse: Yeah, definitely need that.

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
swankjesse: If you wanted to save memory, you could have just a `String[]` indexed by `value.ordinal()`.

swankjesse:
This is great. I’m going to merge as-is and follow-up with my own PR to address some small nitpicks.

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -212,27 +213,46 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
+      this.enumType = enumType;
+      try {
+        T[] constants = enumType.getEnumConstants();
+        for (T constant : constants) {
+          Json annotation = enumType.getField(constant.name()).getAnnotation(Json.class);
+          String name = annotation != null ? annotation.name() : constant.name();
+          nameConstantMap.put(name, constant);
         }
+      } catch (NoSuchFieldException e) {
+        throw new AssertionError("Missing field in " + enumType.getName(), e);
```
serj-lotutovici: btw, fi you going to follow up on this pr... consider changing this to `throw new AssertionError("Missing field in " + enumType.getName());`. The `AssertionError(String, Throwable)` constructor is available on Android only since KitKat.

PR #123, opened 2016-01-18
Use an array indexed by ordinal to go from enum to name.

swankjesse:


moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -215,21 +214,25 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
-    private final Map<String, T> nameConstantMap = new LinkedHashMap<>();
+    private final Map<String, T> nameConstantMap;
+    private final String[] nameStrings;
 
     public EnumJsonAdapter(Class<T> enumType) {
       this.enumType = enumType;
       try {
         T[] constants = enumType.getEnumConstants();
-        for (T constant : constants) {
+        nameConstantMap = new LinkedHashMap<>();
```
adelnizamutdinov: do we need initialCapacity here?

swankjesse: It’s surprisingly hard work. http://google.github.io/guava/releases/snapshot/api/docs/src-html/com/google/common/collect/Maps.html#line.186

JakeWharton: The problem with initialCapacity is that it doesn't respect the loadFactor. So if you set it to 100 for a 100 element enum it will still trigger a resize. So you need to account for the default loadFactor when creating initialCapacity and it just gets crazy.

adelnizamutdinov: good to know, thanks

PR #125, opened 2016-01-21
Break apart Okio-based JSON reader and writer.

JakeWharton:


moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -483,165 +247,17 @@ public JsonWriter value(long value) throws IOException {
...
    * that arbitrary type adapters can use {@link #value(String)} to write a name value.
    */
-  void promoteNameToValue() throws IOException {
-    int context = peek();
-    if (context != NONEMPTY_OBJECT && context != EMPTY_OBJECT) {
-      throw new IllegalStateException("Nesting problem.");
-    }
-    promoteNameToValue = true;
-  }
+  abstract void promoteNameToValue() throws IOException;
```
JakeWharton: This package scoped method (and equivalent in `JsonReader`) prevent these types from being fully re-implementable by consumers. We could switch to having it be public.

swankjesse: I'd rather make the constructors package-private & prevent external subclasses.

JakeWharton: Ok. That also means we can push shared logic down!

JakeWharton: Added!

moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
```diff
@@ -0,0 +1,441 @@
...
+    }
+    sink.flush();
+  }
+
+  /**
+   * Flushes and closes this writer and the underlying {@link Sink}.
+   *
+   * @throws JsonDataException if the JSON document is incomplete.
+   */
+  public void close() throws IOException {
```
swankjesse: Should there be `@Override` on these?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -0,0 +1,1078 @@
...
+final class BufferedSourceJsonReader extends JsonReader {
+  private static final long MIN_INCOMPLETE_INTEGER = Long.MIN_VALUE / 10;
+
+  private static final ByteString SINGLE_QUOTE_OR_SLASH = ByteString.encodeUtf8("'\\");
+  private static final ByteString DOUBLE_QUOTE_OR_SLASH = ByteString.encodeUtf8("\"\\");
+  private static final ByteString UNQUOTED_STRING_TERMINALS
+      = ByteString.encodeUtf8("{}[]:, \n\t\r\f/\\;#=");
+  private static final ByteString LINEFEED_OR_CARRIAGE_RETURN = ByteString.encodeUtf8("\n\r");
+
+  private static final int PEEKED_NONE = 0;
```
swankjesse: `#enumsmatter`

moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
swankjesse: (Trying to remember if this is for Java 5’s benefit, which lacks `@Override` on interfaces)

swankjesse:
LGTM!

moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
JakeWharton: Fixed (for the future):

screen shot 2016-01-21 at 10 16 48 am

PR #129, opened 2016-01-24
Look up a map's key/value adapters using raw types. Add a test for ma…

### LINKED ISSUE #128: Android: failing to find adapter for a 'GenericArrayTypeImpl' type
We have an adapter in place to process `byte[]` as a base64 / base32 string. Everything works fine when a class has a `byte[]` field, but parsing json fails on Android when we have a field of type `Map<String, byte[]>`. The code below works in the standard JVM but fails on Android.

``` java
  class Base32Adapter extends JsonAdapter<byte[]>  {
    @Override
    public byte[] fromJson(JsonReader reader) throws IOException {
      String string = reader.nextString();
      return new BigInteger(string, 32).toByteArray();
    }

    @Override
    public void toJson(JsonWriter writer, byte[] bytes) throws IOException {
      String string = new BigInteger(bytes).toString(32);
      writer.value(string);
    }
  }

  static class FavoriteBytes {
    @Json(name = "Bytes")
    public Map<String, byte[]> keys;
  }

  @Test public void customAdapter() throws Exception {
    String jsonString = "{\"Bytes\":{\"jesse\":\"a\",\"jake\":\"1\"}}";
    Moshi binaryMoshi = new Moshi.Builder().add(byte[].class, new Base32Adapter()).build();
    FavoriteBytes fav = binaryMoshi.adapter(FavoriteBytes.class).fromJson(jsonString);

    assertThat(fav.keys).containsOnlyKeys("jesse", "jake");
    assertThat(fav.keys.get("jesse")).isEqualTo(new byte[] { 0xa });
    assertThat(fav.keys.get("jake")).isEqualTo(new byte[] { 0x1 });
  }
```

On Android, this fails with exception:

```
E/MoshiTest(18268): com.squareup.moshi.JsonDataException: Expected BEGIN_ARRAY but was STRING at path $.Keys.jesse
E/MoshiTest(18268):     at com.squareup.moshi.JsonReader.beginArray(JsonReader.java:346)
E/MoshiTest(18268):     at com.squareup.moshi.ArrayJsonAdapter.fromJson(ArrayJsonAdapter.java:53)
E/MoshiTest(18268):     at com.squareup.moshi.JsonAdapter$1.fromJson(JsonAdapter.java:68)
E/MoshiTest(18268):     at com.squareup.moshi.MapJsonAdapter.fromJson(MapJsonAdapter.java:68)
E/MoshiTest(18268):     at com.squareup.moshi.MapJsonAdapter.fromJson(MapJsonAdapter.java:29)
E/MoshiTest(18268):     at com.squareup.moshi.JsonAdapter$1.fromJson(JsonAdapter.java:68)
E/MoshiTest(18268):     at com.squareup.moshi.ClassJsonAdapter$FieldBinding.read(ClassJsonAdapter.java:183)
E/MoshiTest(18268):     at com.squareup.moshi.ClassJsonAdapter.fromJson(ClassJsonAdapter.java:144)
E/MoshiTest(18268):     at com.squareup.moshi.JsonAdapter$1.fromJson(JsonAdapter.java:68)
E/MoshiTest(18268):     at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:33)
E/MoshiTest(18268):     at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:37)
...
```

bjdodson:
…ps with byte[] values. fixes #128 

Types.canonicalize() didn't work in either JVM or Android, fetching the raw type passes on both.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Types.java
+++ b/moshi/src/main/java/com/squareup/moshi/Types.java
@@ -93,6 +93,30 @@ public final class Types {
     }
   }
 
+  static Type canonicalizePlatformTypes(Type type) {
+    if (type instanceof GenericArrayType) {
+      Type c = ((GenericArrayType) type).getGenericComponentType();
+      if (c.equals(Byte.TYPE)) {
+        return byte[].class;
+      } else if (c.equals(Long.TYPE)) {
+        return long[].class;
+      } else if (c.equals(Integer.TYPE)) {
+        return int[].class;
+      } else if (c.equals(Short.TYPE)) {
+        return short[].class;
+      } else if (c.equals(Character.TYPE)) {
+        return char[].class;
+      } else if (c.equals(Boolean.TYPE)) {
+        return boolean[].class;
+      } else if (c.equals(Float.TYPE)) {
+        return float[].class;
+      } else if (c.equals(Double.TYPE)) {
+        return double[].class;
+      }
+    }
+    return type;
+  }
+
   public static Class<?> getRawType(Type type) {
     if (type instanceof Class<?>) {
       // type is a normal class.
--- a/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
@@ -34,8 +34,7 @@ final class MapJsonAdapter<K, V> extends JsonAdapter<Map<K, V>> {
       Class<?> rawType = Types.getRawType(type);
       if (rawType != Map.class) return null;
       Type[] keyAndValue = Types.mapKeyAndValueTypes(type, rawType);
-      return new MapJsonAdapter<>(moshi, Types.getRawType(keyAndValue[0]),
-              Types.getRawType(keyAndValue[1])).nullSafe();
+      return new MapJsonAdapter<>(moshi, keyAndValue[0], keyAndValue[1]).nullSafe();
     }
   };
 
--- a/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/MapJsonAdapter.java
@@ -34,7 +34,8 @@ final class MapJsonAdapter<K, V> extends JsonAdapter<Map<K, V>> {
       Class<?> rawType = Types.getRawType(type);
       if (rawType != Map.class) return null;
       Type[] keyAndValue = Types.mapKeyAndValueTypes(type, rawType);
-      return new MapJsonAdapter<>(moshi, keyAndValue[0], keyAndValue[1]).nullSafe();
+      return new MapJsonAdapter<>(moshi, Types.getRawType(keyAndValue[0]),
+              Types.getRawType(keyAndValue[1])).nullSafe();
     }
   };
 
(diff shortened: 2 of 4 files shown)


swankjesse:
Sadly, getRawType isn't quite right. That’ll break for say, `Map<String, List<Integer>>`. We probably should just convert `GenericArrayType` to the corresponding `Class` if the generic array’s component type is equal to its own raw type.

bjdodson:
Good point, thanks. Strangely we do have several instances of Map<String, List<String>> that were working fine. Hopefully this patch is more appropriate.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -284,11 +284,21 @@ public static Type collectionElementType(Type context, Class<?> contextRawType)
     Type mapType = getSupertype(context, contextRawType, Map.class);
     if (mapType instanceof ParameterizedType) {
       ParameterizedType mapParameterizedType = (ParameterizedType) mapType;
-      return mapParameterizedType.getActualTypeArguments();
+      return getCanonicalizedMapTypeArguments(mapParameterizedType);
     }
     return new Type[] { Object.class, Object.class };
   }
 
+  private static Type[] getCanonicalizedMapTypeArguments(ParameterizedType parameterizedType) {
```
swankjesse: Oh awesome

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -284,11 +284,21 @@ public static Type collectionElementType(Type context, Class<?> contextRawType)
     Type mapType = getSupertype(context, contextRawType, Map.class);
     if (mapType instanceof ParameterizedType) {
       ParameterizedType mapParameterizedType = (ParameterizedType) mapType;
-      return mapParameterizedType.getActualTypeArguments();
+      return getCanonicalizedMapTypeArguments(mapParameterizedType);
```
swankjesse: Frustratingly, this isn’t even enough. We need to do likewise above for collections, as on 270.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -284,11 +284,21 @@ public static Type collectionElementType(Type context, Class<?> contextRawType)
...
     }
     return new Type[] { Object.class, Object.class };
   }
 
+  private static Type[] getCanonicalizedMapTypeArguments(ParameterizedType parameterizedType) {
+    Type[] types = parameterizedType.getActualTypeArguments();
+    for (int i = 0; i < types.length; i++) {
+      if (types[i] instanceof GenericArrayType) {
+        types[i] = getRawType(types[i]);
```
swankjesse: sigh, this isn’t strictly correct. Imagine `Map<String, List<String>[]>`. The value’s raw type is `List<?>[]`, but we want to retain `List<String>[]`. It’s so absurdly shitty that Java has two ways of modeling `byte[]`.

swankjesse: PROBABLY what we want to do is change the core Moshi loop so that when you call `Moshi.adapter()`, that’s where we convert one `byte[]` to the other. That’s a lot easier to do generally, and it means things that compose like `List` and `Map` will Just Work.

bjdodson:
Trying again. I was hoping to use Types.canonicalize() but unfortunately they are not compatible, even with this hardcoded lookup in place. Added the new `canonicalizePlatformType` instead.

bjdodson:
Any other approach we might try here?

swankjesse:
I tried to rebase this and failed. Looks like it’s all working as you intended though.
https://github.com/square/moshi/pull/278

PR #153, opened 2016-04-18
Optimize reading one of several expected values with Selection

swankjesse:
This isn't yet public API.

This relies on an unreleased Okio API.

This has a significant impact on performance. I measured parsing performance
improve from 89k ops/sec to 140k ops/sec on one benchmark.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -116,11 +116,14 @@ final class ClassJsonAdapter<T> extends JsonAdapter<T> {
   private final ClassFactory<T> classFactory;
   private final Map<String, FieldBinding<?>> fieldsMap;
   private final FieldBinding<?>[] fieldsArray;
+  private final JsonReader.Selection selection;
 
   ClassJsonAdapter(ClassFactory<T> classFactory, Map<String, FieldBinding<?>> fieldsMap) {
     this.classFactory = classFactory;
     this.fieldsMap = new LinkedHashMap<>(fieldsMap);
     this.fieldsArray = fieldsMap.values().toArray(new FieldBinding[fieldsMap.size()]);
+    this.selection = JsonReader.Selection.of(
+        fieldsMap.keySet().toArray(new String[fieldsMap.size()]));
   }
 
   @Override public T fromJson(JsonReader reader) throws IOException {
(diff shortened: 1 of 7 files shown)


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -141,13 +144,19 @@ private boolean includeField(boolean platformType, int modifiers) {
...
-        if (fieldBinding != null) {
-          fieldBinding.read(reader, result);
+        int index = reader.selectName(selection);
+        FieldBinding<?> fieldBinding;
+        if (index != -1) {
+          fieldBinding = fieldsArray[index];
         } else {
-          reader.skipValue();
+          String name = reader.nextName();
+          fieldBinding = fieldsMap.get(name);
```
JakeWharton: Is this code path necessary? When would this result in a hit when the selection did not?

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -348,6 +364,39 @@ public static JsonReader of(BufferedSource source) {
...
+      this.doubleQuoteSuffix = doubleQuoteSuffix;
+    }
+
+    public static Selection of(String... strings) {
+      try {
+        ByteString[] result = new ByteString[strings.length];
+        Buffer buffer = new Buffer();
+        for (int i = 0; i < strings.length; i++) {
+          BufferedSinkJsonWriter.string(buffer, strings[i]);
+          buffer.readByte(); // Skip the leading double quote (but leave the trailing one).
```
JakeWharton: Why skip the leading quote instead of just matching the entire thing?

JakeWharton: Actually maybe that's too big of a behavior change. When you peek you consume the first double quote, right?

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
JakeWharton: Looks like it's the needlessly encoded case. :+1:

JakeWharton:
Looks great. :+1:

moshi/src/main/java/com/squareup/moshi/JsonReader.java
swankjesse: Exactly.

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
swankjesse: Yeah, and I'm tempted to change it so that Selection handles that internally. Not yet though. Maybe if we promote this to being a public API.

Issue #148, opened 2016-04-06
Android does not recognize parametrized type adapters


kotucz:
On android parametrized types are not matched properly against `@FromJson` and `@ToJson` annotated method and that fails to recognize and use provided adapter. In this case the `RulesListAdapter` is recognized.

```java
    JsonAdapter<RulesList<Rule>> jsonAdapter() {
        final Moshi moshi = new Moshi.Builder().add(new RulesListAdapter()).build();
        final Type type = Types.newParameterizedType(RulesList.class, Rule.class);
        return moshi.adapter(type);
    }
```

[Full failing androidTest is here](https://gist.github.com/kotucz/24d3932e87d5aa41363fbb12dccef9ec)

Stepping through the code I found, that [`adapterMethod.type.equals(type)`](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java#L231) does evaluate incorrectly to `false` when matching two same types (`toString()` value is the same) `moshi.MoshiParametrizedTypeAndroidTest$RulesList<moshi.MoshiParametrizedTypeAndroidTest$Rule>`

At runtime `adapterMethod.type` is instance of [`org.apache.harmony.luni.lang.reflect.ImplForType` ](https://android.googlesource.com/platform/libcore/+/b8f93421684a6292496804ca30d8428d2602ce7a/libdvm/src/main/java/org/apache/harmony/luni/lang/reflect/ImplForType.java
) on android 4.1 and 4.4 which seems that does not have `equals` implemented.
On Lollipop runtime type is [`libcore.reflect.ParameterizedTypeImpl`](https://android.googlesource.com/platform/libcore/+/b8f93421684a6292496804ca30d8428d2602ce7a/libdvm/src/main/java/org/apache/harmony/luni/lang/reflect/ImplForType.java) which works.
Running androidTest (attached) fails even on Lollipop.

PR #154, opened 2016-04-18
Workaround for android reflection bugs

kotucz:
There are android issues with reflection types, that cause Moshi to not find adapters for parametrized types.
Full description is in #148. Here I provide runnable android tests that show the issue and fix for the current issue.
The android test project is set up via gradle for now (sorry I can't maven :))


swankjesse:
Please don't change our build system in your pull request.

kotucz:
Sorry, I need to have the issue fixed. How can I run the android tests on the build system? The issue may not occur otherwise...

JakeWharton:
We don't want Android-specific tests as part of the project. You should be able to implement your own `ParameterizedType` which violates the `equals` contract in the same way to trigger the erroneous behavior on the JVM.

kotucz:
Created new pull request #155

PR #155, opened 2016-04-22
Workaround for Android ParameterizedType implementation bug

### LINKED ISSUE #148: Android does not recognize parametrized type adapters

kotucz:
Fixes #148

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
@@ -228,7 +228,7 @@ final class AdapterMethodsFactory implements JsonAdapter.Factory {
       List<AdapterMethod> adapterMethods, Type type, Set<? extends Annotation> annotations) {
     for (int i = 0, size = adapterMethods.size(); i < size; i++) {
       AdapterMethod adapterMethod = adapterMethods.get(i);
-      if (adapterMethod.type.equals(type) && adapterMethod.annotations.equals(annotations)) {
+      if (Types.equals(adapterMethod.type, type) && adapterMethod.annotations.equals(annotations)) {
         return adapterMethod;
       }
     }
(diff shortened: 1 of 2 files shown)


swankjesse:
Merged manually as d203cf741a41ea0d09116b5ec732f4a2a4ed2626.

swankjesse:
Thanks!

PR #156, opened 2016-04-24
Track Okio snapshot API change to Options.

swankjesse:
Also rename our own Selection class to Options since it's semantically
the same thing.

https://github.com/square/okio/pull/215


JakeWharton:
Word

PR #158, opened 2016-04-29
Add writer value overload for boxed booleans.

JakeWharton:
Autoboxing resolves boxed longs and doubles to `value(Number)`, but a boxed boolean would otherwise resolve to `value(boolean)` with an implicit call to `booleanValue()` which has the potential to throw NPEs.

Same change as https://github.com/google/gson/pull/836

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
@@ -260,6 +260,13 @@ final class BufferedSinkJsonWriter extends JsonWriter {
     return this;
   }
 
+  @Override public JsonWriter value(Boolean value) throws IOException {
+    if (value == null) {
+      return nullValue();
+    }
+    return value(value.booleanValue());
+  }
+
   @Override public JsonWriter value(double value) throws IOException {
     if (Double.isNaN(value) || Double.isInfinite(value)) {
       throw new IllegalArgumentException("Numeric values must be finite, but was " + value);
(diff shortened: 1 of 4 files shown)


moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
```diff
@@ -260,6 +260,17 @@ private void writeDeferredName() throws IOException {
     return this;
   }
 
+  @Override public JsonWriter value(Boolean value) throws IOException {
+    if (value == null) {
+      return nullValue();
+    }
+    writeDeferredName();
```
swankjesse: `return value(value.booleanValue())` ?

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -76,7 +76,7 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
     }
 
     @Override public void toJson(JsonWriter writer, Boolean value) throws IOException {
-      writer.value(value);
+      writer.value(value.booleanValue());
```
swankjesse: NPE here? hmm

swankjesse:
LGTM

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
JakeWharton: By design. There's a test for it! `Boolean.class` calls `.nullSafe()`

moshi/src/main/java/com/squareup/moshi/BufferedSinkJsonWriter.java
JakeWharton: yep. Shoulda done that for Gson as well...

PR #166, opened 2016-06-02
Fail on invalid escapes.

swankjesse:
Related to https://github.com/google/gson/issues/824


lzvoyager:
I'm trying to upgrade our app to use latest Gson from 2.2.4 and need help clarify the behaviour of escape sequences. e.g. Is the following string a well-formed Json string?

"{'value': '\\$pRice'}"

Using Gson 2.7 or 2.8 to parse the above string results in: "Invalid escape sequence at ..." - is gson trying to do recursive escaping?

My expectation of the above is - the value should be "\$pRice".

lzvoyager:
Sorry - I missed 1 backslash in my example string, the string which I think should be accepted as valid JSON is:

'value': '\\$pRice'

PR #170, opened 2016-06-06
Ignore BOM

### LINKED ISSUE #169: UTF-8 Byte-order Mark causes parsing failure
When an otherwise valid JSON response from an API includes the optional BOM (U+FEFF BYTE ORDER MARK), it fails to parse the stream because it doesn't know what to do with it. Can you start consuming it correctly?

My situation is a .NET Web API backend that is including the BOM and I can't make it stop.

toadzky:
Ignore byte order mark when doing first peek. Fixes #169

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -240,8 +240,7 @@ final class BufferedSourceJsonReader extends JsonReader {
     int peekStack = stack[stackSize - 1];
 
     if (peekStack == JsonScope.EMPTY_DOCUMENT
-            && source.request(BYTE_ORDER_MARK.size())
-            && buffer.indexOf(BYTE_ORDER_MARK) == 0) {
+            && source.rangeEquals(0, BYTE_ORDER_MARK)) {
       buffer.skip(BYTE_ORDER_MARK.size());
     }
 
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -29,6 +29,7 @@ final class BufferedSourceJsonReader extends JsonReader {
   private static final ByteString UNQUOTED_STRING_TERMINALS
       = ByteString.encodeUtf8("{}[]:, \n\t\r\f/\\;#=");
   private static final ByteString LINEFEED_OR_CARRIAGE_RETURN = ByteString.encodeUtf8("\n\r");
+  private static final ByteString BYTE_ORDER_MARK = ByteString.encodeUtf8("\ufeff");
 
   private static final int PEEKED_NONE = 0;
   private static final int PEEKED_BEGIN_OBJECT = 1;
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -236,6 +237,9 @@
   }
 
   private int doPeek() throws IOException {
+    if (stackSize == 1 && buffer.indexOf(BYTE_ORDER_MARK) == 0) {
```
JakeWharton: This only works because the test uses a `Buffer` as its `BufferedSource`. In a streamed source this buffer will be empty unless you do `request(BYTE_ORDER_MARK.size())` first

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -236,6 +237,9 @@
...
+      this.buffer.skip(BYTE_ORDER_MARK.size());
```
JakeWharton: Needless `this.` qualifier

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
JakeWharton: What this really wants is `rangeEquals()`: https://github.com/square/okio/issues/165

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -236,6 +237,11 @@
   }
 
   private int doPeek() throws IOException {
+    if (stackSize == 1
+            && buffer.request(BYTE_ORDER_MARK.size())
```
JakeWharton: This should be `source.request`. We need to fill `buffer` by requesting from `source`

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -236,6 +237,11 @@
   }
 
   private int doPeek() throws IOException {
+    if (stackSize == 1
```
JakeWharton: I'm thinking this should be below `peekStack` and check for `EMPTY_DOCUMENT` instead? Otherwise we potentially match the BOM anywhere in the root object.

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -237,6 +238,13 @@
 
   private int doPeek() throws IOException {
     int peekStack = stack[stackSize - 1];
+
+    if (peekStack == JsonScope.EMPTY_DOCUMENT
+            && source.request(BYTE_ORDER_MARK.size())
+            && buffer.indexOf(BYTE_ORDER_MARK) == 0) {
```
JakeWharton: `rangeEquals` just got merged into Okio so once it's released this can become:
```java
if (peekStack == JsonScope.EMPTY_DOCUMENT && source.rangeEquals(0, BYTE_ORDER_MARK)) {
  source.skip(BYTE_ORDER_MARK.size());
}
```
which reads a lot nicer.

JakeWharton: This was released in 1.9.0. I'm tempted to say that since it's so easy to test and skip the BOM that we don't really need to add it to Moshi.

toadzky: The reason to add it to moshi is so when people use moshi with retrofit (or
any other library on top of its reflection parsing), they don't have to
write manual parsers for everything.

JakeWharton: Yep. True.

JakeWharton:
Thanks for the PR, and keeping up with it. See my comment on the original issue as to how we're going to handle this.

toadzky:
No problem. Thanks for the feedback to help me make it better.

PR #175, opened 2016-06-22
Reduce Visibility of methods

vanniktech:
Dunno if this is wanted or not.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.java
+++ b/moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.java
@@ -44,18 +44,18 @@ final class LinkedHashTreeMap<K, V> extends AbstractMap<K, V> implements Seriali
     }
   };
 
-  Comparator<? super K> comparator;
-  Node<K, V>[] table;
-  final Node<K, V> header;
+  private Comparator<? super K> comparator;
+  private Node<K, V>[] table;
+  private final Node<K, V> header;
   int size = 0;
-  int modCount = 0;
-  int threshold;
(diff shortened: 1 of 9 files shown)


moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.java
```diff
@@ -44,18 +44,18 @@ public int compare(Comparable a, Comparable b) {
...
-  Comparator<? super K> comparator;
-  Node<K, V>[] table;
-  final Node<K, V> header;
+  private Comparator<? super K> comparator;
+  private Node<K, V>[] table;
+  private final Node<K, V> header;
   int size = 0;
-  int modCount = 0;
-  int threshold;
+  private int modCount = 0;
```
NightlyNexus: `table` and `modCount` are accessed by the inner class `LinkedTreeMapIterator`, so synthetic accessor methods would be created for them if set to private.
Also, I think this class is copied from Gson, so it's probably best not to modify it here. (Just imo, I don't know.)

swankjesse:
For package-private classes the `public` visibility is just a signal to the reader that this method is expected to be used by other classes in the package. We tend to avoid some uses of `private` visibility because it may make the compiler and runtime do more work (for private classes).

swankjesse:
Thanks though.

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.java
vanniktech: Uh right didn't see that. Everywhere else I thought of that.

Issue #142, opened 2016-03-22
@ToJson and @FromJson should permit any number of JsonAdapters


szurbrigg:
That way we can make building custom delegating type adapters easier.

Also, cut & paste error on the `@ToJson` bad signature exception message.

maxandron:
Do you mean having different adapters for different json types?
e.g: use one adapter if encountering a `BEGIN_OBJECT` and another if encountering a `STRING` - both producing the same type

If this is what you mean, is there a way to do this currently?

swankjesse:
The goal here is to make composition easier:

```java
class CardAdapter {
  @ToJson String toJson(Card card, JsonAdapter<Suit> suitAdapter) {
    ...
  }

  @FromJson Card fromJson(String card, JsonAdapter<Suit> suitAdapter) {
    ..
  }
}
```

PR #197, opened 2016-10-14
Permit JsonAdapters to be injected in toJson(), fromJson() methods.

swankjesse:
Closes: https://github.com/square/moshi/issues/142

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
@@ -406,6 +406,66 @@ public final class AdapterMethodsTest {
     assertThat(moshi.adapter(c)).isSameAs(moshi.adapter(a));
   }
 
+  @Test public void writerAndReaderTakingJsonAdapterParameter() throws Exception {
+    Moshi moshi = new Moshi.Builder()
+        .add(new PointWriterAndReaderJsonAdapter())
+        .add(new JsonAdapterWithWriterAndReaderTakingJsonAdapterParameter())
+        .build();
+    JsonAdapter<Line> lineAdapter = moshi.adapter(Line.class);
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
```diff
@@ -406,6 +406,60 @@
...
+        .add(new JsonAdapterTakingJsonAdapterParameter())
+        .build();
+    JsonAdapter<Line> lineAdapter = moshi.adapter(Line.class);
+    Line line = new Line(new Point(5, 8), new Point(3, 2));
+    assertThat(lineAdapter.toJson(line)).isEqualTo("\"[5,8]..[3,2]\"");
+    assertThat(lineAdapter.fromJson("\"[5,8]..[3,2]\"")).isEqualTo(line);
+  }
+
+  static class JsonAdapterTakingJsonAdapterParameter {
+    @ToJson String lineToJson(Line line, JsonAdapter<Point> pointAdapter) throws IOException {
```
swankjesse: There’s two ways to do `@ToJson`: 
1. Returning an intermediate value that’s itself converted (like here).
2. Getting a JsonWriter passed in and using that.

I’m actually thinking the first isn’t a valid uses case. The problem is that you’re converting a thing to JSON but returning a thing that isn’t, yielding double encoding. In the test case above that’s a JSON array inside of a string. There’s plausible ways to make this work but I think it’s going to be rare.

This problem is symmetric for `@FromJson`.

Any opinions on supporting `JsonAdapter` parameters _only_ for methods that take a `JsonReader` or `JsonWriter` ?

JakeWharton: Sounds good to me

PR #201, opened 2016-10-15
Fix enclosed types with adapter methods.

swankjesse:
This was broken becase reflection was providing owner types but our
API didn't have a way to specify them.

Closes: https://github.com/square/moshi/issues/148


JakeWharton:
Ugh, this sucks :(

JakeWharton:
This can't be pulled from `getEnclosingClass()`?

swankjesse:
Yeah, if the class is static we could do the right thing automatically. But certain parts are potentially difficult:
```
class Box<T extends CharSequence> {
  static class Content {
  }
}
```
It’s unclear whether the enclosing type of `Content` is the raw `Box` or the wildcard type `Box<? extends CharSequence`.

Issue #189, opened 2016-08-29
Reference cycles fail badly


swankjesse:
```
java.lang.ArrayIndexOutOfBoundsException: 32
	at com.squareup.moshi.BufferedSinkJsonWriter.open(BufferedSinkJsonWriter.java:146)
	at com.squareup.moshi.BufferedSinkJsonWriter.beginObject(BufferedSinkJsonWriter.java:132)
	at com.squareup.moshi.ClassJsonAdapter.toJson(ClassJsonAdapter.java:170)
	at com.squareup.moshi.JsonAdapter$1.toJson(JsonAdapter.java:75)
	at com.squareup.moshi.StandardJsonAdapters$ObjectJsonAdapter.toJson(StandardJsonAdapters.java:322)
	at com.squareup.moshi.JsonAdapter$1.toJson(JsonAdapter.java:75)
	at com.squareup.moshi.ClassJsonAdapter$FieldBinding.write(ClassJsonAdapter.java:204)
	at com.squareup.moshi.ClassJsonAdapter.toJson(ClassJsonAdapter.java:173)
	at com.squareup.moshi.JsonAdapter$1.toJson(JsonAdapter.java:75)
	at com.squareup.moshi.ClassJsonAdapter$FieldBinding.write(ClassJsonAdapter.java:204)
	at com.squareup.moshi.ClassJsonAdapter.toJson(ClassJsonAdapter.java:173)
```

JakeLoo:
Why are the pathNames and pathIndices have the fixed size of 32?

PR #202, opened 2016-10-15
Limit to 31 levels of nested structure.

swankjesse:
With the implicit enclosing document this is 32 levels total. The limit
is arbitrary but sufficient - deeper limits yield code that fails with
StackOverflowExceptions during the depth-first traversal.

We were previously broken in JsonWriter on this - the code we added to
support path building was accessing the top of the stack before the
stack had been resized, causing a crash. Because this crash has existed
forever without much outcry we know the limit is likely sufficient for
our existing users.

Closes: https://github.com/square/moshi/issues/189

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -92,17 +92,17 @@ final class BufferedSourceJsonReader extends JsonReader {
    */
   private String peekedString;
 
-  /*
-   * The nesting stack. Using a manual array rather than an ArrayList saves 20%.
-   */
-  private int[] stack = new int[32];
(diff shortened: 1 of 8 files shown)


moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -92,9 +92,9 @@
    */
   private String peekedString;
 
-  /*
-   * The nesting stack. Using a manual array rather than an ArrayList saves 20%.
-   */
+  // The nesting stack. Using a manual array rather than an ArrayList saves 20%. This stack permits
+  // up to 32 levels of nesting including the top-level document. Deeper nesting is prone to trigger
+  // StackOverflowErrors.
   private int[] stack = new int[32];
```
JakeWharton: `final`

PR #203, opened 2016-10-16
Add support for static ToJson/FromJson methods if given Class

### LINKED ISSUE #140: When given a Class, look for static ToJson/FromJson methods.
Current behavior:

```
Exception in thread "main" java.lang.IllegalArgumentException: Expected at least one @ToJson or @FromJson method on java.lang.Class
    at com.squareup.moshi.AdapterMethodsFactory.get(AdapterMethodsFactory.java:129)
    at com.squareup.moshi.Moshi$Builder.add(Moshi.java:186)
```

serj-lotutovici:
Added `Moshi.Builder#add(Class<?> adapter)` to avoid unnecessary casting.
Renamed `AdapterMethodsFactory#get()` to `fromInstance()`, created a separate `fromClass()`. If was ether that, or inflate `get()` will multiple if-else statements.

Closes #140

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


swankjesse:
I literally just put this issue in our icebox. I don’t think we want this, only because it gives our users another decision to make.

@JakeWharton this was your request – fill me in?

JakeWharton:
I wrote static methods instead of instance methods because they're
stateless adapters. I assumed we required static methods and was surprised
to learn it was the opposite. I don't think it's a huge deal to omit
support for, it just surprised me and violated an assumption so I figured
I'd document it.

swankjesse:
Should you get static methods by passing `Foo.class` or `new Foo()` ?

serj-lotutovici:
The factory will pick up static methods in any case. If you pass `Foo.class` then instance methods should be omitted.

swankjesse:
I think I’m gonna decline this and the iceboxed issue. I could imagine supporting finding static methods on instances, but supporting both `Object` and `Class<?>` is needless complexity.

serj-lotutovici:
Sounds reasonable for me.

PR #210, opened 2016-10-20
Fix #209

### LINKED ISSUE #209: Escaped slashes fail to be deserialized with Moshi 1.3.0
Deserializing an escaped slash fails with version 1.3.0 of moshi where as with version 1.2.0 everything worked.

Test case that works for 1.2.0 but not for 1.3.0:

```java
@Test public void escapedSlash() throws Exception {
  JsonAdapter<String[]> adapter = moshi.adapter(String[].class);
  assertThat(adapter.fromJson("[\"\\/a\",\"b\"]")).containsExactly("/a", "b");
}
```

Is this a desirable regression? As far as I know this is a valid JSON and should work.

cbruegg:
Escaped forward slashes should be properly processed instead of crashing.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -1073,6 +1073,7 @@ final class BufferedSourceJsonReader extends JsonReader {
       case '\'':
       case '"':
       case '\\':
+      case '/':
         return (char) escaped;
 
       default:


JakeWharton:
Needs a test

cbruegg:
@JakeWharton Would [this one](https://github.com/square/moshi/issues/209#issue-184225228) be suitable or would you like to have a test that covers all escaped characters?

JakeWharton:
Jesse beat you to it. Thanks though!

JakeWharton:
Ideally we would cover all to prevent accidental regressions like this. If
you just want to pull in one for this change only that's better than
nothing though.

cbruegg:
@JakeWharton @swankjesse Actually I think it's wrong to throw an Exception in any case here. [RFC 7159](https://tools.ietf.org/html/rfc7159#section-7) states that `Any character may be escaped`. This [page](http://seriot.ch/parsing_json.html) may be of interest for you as well.

swankjesse:
While any character may be escaped, only a few characters may be escaped using a single leading `\`. For example, `n` cannot be escaped as `\n`.

PR #213, opened 2016-11-02
Fixed LinkedHashTreeMap not being serializable.

ps-feng:


JakeWharton:
Can we get a test that exercises serialization that would have failed previously?

I'm surprised we don't just writeReplace the whole thing to a LinkedHashMap...

ps-feng:
Will try to write the test when I have time. We discovered it because we had a crash (Android) when saving the state of a fragment that had a reference to an object with a `LinkedHashTreeMap`.

swankjesse:
Yeah, I thought writeReplace would have triggered this. Test case please?

ps-feng:
Hmm, trying to reproduce it with a test case and failing so far. `writeReplace()` is being called, like you were saying. 

Will try tomorrow again at work... weird, as the crash we were experiencing was clearly pointing in this direction :S

ps-feng:
Sorry guys, it was due to Proguard. Disabling it for Moshi eliminates the problem, `writeReplace` was being stripped out or obfuscated, not sure.

Closing the PR.

PR #215, opened 2016-11-07
Fix Long parsing for large values.

serj-lotutovici:
Made `BufferedSourceJsonReader#nextLong()` to rely on `BigDecemal` instead of `Double`.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -17,6 +17,7 @@ package com.squareup.moshi;
 
 import java.io.EOFException;
 import java.io.IOException;
+import java.math.BigDecimal;
 import okio.Buffer;
 import okio.BufferedSource;
 import okio.ByteString;
@@ -718,15 +719,15 @@ final class BufferedSourceJsonReader extends JsonReader {
     }
 
     peeked = PEEKED_BUFFERED;
-    double asDouble;
+    BigDecimal asDecimal;
     try {
-      asDouble = Double.parseDouble(peekedString);
+      asDecimal = new BigDecimal(peekedString);
     } catch (NumberFormatException e) {
       throw new JsonDataException("Expected a long but was " + peekedString
           + " at path " + getPath());
     }
-    long result = (long) asDouble;
-    if (result != asDouble) { // Make sure no precision was lost casting to 'long'.
+    long result = asDecimal.longValue();
+    if (result != asDecimal.doubleValue()) { // Make sure no precision was lost casting to 'long'.
       throw new JsonDataException("Expected a long but was " + peekedString
           + " at path " + getPath());
     }
--- a/moshi/src/test/java/com/squareup/moshi/BufferedSourceJsonReaderTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/BufferedSourceJsonReaderTest.java
@@ -621,7 +621,7 @@ public final class BufferedSourceJsonReaderTest {
    * This test fails because there's no double for 9223372036854775808, and our
    * long parsing uses Double.parseDouble() for fractional values.
    */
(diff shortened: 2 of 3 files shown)


swankjesse:
I like this because it’s more correct, but it’s going to harm us on benchmarks.

serj-lotutovici:
Yeah, was acting partially on impulse because of the wtf comment  while trying to fix another issue :) would it make sense to try to fork BigDecemal's parsing just for longs?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -718,15 +719,15 @@ private boolean isLiteral(int c) throws IOException {
...
-      asDouble = Double.parseDouble(peekedString);
+      asDecimal = new BigDecimal(peekedString);
     } catch (NumberFormatException e) {
       throw new JsonDataException("Expected a long but was " + peekedString
           + " at path " + getPath());
     }
-    long result = (long) asDouble;
-    if (result != asDouble) { // Make sure no precision was lost casting to 'long'.
+    long result = asDecimal.longValue();
+    if (result != asDecimal.doubleValue()) { // Make sure no precision was lost casting to 'long'.
```
swankjesse: does this still make sense?

swankjesse: I think the right check would be to convert the long back into a BigDecimal. Otherwise we might be truncated somehow.

serj-lotutovici: That will not work if the BigDecemal is equal to a floating point number. Let me check what breaks if we remove that

serj-lotutovici: Ok, the limit check breaks... If the parsed long is bigger then MAX_VALUE or smaller then MIN_VALUE this ensure validity. 
Trying to convert the result into a BigDecemal will break the comparison for some valid use cases.

serj-lotutovici:
Regarding performance I've run a small bench mark and to be hones I'm surprised.

Parsing an array of 300 longs into a list (ms/op)

Moshi 1.3.0: `parseLongs  avgt   10  0.055 ± 0.004  ms/op`
Moshi 1.4.0-SNAPSHOT: `parseLongs  avgt   10  0.051 ± 0.002  ms/op`

Could it be that `BigDecemal`'s parsing is faster then `Double.parseDouble()`?

Here's the [gist](https://gist.github.com/serj-lotutovici/e6ebfb61f3f675036759325fecbfe7d8)

dragonsinth:
It probably won't make much difference for any string that's actually a valid Long since the length will be relatively short and there's no decimals or exponents to worry about.

swankjesse:
@serj-lotutovici yeah, I think I’m mistaken to care much about the performance here. The happy path is already covered by `PEEK_LONG`. But I think your new code might have a bug where values like `9223372036854775805` or `9223372036854775808` might incorrectly fail.

serj-lotutovici:
@swankjesse how do I test that?
```
assertThat(adapter.fromJson("9223372036854775805")).isEqualTo(9223372036854775805L);
```
seams to pass and

 ```
adapter.fromJson("9223372036854775808");
```
should fail, since the value is Long.MAX_VALUE + 1L. no?

**Comment by swankjesse** [comment]
I merged this manually after making a small change to the precision loss check:

```
    long result;
    try {
      BigDecimal asDecimal = new BigDecimal(peekedString);
      result = asDecimal.longValueExact();
    } catch (NumberFormatException | ArithmeticException e) {
      throw new JsonDataException("Expected a long but was " + peekedString
          + " at path " + getPath());
    }
```

Otherwise we incorrectly `9223372036854775806.5` to `9223372036854775806`.

swankjesse:
Thanks!

PR #216, opened 2016-11-19
Make Builder.addAll public.

serj-lotutovici:


swankjesse:
I think the loop is probably sufficient? Bulk adding factories is pretty edge case, isn’t it?

serj-lotutovici:
It is. I was thinking of some extensions or dagger modules:
```
@Provides List<Factory> providesFactories() {
  returns lsitOfFactories;
}
```
Had similar code in my last project. But feel free to close.

swankjesse:
Cool. I think the closest we come is a Factory that just delegates to other Factories.

Issue #165, opened 2016-05-29
Expose JsonReader.selectName


adriancole:
In zipkin, we have a custom json adapter for reading and writing trace spans. I did an experiment moving from switching on key names to `reader.selectName(OPTIONS)`. This was very little work and resulted in performance improvements in our read benchmarks. It would be nice to have a way to access `JsonReader.selectName` method without sharing Moshi's package

Here are improvements measured in ops/microsecond from my laptop:

reading a client-only span: 106 -> 118
reading a local span: 173 -> 201
reading a client + server span: 107 -> 127

adriancole:
ps the implementation change looks something like this. Costs a little indirection, but since the struct hasn't changed in ages, not a biggie.

```java
// Add the universe of known fields
JsonReader.Options FIELDS = JsonReader.Options.of("serviceName", "ipv4", "port");

// later instead of something like this:
switch (reader.nextName()) {
  case "serviceName":

// do something like this:
switch (reader.selectName(FIELDS)) {
  case 0:
```

swankjesse:
Yeah, we can open it up. Gotta figure out if options if the top level type, and how graceful we are when the encoding isn't canonical.

serj-lotutovici:
This can be closed.

PR #217, opened 2016-11-19
Make JsonReader.selectName and JsonReader.selectString public.

serj-lotutovici:
This is just to make the discusion on #165 rolling.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
@@ -47,10 +47,11 @@ final class BufferedSourceJsonReader extends JsonReader {
   private static final int PEEKED_SINGLE_QUOTED_NAME = 12;
   private static final int PEEKED_DOUBLE_QUOTED_NAME = 13;
   private static final int PEEKED_UNQUOTED_NAME = 14;
+  private static final int PEEKED_BUFFERED_NAME = 15;
   /** When this is returned, the integer value is stored in peekedLong. */
-  private static final int PEEKED_LONG = 15;
-  private static final int PEEKED_NUMBER = 16;
-  private static final int PEEKED_EOF = 17;
+  private static final int PEEKED_LONG = 16;
+  private static final int PEEKED_NUMBER = 17;
+  private static final int PEEKED_EOF = 18;
 
   /* State machine when parsing numbers */
   private static final int NUMBER_CHAR_NONE = 0;
@@ -98,6 +99,7 @@ final class BufferedSourceJsonReader extends JsonReader {
   // StackOverflowErrors.
   private final int[] stack = new int[32];
   private int stackSize = 0;
+
   {
     stack[stackSize++] = JsonScope.EMPTY_DOCUMENT;
   }
@@ -215,6 +217,7 @@ final class BufferedSourceJsonReader extends JsonReader {
       case PEEKED_SINGLE_QUOTED_NAME:
       case PEEKED_DOUBLE_QUOTED_NAME:
       case PEEKED_UNQUOTED_NAME:
+      case PEEKED_BUFFERED_NAME:
         return Token.NAME;
       case PEEKED_TRUE:
       case PEEKED_FALSE:
@@ -545,6 +548,8 @@ final class BufferedSourceJsonReader extends JsonReader {
       result = nextQuotedValue(DOUBLE_QUOTE_OR_SLASH);
     } else if (p == PEEKED_SINGLE_QUOTED_NAME) {
       result = nextQuotedValue(SINGLE_QUOTE_OR_SLASH);
+    } else if (p == PEEKED_BUFFERED_NAME) {
+      result = peekedString;
     } else {
       throw new JsonDataException("Expected a name but was " + peek() + " at path " + getPath());
     }
@@ -553,12 +558,12 @@ final class BufferedSourceJsonReader extends JsonReader {
     return result;
   }
 
-  @Override int selectName(Options options) throws IOException {
+  @Override public int selectName(Options options) throws IOException {
     int p = peeked;
     if (p == PEEKED_NONE) {
       p = doPeek();
     }
-    if (p != PEEKED_DOUBLE_QUOTED_NAME) {
+    if (p  < PEEKED_SINGLE_QUOTED_NAME || p > PEEKED_BUFFERED_NAME) {
       return -1;
     }
 
@@ -566,8 +571,30 @@ final class BufferedSourceJsonReader extends JsonReader {
     if (result != -1) {
       peeked = PEEKED_NONE;
       pathNames[stackSize - 1] = options.strings[result];
+
+      return result;
     }
-    return result;
+
+    // The next name may be unnecessary escaped. Save the last recorded path name, so that we
+    // can restore the peek state in case we fail to find a match.
+    String lastPathName = pathNames[stackSize - 1];
+
+    String nextName = nextName();
+    for (int i = 0, size = options.strings.length; i < size; i++) {
+      if (nextName.equals(options.strings[i])) {
+        peeked = PEEKED_NONE;
+        pathNames[stackSize - 1] = nextName;
+
+        return i;
+      }
+    }
+
+    peeked = PEEKED_BUFFERED_NAME;
+    peekedString = nextName;
+    // We can't push the path further, make it seem like nothing happened.
+    pathNames[stackSize - 1] = lastPathName;
+
+    return -1;
   }
 
   @Override public String nextString() throws IOException {
@@ -597,12 +624,12 @@ final class BufferedSourceJsonReader extends JsonReader {
     return result;
   }
 
-  @Override int selectString(Options options) throws IOException {
+  @Override public int selectString(Options options) throws IOException {
     int p = peeked;
     if (p == PEEKED_NONE) {
       p = doPeek();
     }
-    if (p != PEEKED_DOUBLE_QUOTED) {
+    if (p < PEEKED_SINGLE_QUOTED || p > PEEKED_BUFFERED) {
       return -1;
     }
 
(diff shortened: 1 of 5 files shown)


swankjesse:
The trap we’re still facing here is that encoding might not be canonical. For example if any `\u0000`-style escapes are used, we won’t do the right thing. That leaves us with a couple options:

 * An API that sometimes doesn’t work.
 * An implementation that’s more complex to handle such cases.

Will peeking strings just work here? I’m unsure.

serj-lotutovici:
Since the later is kinda an edge case. What do you say to an api like: `int selectName(Options options, boolean canonical);`? The the default being `true` with `int selectName(Options options)`.
Then if the api doesn't work it's the users fault not the libraries.

I would be happy to give this a try, if you're ok with this.

swankjesse:
I’m convincing myself that it’s worth the effort to make it always work regardless of encoding. Wanna try to make that work?

serj-lotutovici:
Yup... Will try to come-up with something in next few days.

serj-lotutovici:
@swankjesse I've managed to come up with a working solution, but not 100% sure this is the right way. 
Also this still need some cleaning up:
- Updating java docs
- Deciding on whether Options should be an inner type or not.

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -562,12 +566,60 @@ private boolean isLiteral(int c) throws IOException {
...
+            byte[] hex = new byte[4];
+            int offset = 2;
+            while (offset != ESCAPE_BYTES_SIZE) {
+              hex[offset - 2] = buffer.getByte(i + offset);
+              offset++;
+            }
+
+            int value;
+            try {
+              value = Integer.parseInt(ByteString.of(hex).utf8(), 16);
```
serj-lotutovici: I don't like this in particular. Would appreciate a hint towards a more elegant way.

moshi/src/test/java/com/squareup/moshi/BufferedSourceJsonReaderTest.java
```diff
@@ -1927,9 +1927,9 @@ private String repeat(char c, int count) {
 
   /** Select does match necessarily escaping. The decoded value is used in the path. */
   @Test public void selectNecessaryEscaping() throws IOException {
-    JsonReader.Options options = JsonReader.Options.of("\n", "\u0000", "\"");
+    JsonReader.Options options = JsonReader.Options.of("\n", "\u0000", "\"", "coffee");
```
swankjesse: This probably wants to be another, independent test case.

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
swankjesse: I think this might be a decent strategy.

1. First attempt `source.select()`. If that works, we’re good.
2. Next call `nextString()`.
3. Check if the string matches any of the options. If it does, we’re done.
4. Push the consumed string back by setting `peeked` and `peekedString`. See `promoteNameToValue()` for an example.

The tricky part is the weird combinations of calls that should always leave `JsonReader` in a consistent state. It’s difficult ’cause we coerce types.

```
    JsonReader reader = newReader("[123]");
    reader.beginArray();
    assertEquals(-1, reader.selectString(JsonReader.Options.of("124")));
    assertEquals(123, reader.nextLong());
```

serj-lotutovici:
Implemented second run with `nextName()`. Had to introduce a new peek constant `PEEKED_BUFFERED_NAME`.

Few questions:
1. Should `selectString` follow the same logic (for consistency)? 
2. Should `Options` remain a sub class? or not?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -47,6 +47,7 @@
   private static final int PEEKED_SINGLE_QUOTED_NAME = 12;
   private static final int PEEKED_DOUBLE_QUOTED_NAME = 13;
   private static final int PEEKED_UNQUOTED_NAME = 14;
+  private static final int PEEKED_BUFFERED_NAME = 18;
```
swankjesse: Renumber these to be in sequence?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -563,10 +568,34 @@ private boolean isLiteral(int c) throws IOException {
     }
 
     int result = source.select(options.doubleQuoteSuffix);
-    if (result != -1) {
+    if (result == -1) {
+      // In case we need to push back
```
swankjesse: nit: http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -563,10 +568,34 @@ private boolean isLiteral(int c) throws IOException {
...
+
+          break;
+        }
+      }
+
+      // No luck, restore peeked
+      if (result == -1) {
+        peeked = PEEKED_BUFFERED_NAME;
+        peekedString = nextName;
+        // We can't push the path further, make it seam like nothing happened.
```
swankjesse: nit: seem

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -563,10 +568,34 @@ private boolean isLiteral(int c) throws IOException {
...
+        pathNames[stackSize - 1] = lastPathName;
```
swankjesse: ooh, I would have probably missed this.

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -607,7 +636,27 @@ private boolean isLiteral(int c) throws IOException {
...
+    if (result == -1) {
+      String nextString = nextString();
+      for (int i = 0, size = options.strings.length; i < size; i++) {
+        if (nextString.equals(options.strings[i])) {
+          result = i;
+
+          peeked = PEEKED_NONE;
+          pathIndices[stackSize - 1]++;
+
+          break;
```
swankjesse: replace break with return?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -607,7 +636,27 @@ private boolean isLiteral(int c) throws IOException {
...
+        }
+      }
+
+      // Restore the peek we must
+      if (result == -1) {
```
swankjesse: ... and then you don’t need this if

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -607,7 +636,27 @@ private boolean isLiteral(int c) throws IOException {
     }
 
     int result = source.select(options.doubleQuoteSuffix);
-    if (result != -1) {
+    if (result == -1) {
```
swankjesse: invert this if to easy-out?

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
swankjesse: (and similarly for `selectName()`)

moshi/src/main/java/com/squareup/moshi/BufferedSourceJsonReader.java
```diff
@@ -607,7 +636,27 @@ private boolean isLiteral(int c) throws IOException {
...
+
+          break;
+        }
+      }
+
+      // Restore the peek we must
+      if (result == -1) {
+        peeked = PEEKED_BUFFERED;
+        peekedString = nextString;
+        // Decrement the path indices
```
swankjesse: You can nuke this comment, the line of code says it all

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -359,13 +359,12 @@ public static JsonReader of(BufferedSource source) {
 
   /**
    * A set of strings to be chosen with {@link #selectName} or {@link #selectString}. This prepares
-   * the encoded values of the strings so they can be read directly from the input source. It cannot
-   * read arbitrary encodings of the strings: if any of a string's characters are unnecessarily
-   * escaped in the source JSON, that string will not be selected. Similarly, if the string is
-   * unquoted or uses single quotes in the source JSON, it will not be selected. Client code that
-   * uses this class should fall back to another mechanism to accommodate this possibility.
+   * the encoded values of the strings so they can be read directly from the input source.
+   * If the string is unquoted or uses single quotes in the source JSON, it will not be selected.
```
swankjesse: this is untrue now right?

swankjesse: ie. selection always works?

swankjesse: (I bet you can probably simplify the current only user of this now!)

swankjesse:
Love this. Thanks much.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
serj-lotutovici: Now the selections checks for a double quoted token. But let me check

serj-lotutovici: Yup, your rite... the doc is un-true. Will fix the tests an will push asap.

serj-lotutovici:
Btw, it just struck me. If the current piked value is a buffered one, I bet we can avoid a call to `source.select()` in that case. What do you think? I can do this in a follow up pr, when this is merged to check the assumption.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
serj-lotutovici: Remove the restriction doc. Will simplify ClassJsonAdapter and EnumJsonAdapter after this is merged (to keep the pr scope small).

swankjesse:
Thanks again. I'm ready for the follow ups!

PR #219, opened 2016-12-28
Make ArrayJsonAdapter implement toString().

serj-lotutovici:
It's not a bug. But really helps during debugging.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ArrayJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ArrayJsonAdapter.java
@@ -69,4 +69,8 @@ final class ArrayJsonAdapter extends JsonAdapter<Object> {
     }
     writer.endArray();
   }
+
+  @Override public String toString() {
+    return elementAdapter + ".array()";
+  }
 }


moshi/src/main/java/com/squareup/moshi/ArrayJsonAdapter.java
```diff
@@ -69,4 +69,8 @@
     }
     writer.endArray();
   }
+
+  @Override public String toString() {
+    return "JsonAdapter(" + elementClass.getSimpleName() + ").array()";
```
swankjesse: How about `"array(" + elementAdapter + ")"`

swankjesse: or `elementAdapter + "[]"` ?

swankjesse: (because the elementClass doesn‘t tell as much as its adapter. I’m thinking things like `Date[]` vs. `Iso8601Adapter[]`

serj-lotutovici: makes sense, but then we can get something like `JsonAdapter(Integer).nullSafe()[]`. I'm between that and `elementaAdapter + ".array()"`, in both cases we maintain the chain.

swankjesse: I'm not particular!

swankjesse:
Thanks! Sorry for the slow response.

PR #220, opened 2017-01-07
JsonAdapter.indent(String indent)

jmfayard:
Hello, 

Researching how to output human readable (prettyprinted) json turns to more longer as expected, which I don't regret because okio, okio.Buffer and JsonWriter are super interesting 👍 

My code now contains

```kotlin
fun  <T> JsonAdapter<T>.toPrettyJson(value: T, indent: String = "  ") : String {
    val buffer = Buffer()
    val writer = JsonWriter.of(buffer).apply { setIndent(indent) }
    toJson(writer, value)
    return buffer.readUtf8()
}
```

Still it's such a common use case, could a configuration option be added?

I implemented a failing test to explain myself in code, borrowing from `gson` 

[Gson: Compact Vs. Pretty Printing for JSON Output Format](https://github.com/google/gson/blob/master/UserGuide.md#TOC-Compact-Vs.-Pretty-Printing-for-JSON-Output-Format)

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -136,6 +136,31 @@ public abstract class JsonAdapter<T> {
     };
   }
 
+  /**
+   * Return a JSON adapter equal to this, but with pretty printing instead of the default compact form.
+   */
+  public JsonAdapter<T> indent(final String indent) {
+    final JsonAdapter<T> delegate = this;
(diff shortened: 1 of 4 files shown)


swankjesse:
Would this API work?

```
JsonAdapter<Pizza> pizzaAdapter = moshi.adapter(Pizza.class);
String json = pizzaAdapter.indenting("  ").toJson(pepperoniPizza);
```

jmfayard:
Yes!

swankjesse:
I merged this manually. Nice PR. Thanks!

PR #225, opened 2017-01-18
Disallow Moshi.Builder.add(null)

NightlyNexus:
I don't know if that test is really necessary.
I used `IllegalArgumentException` (not `NullPointerException`) for consistency with the existing usage.


NightlyNexus:
Oops, sorry, I would have squashed manually first. I just upload patch commits so incremental changes can be seen during review. I don't mean to clutter the git history.

PR #228, opened 2017-01-22
Prepare tests for a 2nd implementation of JsonWriter.

swankjesse:


moshi/src/test/java/com/squareup/moshi/JsonWriterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi;
+
+import java.util.Arrays;
+import java.util.List;
+import okio.Buffer;
+
+abstract class JsonWriterFactory {
```
swankjesse: This one’s stateful and so I changed JsonReaderFactory to be stateful for symmetry.

PR #230, opened 2017-01-22
Avoid calling source.select() if the value is buffered.

serj-lotutovici:
This will save us one loop. Not sure how to test this tho, or even if we need to.


swankjesse:
No test changes necessary; this is not a behavior change.

PR #233, opened 2017-01-25
Promote stack management to the JsonReader/JsonWriter supertypes.

swankjesse:
It turns out that we can reuse a lot of code with inheritance. Who knew?

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -171,15 +171,52 @@ import okio.ByteString;
  * of this class are not thread safe.
  */
 public abstract class JsonReader implements Closeable {
-  /**
-   * Returns a new instance that reads a JSON-encoded stream from {@code source}.
-   */
+  // The nesting stack. Using a manual array rather than an ArrayList saves 20%. This stack permits
+  // up to 32 levels of nesting including the top-level document. Deeper nesting is prone to trigger
+  // StackOverflowErrors.
+  int stackSize = 0;
+  final int[] scopes = new int[32];
+  final String[] pathNames = new String[32];
+  final int[] pathIndices = new int[32];
+
+  /** True to accept non-spec compliant JSON */
+  boolean lenient;
+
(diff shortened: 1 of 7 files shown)


moshi/src/test/java/com/squareup/moshi/JsonWriterPathTest.java
```diff
@@ -17,14 +17,27 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assume.assumeTrue;
 
+@RunWith(Parameterized.class)
 public final class JsonWriterPathTest {
+  @Parameter public JsonWriterFactory factory;
```
swankjesse: now this test operates on both DOMs and bytes

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -349,7 +394,9 @@ public static JsonReader of(BufferedSource source) {
    * Returns a <a href="http://goessner.net/articles/JsonPath/">JsonPath</a> to
    * the current location in the JSON value.
    */
-  public abstract String getPath();
+  public String getPath() {
```
JakeWharton: Should these all be `final` then?

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -276,5 +322,7 @@ public static JsonWriter of(BufferedSink sink) {
    * Returns a <a href="http://goessner.net/articles/JsonPath/">JsonPath</a> to
    * the current location in the JSON value.
    */
-  public abstract String getPath();
+  public String getPath() {
```
JakeWharton: Same `final` question.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
swankjesse: Fixed ’em all.

PR #235, opened 2017-01-25
Improve error message for qualified platform types

NightlyNexus:
Previously, the ClassJsonAdapter saw the platform type and gave its `IllegalArgumentException: Platform {type} annotated {annotations} requires explicit JsonAdapter to be registered`.
If the platform types are allowed but are just not caught by an adapter because they are qualified, they should fall through to the "no adapter found" Moshi error.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -111,11 +111,11 @@ final class ClassJsonAdapter<T> extends JsonAdapter<T> {
      */
     private boolean isPlatformType(Class<?> rawType) {
       String name = rawType.getName();
-      return name.startsWith("android.")
+      return (name.startsWith("android.")
           || name.startsWith("java.")
           || name.startsWith("javax.")
           || name.startsWith("kotlin.")
-          || name.startsWith("scala.");
+          || name.startsWith("scala.")) && !Types.isAllowedPlatformType(rawType);
     }
 
     /** Returns true if fields with {@code modifiers} are included in the emitted JSON. */
--- a/moshi/src/main/java/com/squareup/moshi/Types.java
+++ b/moshi/src/main/java/com/squareup/moshi/Types.java
@@ -386,6 +386,22 @@ public final class Types {
     return unknown;
   }
 
+  /**
+   * Returns true if this is a Type supported by {@link StandardJsonAdapters#FACTORY}.
+   */
+  static boolean isAllowedPlatformType(Type type) {
+    return type == Boolean.class
+        || type == Byte.class
+        || type == Character.class
+        || type == Double.class
+        || type == Float.class
+        || type == Integer.class
+        || type == Long.class
+        || type == Short.class
+        || type == String.class
+        || type == Object.class;
+  }
+
   private static int indexOf(Object[] array, Object toFind) {
     for (int i = 0; i < array.length; i++) {
       if (toFind.equals(array[i])) return i;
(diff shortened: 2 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -111,11 +111,20 @@ private void createFieldBindings(
...
     private boolean isPlatformType(Class<?> rawType) {
       String name = rawType.getName();
-      return name.startsWith("android.")
+      return (name.startsWith("android.")
           || name.startsWith("java.")
           || name.startsWith("javax.")
           || name.startsWith("kotlin.")
-          || name.startsWith("scala.");
+          || name.startsWith("scala.")) && !(rawType.equals(Boolean.class)
+          || rawType.equals(Byte.class)
```
swankjesse: maybe we want `Types.isPrimitive()` ?

swankjesse: or it’d be `Types.isBoxedPrimitive()` I suppose. Package-private.

moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -776,6 +776,24 @@
     }
   }
 
+  @Test public void noTypeAdapterForQualifiedPlatformType() {
+    Moshi moshi = new Moshi.Builder().build();
+    Field uppercaseStringField;
+    try {
+      uppercaseStringField = MoshiTest.class.getDeclaredField("uppercaseString");
+    } catch (NoSuchFieldException e) {
```
swankjesse: you can just declare the test as `throws Exception`

moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -776,6 +776,24 @@
...
+      throw new AssertionError(e);
+    }
+    try {
+      moshi.adapter(uppercaseStringField.getGenericType(),
+          Util.jsonAnnotations(uppercaseStringField));
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("No JsonAdapter for class java.lang.String "
```
swankjesse: NICE

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -386,6 +386,23 @@ static Type resolveTypeVariable(Type context, Class<?> contextRawType, TypeVaria
     return unknown;
   }
 
+  /**
+   * Returns true if this is a boxed primitive type like {@code Integer}. Returns false for all
+   * other types types including unboxed primitives and {@code java.lang.Void}.
+   */
```
NightlyNexus: I stole the doc from JavaPoet.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -386,6 +386,23 @@ static Type resolveTypeVariable(Type context, Class<?> contextRawType, TypeVaria
...
+    return type.equals(Boolean.class)
+        || type.equals(Byte.class)
+        || type.equals(Character.class)
+        || type.equals(Double.class)
+        || type.equals(Float.class)
+        || type.equals(Integer.class)
+        || type.equals(Long.class)
+        || type.equals(Short.class)
+        || type.equals(String.class)
+        || type.equals(Object.class);
```
NightlyNexus: Oh, shoot, wait, I wasn't thinking.
String.class and Object.class make this method name (and doc) lie.
This is more like "isTypeSupportedByStandardJsonAdapter" but Not That.

NightlyNexus: I renamed from `isBoxedPrimitive` to `isAllowedPlatformType`. Not super confident on that name.

NightlyNexus:
Tangental: Should Moshi support java.lang.Number?

swankjesse:
Probably not. `Number` is very difficult to do well.

Issue #136, opened 2016-02-22
Examples for JsonAdapter.Factory


swankjesse:
We don’t have anything in the readme, or any samples. We should probably add these.

feinstein:
@swankjesse it would be nice to have some documentation explaining the purpose of a `JsonAdapter.Factory`. Maybe this is obvious for the more experienced, but it's hard for the beginners to understand when it should be used and all it can do.

feinstein:
@swankjesse for example, as a beginner, I don't get when should I create a Moshi instance with:
````
Moshi moshi = new Moshi.Builder().add(new MyCustomAdapter()).build();
````

Or 
````
Moshi moshi = new Moshi.Builder().add(new MyCustomAdapterFactory()).build();
````

(I also don't understand when should I create an Adapter that extends `JsonAdapter` or one there uses annotations such as `@ToJson`, which are btw the only type the documentations mention)

JakeWharton:
Not in a position to provide a longer answer right now, but to your first
point, an adapter is for a specific type, a factory is called for every
type and gets to choose whether to return an adapter or ignore it (return
null).

feinstein:
I see, so passing an Adapter should be a lot better right? I guess the Adapter will be reused, while the factory will be creating a new adapter all the time, besides the performance hit for scanning multiple factories.

If this is true, Moshi should have a way to get a JSON array adapter without the need for a factory being passed to the Builder, since this is a very common use case.

JakeWharton:
No, it doesn't matter. The factory is only called once per-type. If it
returns an adapter that instance is cached.

feinstein:
Oh that's very nice.....but still feels a bit cumbersome to make an entire Factory (instead to just 1 line of code) to parse a JSON array to `List`, `Collection`, `Iterable` and etc.

JakeWharton:
That code only has to be written once so it's acceptable that it requires a
few lines.

feinstein:
I agree, but maybe it should come packed with Moshi, so we would just add the `JsonArrayAdapterFactory<T>` to the `Builder`?

I understand Moshi is very streamlined, but AFAICS this is such a regular use case, people will be doing it anyway.

feinstein:
Just an idea, I appreciate you guys being welcome to user input :)

JakeWharton:
I'm not sure this is a common case.

You should also be able to write a @FromJson method with a
JsonAdapter<List<Whatever>> argument to get access to a delegate adapter.

[...]

Issue #188, opened 2016-08-15
Making parsing more lenient


lyptt:
Currently if the format of the value doesn't match what was expected, a JsonDataException is thrown. Ideally it'd be nice to resume parsing the rest of the data, but set that value to null.

We don't have any control over the data we're being sent, and we -do- need to be able to parse whatever's valid.

If there's any advice on how to approach this, it would be greatly appreciated.

swankjesse:
That’s pretty tough to do generally. You might be able to do something with a JsonAdapter.Factory to use `peek()` and `skip()` if the types don’t match.

lyptt:
Yeah, so looking at the exception further it seems it's the StandardTypeAdapter's ObjectTypeAdapter that's throwing the exception. It sees an object but it's supposed to return a string in my case. 

I've experimented with doing this with GSON using skip and it forever complains about that original issue on every object - it seems like it's not possible to get it to move onto the next field in the class, only the next entry in the JSON object.

Since this library is based on GSON I assume this would have the same issue? If there's any insight here you could provide that'd be incredibly helpful. I'm happy to work on a PR if necessary.

swankjesse:
Not sure what action to take here. Seems difficult to do in general.

NightlyNexus:
@lyptt 
This is now possible in 1.4:
https://github.com/square/moshi/blob/49092ece9623c8acd10cd509755b459460a5a40e/examples/src/main/java/com/squareup/moshi/recipes/DefaultOnDataMismatchAdapter.java

lyptt:
Wow that's awesome! I'll give that a go, thanks!

lyptt:
This worked absolutely perfectly. Thanks a lot for making this possible! 😀

PR #236, opened 2017-01-25
Add Unwrap JsonAdapter.Factory sample

NightlyNexus:
Here's a swing to help https://github.com/square/moshi/issues/136
I won't take offense if you have different examples in mind or just don't like these, obviously, haha!

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
+++ b/examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi.recipes;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonQualifier;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.Types;
+import com.squareup.moshi.recipes.Unwrap.EnvelopeJsonAdapter.Enveloped;
+import com.squareup.moshi.recipes.models.Card;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Retention;
+import java.lang.reflect.Type;
+import java.util.Set;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class Unwrap {
+  public static void main(String[] args) throws Exception {
+    String json = ""
+        + "{\"data\":"
+        + "  {\n"
+        + "    \"rank\": \"4\",\n"
+        + "    \"suit\": \"CLUBS\"\n"
+        + "  }"
+        + "}";
+    Moshi moshi = new Moshi.Builder().add(EnvelopeJsonAdapter.FACTORY).build();
+    JsonAdapter<Card> adapter = moshi.adapter(Card.class, Enveloped.class);
+    Card out = adapter.fromJson(json);
+    System.out.println(out);
+  }
+
+  public static final class EnvelopeJsonAdapter extends JsonAdapter<Object> {
+    public static final JsonAdapter.Factory FACTORY = new Factory() {
+      @Override
+      public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+        Set<? extends Annotation> delegateAnnotations =
+            Types.nextAnnotations(annotations, Enveloped.class);
+        if (delegateAnnotations == null) {
+          return null;
+        }
+        Type envelope =
+            Types.newParameterizedTypeWithOwner(EnvelopeJsonAdapter.class, Envelope.class, type);
+        JsonAdapter<Envelope<?>> delegate = moshi.nextAdapter(this, envelope, delegateAnnotations);
+        return new EnvelopeJsonAdapter(delegate);
+      }
+    };
+
+    @Retention(RUNTIME) @JsonQualifier public @interface Enveloped {
+    }
+
+    private static final class Envelope<T> {
+      final T data;
+
+      Envelope(T data) {
+        this.data = data;
+      }
+    }
+
+    private final JsonAdapter<Envelope<?>> delegate;
+
+    EnvelopeJsonAdapter(JsonAdapter<Envelope<?>> delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override public Object fromJson(JsonReader reader) throws IOException {
+      return delegate.fromJson(reader).data;
+    }
+
+    @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      delegate.toJson(new Envelope<>(value));
+    }
+  }
+}


examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+      @Override public Class<? extends Annotation> annotationType() {
+        return Enveloped.class;
+      }
+    };
+    Moshi moshi = new Moshi.Builder().add(new EnvelopeJsonAdapterFactory()).build();
+    JsonAdapter<Train> adapter = moshi.adapter(Train.class, Collections.singleton(annotation));
+    Train out;
+    try {
+      out = adapter.fromJson("{\"data\":{\"name\":\"James\"}}");
+    } catch (IOException e) {
```
swankjesse: just declare that main `throws Exception` ?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Retention;
+import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.Locale;
+import java.util.Set;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class EnvelopeJsonAdapterFactory implements JsonAdapter.Factory {
```
swankjesse: mind renaming this to be the example name, say “Unwrap“, and then making the factory a nested class? Otherwise `main()` is in the middle of the actual sample stuff

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final class Train {
+    String name;
+  }
+
+  public static void main(String[] args) {
+    Enveloped annotation = new Enveloped() {
```
swankjesse: ick. This works but I never indended for end-users to implement annotation interfaces this way. It’s gnarly.  Particularly since equals() and hashCode() don’t quite do what we want.

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+
+      @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+        delegate.toJson(new Envelope<>(value));
+      }
+    };
+  }
+
+  private static boolean hasJsonQualifier(Set<? extends Annotation> annotations,
+      Class<? extends Annotation> jsonQualifier) {
+    if (!jsonQualifier.isAnnotationPresent(JsonQualifier.class)) {
```
swankjesse: Callers should be able to assume that Moshi is doing this check for them already.

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+      throw new IllegalArgumentException(
+          String.format(Locale.US, "%1$s is not a JsonQualifier.", jsonQualifier));
+    }
+    for (Annotation annotation : annotations) {
```
swankjesse: We should probably stash this in a sample somewhere.

swankjesse: Or in a helper method?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
```diff
@@ -0,0 +1,101 @@
...
+        return delegate.fromJson(reader).data;
+      }
+
+      @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+        delegate.toJson(new Envelope<>(value));
+      }
+    };
+  }
+
+  private static boolean hasJsonQualifier(Set<? extends Annotation> annotations,
```
swankjesse: This example tangles up two complimentary features:
 - annotated types
 - adapter factories

I think we want to make it easier (ie. write less code) to do annotations before we go off writing samples where end users need them this much. As is the experience of dealing with ’em is pretty tragic.

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
```diff
@@ -0,0 +1,92 @@
...
+      @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+        if (value == null) {
+          writer.nullValue();
+          return;
+        }
+        checkRequiredFields(value);
+        delegate.toJson(writer, value);
+      }
+
+      private void checkRequiredFields(Object value) {
```
swankjesse: This example hurts because it does reflection many times where it could be done once.
https://publicobject.com/2016/03/24/reflection-machines/

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
```diff
@@ -0,0 +1,92 @@
...
+      adapter.fromJson("{}");
+      throw new AssertionError();
+    } catch (IOException e) {
+      throw new AssertionError(e);
+    } catch (IllegalStateException expected) {
+      System.err.println(expected.getMessage());
+    }
+  }
+
+  @Retention(RUNTIME) @interface Required {
```
swankjesse: The lack of a `@JsonAdapter` here and the reasons why it wouldn’t work are interesting and subtle

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+    final JsonAdapter<Object> delegate = moshi.nextAdapter(this, type, annotations);
+    return new JsonAdapter<Object>() {
+      @Override public Object fromJson(JsonReader reader) throws IOException {
+        Object value = delegate.fromJson(reader);
```
swankjesse: dig.

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
swankjesse: Better would be to precompute the `List<Field>` that are required, and only if non-empty do you even return an adapter from the factory

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
```diff
@@ -0,0 +1,92 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Retention;
+import java.lang.reflect.Field;
+import java.lang.reflect.Type;
+import java.util.Locale;
+import java.util.Set;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class RequiredJsonAdapterFactory implements JsonAdapter.Factory {
```
swankjesse: As before, I have slight preference to naming the example differently from the class it’s exampling.

ALSO I really like `FACTORY` to be a static final instance inside the type it’s factorying

examples/src/main/java/com/squareup/moshi/recipes/TypeFailSafeJsonAdapterFactory.java
```diff
@@ -0,0 +1,102 @@
...
+    if (delegateAnnotations == null) {
+      return null;
+    }
+    final JsonAdapter<Object> delegate = moshi.nextAdapter(this, type, delegateAnnotations);
+    return new JsonAdapter<Object>() {
+      @Override public Object fromJson(JsonReader reader) throws IOException {
+        try {
+          return delegate.fromJson(reader);
+        } catch (JsonDataException e) {
+          reader.skipValue();
```
swankjesse: Unsure how I feel about this. It won’t do the right thing if the matched type isn’t a primitive or string because the exception might be thrown by a deeper method.

swankjesse: Does anyone want this behavior?

NightlyNexus:
oops, just saw the review here. I'll take a look, thank you!

NightlyNexus:
>But sample code must be judged most harshly.

Perfect, that's what I selfishly want! I really appreciate it!
I will follow up with just the Envelop/Unwrap later tonight. Thanks again!

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
NightlyNexus: Yep, nobody should do this for real, but I am unsure of a good demo for it here.
- Have another wrapper type, so that this type can be annotated in that class.
- Have a field in `Unwrap` that has the annotation and then get the field with reflection.
- Implement the annotation correctly.
- Something else?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
NightlyNexus: Cool, should I remove the annotation and just do the unwrapping on everything?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
NightlyNexus: Okay, I'll take an look!

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
NightlyNexus: This would of course fail at the top. This whole thing is weird and unfortunate, I think. Removing, anyway.

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
NightlyNexus: Oh, right! I was starting to think through a `Map<Class<?>, Field[]>` in the JsonAdapter and thinking there had to be a better way.
Simply grabbing the fields from the class in the Factory is fantastically better!

examples/src/main/java/com/squareup/moshi/recipes/TypeFailSafeJsonAdapterFactory.java
NightlyNexus: ah, yep, this would require more complexity.

NightlyNexus: The behavior of skipping mismatched types? It's hopefully not a common case; it was something I had to work around at work at one point for optional id fields that were coming back as unrelated strings.
Anyway, I'm throwing out all but the `Unwrap`. Thank you for all the feedback!

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override public Object fromJson(JsonReader reader) throws IOException {
+      return delegate.fromJson(reader).data;
+    }
+
+    @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      delegate.toJson(new Envelope<>(value));
+    }
+  }
+
+  public static boolean hasJsonQualifier(Set<? extends Annotation> annotations,
```
NightlyNexus: Where should this method go?

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+      Class<? extends Annotation> jsonQualifier) {
+    if (!jsonQualifier.isAnnotationPresent(JsonQualifier.class)) {
```
NightlyNexus: If this method stays here as a helper, I think this check should be removed, and the check around 107 should just be `annotation instanceof Enveloped`.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+import java.lang.annotation.Retention;
+import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.Locale;
+import java.util.Set;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class Unwrap {
+  @Enveloped private static Object unused;
```
NightlyNexus: pretty gross. What might be a better example?

NightlyNexus: Is this still a confusing example, mixing the Factory with JsonQualifiers?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
swankjesse: We just need a better way to refer to an annotation. Implementing it isn’t good, especially since it has particular requirements for equals() and hashCode().

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
swankjesse: yep

swankjesse: Or everything that is an `Envelope<T>`

examples/src/main/java/com/squareup/moshi/recipes/RequiredJsonAdapterFactory.java
swankjesse: Yeah. Having a both `Factory` and a `JsonAdapter` isn’t there to make you do more work; it’s so you can precompute stuff in the `Factory` so the `JsonAdapter` is fast.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+        + "}";
+    Annotation annotation = Unwrap.class.getDeclaredField("unused").getAnnotations()[0];
+    Moshi moshi = new Moshi.Builder().add(EnvelopJsonAdapter.FACTORY).build();
+    JsonAdapter<Card> adapter =
+        moshi.adapter(Card.class, Collections.<Annotation>singleton(annotation));
+    Card out = adapter.fromJson(json);
+    System.out.println(out);
+  }
+
+  public static final class EnvelopJsonAdapter extends JsonAdapter<Object> {
```
swankjesse: nit. Envelope

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
swankjesse: It’s probably the best we can do. I think we probably want a mechanism on Moshi where you can specify an annotation using a type literal `Enveloped.class` and just expand it out to the only reasonable implementation. I can do that, or you can read about dynamic proxies and annotation’s `hashCode()` method if you wanna try it.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+      @Override
+      public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+        if (!hasJsonQualifier(annotations, Enveloped.class)) {
+          return null;
+        }
+        Type envelope =
+            Types.newParameterizedTypeWithOwner(EnvelopJsonAdapter.class, Envelope.class,
+                Types.getRawType(type));
+        JsonAdapter<Envelope<?>> delegate =
+            moshi.nextAdapter(this, envelope, Collections.<Annotation>emptySet());
```
swankjesse: Should we return a new set with the original annotations still there, but `Enveloped` gone?

examples/src/main/java/com/squareup/moshi/recipes/EnvelopeJsonAdapterFactory.java
JakeWharton: There's `@AutoAnnotation`: https://github.com/google/auto/blob/master/value/userguide/howto.md#annotation

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
swankjesse: ... in case you’re both `@Enveloped` and say, `@Required` or something?

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+        return new EnvelopJsonAdapter(delegate);
+      }
+    };
+
+    @Retention(RUNTIME) @JsonQualifier @interface Enveloped {
+    }
+
+    private static final class Envelope<T> {
```
swankjesse: you know what? We can probably remove this type ’cause it’s a bit distracting, and just change the contents of `fromJson` to do the boxing+unboxing manually:

```
@Override public Object fromJson(JsonReader reader) throws IOException {
  T result = null;
  reader.beginObject();
  while (reader.hasNext()) {
    String name = reader.nextName();
    if (name.equals("data")) {
      result = delegate.fromJson(reader);
    } else {
      reader.skipValue();
  }
  reader.endObject();
  return result;
```

swankjesse: Or maybe that’s worse? I’m too deep into it to know

swankjesse: Yeah staring at it some more I think what you got is better

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+  public static final class EnvelopJsonAdapter extends JsonAdapter<Object> {
+    public static final JsonAdapter.Factory FACTORY = new Factory() {
+      @Override
+      public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+        if (!hasJsonQualifier(annotations, Enveloped.class)) {
+          return null;
+        }
+        Type envelope =
+            Types.newParameterizedTypeWithOwner(EnvelopJsonAdapter.class, Envelope.class,
+                Types.getRawType(type));
```
swankjesse: why call `getRawType()` here? In theory this breaks for say, `Envelope<List<String>>`

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
NightlyNexus: I will read up on it and give it a shot! Thank you!

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
NightlyNexus: oops, :man_facepalming: ha

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
NightlyNexus: Yep! thanks for the catch.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
NightlyNexus: oh, yeah. will fix.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
NightlyNexus: Yes, I considered this, and it's what I do at work, but I thought it might get weird for this example. Cool, I'll leave the actual envelope class here, then.

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override public Object fromJson(JsonReader reader) throws IOException {
+      return delegate.fromJson(reader).data;
+    }
+
+    @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      delegate.toJson(new Envelope<>(value));
+    }
+  }
+
+  public static Set<? extends Annotation> delegateAnnotations(Set<? extends Annotation> annotations,
```
NightlyNexus: This utility mechanism is pretty distracting for the sample. Is there a better way to do this?

NightlyNexus:
>I think we probably want a mechanism on Moshi where you can specify an annotation using a type literal Enveloped.class and just expand it out to the only reasonable implementation. I can do that, or you can read about dynamic proxies and annotation’s hashCode() method if you wanna try it.

I'm going to try to get this going tonight; sounds exciting. Thanks!

examples/src/main/java/com/squareup/moshi/recipes/TypeFailSafeJsonAdapterFactory.java
NightlyNexus: interestingly, seems like this fellow needed something like this, too, sadly: https://github.com/square/moshi/issues/188

examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java
swankjesse: Should we promote this to a helper method on `Types` ?

```
@Override public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
  Set<? extends Annotation> nextAnnotations = Types.nextAnnotations(annotations, Enveloped.class);
  if (nextAnnotations == null) return null;
  JsonAdapter<?> next = moshi.nextAdapter(this, type, nextAnnotations);
  ...
}
```

swankjesse: Docs would be a little weird: _“Checks if {@code jsonQualifier} is in the set, and if it is returns a set without it. Otherwise returns null.“_

NightlyNexus: Yeah, the weirdness is that this is both a `hasJsonQualifier` and `nextAnnotations`, as you said.
At the same time, this seems to be the best way to accomplish this.
I'll send a PR and discuss there.

swankjesse:
I really like where this landed. Thanks for driving API improvements through sample code!

adriancole:
really cool

NightlyNexus:
Yeah!
Thanks for the feedback and patience with the changes!

PR #237, opened 2017-01-29
Add Types.createJsonQualifierImplementation

NightlyNexus:
From https://github.com/square/moshi/pull/236.
I don't know what this method should be called yet.

This is way too easy, so I am misunderstanding something. Can I get a nudge in the right direction?

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -61,6 +61,11 @@ public final class Moshi {
     return adapter(type, Util.NO_ANNOTATIONS);
   }
 
+  public <T> JsonAdapter<T> adapter(Type type, Class<? extends Annotation> annotationType) {
+    return adapter(type,
+        Collections.singleton(Types.createJsonQualifierImplementation(annotationType)));
+  }
+
   @SuppressWarnings("unchecked") // Factories are required to return only matching JsonAdapters.
   public <T> JsonAdapter<T> adapter(Type type, Set<? extends Annotation> annotations) {
     type = Types.canonicalize(type);
--- a/moshi/src/main/java/com/squareup/moshi/Types.java
+++ b/moshi/src/main/java/com/squareup/moshi/Types.java
@@ -15,10 +15,14 @@
  */
 package com.squareup.moshi;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Array;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.GenericDeclaration;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.lang.reflect.WildcardType;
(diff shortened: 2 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -137,6 +141,38 @@ static Type canonicalize(Type type) {
     }
   }
 
+  public static <T extends Annotation> T createJsonQualifierImplementation(
+      final Class<T> annotationType) {
```
swankjesse: Confirm that it’s an Annotation? `annotationType.isAnnotation()`

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -137,6 +141,38 @@ static Type canonicalize(Type type) {
...
+        new Class<?>[] { annotationType }, new InvocationHandler() {
+          @Override public Object invoke(Object proxy, Method method, Object[] args)
+              throws Throwable {
+            String methodName = method.getName();
+            if (methodName.equals("annotationType")) {
+              return annotationType;
+            }
+            if (methodName.equals("equals")) {
+              Object o = args[0];
+              return o.getClass().isInstance(annotationType);
```
swankjesse: dig

swankjesse: oh, by the way `o` can be null!

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -137,6 +141,38 @@ static Type canonicalize(Type type) {
...
+            }
+            if (methodName.equals("hashCode")) {
```
swankjesse: dig

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -137,6 +141,38 @@ static Type canonicalize(Type type) {
...
+              return 0;
+            }
+            if (methodName.equals("toString")) {
+              return "@" + annotationType.getName() + "()";
+            }
+            throw new AssertionError();
```
swankjesse: I think you probably want just `return method.invoke(this, args)`. Or at least that’s what I’ve done elsewhere.

swankjesse:
Yep, this is pretty much exactly what I want. But never satisfied, there’s a few more things I want.

The way I’m imagining this being exposed is not as a factory method for an annotation; but instead as an overload of `Moshi.adapter()`.

```
JsonAdapter<String> uppercaseAdapter = moshi.adapter(String.class, Uppercase.class);
```

The generated annotation instance will be used to resolve the type adapter.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -137,6 +141,38 @@ static Type canonicalize(Type type) {
...
+    if (annotationType.getDeclaredMethods().length != 0) {
+      throw new IllegalArgumentException(annotationType + " must not declare methods.");
+    }
+    // noinspection unchecked
+    return (T) Proxy.newProxyInstance(annotationType.getClassLoader(),
+        new Class<?>[] { annotationType }, new InvocationHandler() {
+          @Override public Object invoke(Object proxy, Method method, Object[] args)
+              throws Throwable {
+            String methodName = method.getName();
+            if (methodName.equals("annotationType")) {
```
JakeWharton: Use `switch`?

moshi/src/main/java/com/squareup/moshi/Types.java
NightlyNexus: Yep!
Why `this` instead of `proxy`?

swankjesse: I think `proxy` would just stack overflow?

NightlyNexus:
>as an overload of Moshi.adapter()

Cool, that's what I thought at first, but then I was unsure because:
- Should `nextAdapter` also have an overload?
- Should it be `adapter(Type, Set<Class<? extends Annotation>>)` or `adapter(Type, Class<? extends Annotation>)` The first won't be a valid overload (type erasure), and the second seems a little limiting or at least different from the current method.

It sounds like the answer is no and the latter, respectively, but I wanted to confirm.

swankjesse:
I think the only one we need is `adapter(Type, Class<? extends Annotation>)`. I think it’s unlikely anyone will ever need multiple annotations for their entry point type.

NightlyNexus:
Sounds good, I'll make tests, then!

NightlyNexus:
Not really sure what a test on the `adapter` method should do, if necessary.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -61,6 +61,11 @@
     return adapter(type, Util.NO_ANNOTATIONS);
   }
 
+  public <T> JsonAdapter<T> adapter(Type type, Class<? extends Annotation> annotationType) {
+    return adapter(type,
+        Collections.singleton(Types.createJsonQualifierImplementation(annotationType)));
```
swankjesse: nice

PR #238, opened 2017-01-30
New JsonAdapter.serializeNulls() method.

swankjesse:
This makes it possible to force nulls into the document without
much fuss.


NightlyNexus:
Yay! https://github.com/square/retrofit/blob/dec2f1a675a77c12bd28f8f9a25a45a9cc31888d/retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java#L94 will like this.

PR #239, opened 2017-01-30
Add Flatten JsonAdapter.Factory sample

NightlyNexus:
This is https://github.com/square/moshi/pull/236 but without the confusion of the also having a `@JsonQualifier`.

But, maybe this discourages people from simply using `@FromJson` and `@ToJson` and having Moshi create the factory.
Is this useful?


swankjesse:
Yeah, I think this is one of the cases where `@FromJson` is a better fit. That said, I really do like the idea of having a generic example of an envelope. 

I think the right fix there is to make annotations easier.

NightlyNexus:
👍 Cool.

PR #242, opened 2017-01-31
Add Types.nextAnnotations

NightlyNexus:
This is useful for JsonAdapter factories.
From https://github.com/square/moshi/pull/236/files#r98377813

The method is a little weird because it is both `hasJsonQualifier` (null means false) and `nextAnnotations`.

~~Maybe this method should not also be `hasJsonQualifier`, though. Maybe this method should just throw an IllegalArgumentException instead of returning null?
The combination of these two methods is not a real performance (`annotations.contains(jsonQualifier)` is O(1)) or simplicity gain.~~
Edit: wasn't thinking, forgetting that we don't have the jsonQualifier instance.


swankjesse:
I think this is great. It won’t be used in a lot of places, but where it is used it’ll fit neatly.

PR #243, opened 2017-01-31
Add tests for duplicate JSON keys

NightlyNexus:
Should `duplicateKeysAllowedInCustomType()` fail, though?


swankjesse:
The code required to make it fail dissuades me from doing the work. We’d need another datastructure to track which fields we’d set and I don’t think the cost is worth the benefit.

NightlyNexus:
Yep, makes sense! 🚚

PR #244, opened 2017-02-01
Get opinionated about number types in toJsonObject().

swankjesse:
Previously we'd retain whatever types the caller passed in. This was
potentially problematic for non-immutable numeric types like AtomicInteger.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ObjectJsonWriter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ObjectJsonWriter.java
@@ -16,17 +16,26 @@
 package com.squareup.moshi;
 
 import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import static com.squareup.moshi.JsonScope.EMPTY_ARRAY;
 import static com.squareup.moshi.JsonScope.EMPTY_DOCUMENT;
 import static com.squareup.moshi.JsonScope.EMPTY_OBJECT;
 import static com.squareup.moshi.JsonScope.NONEMPTY_DOCUMENT;
+import static java.lang.Double.NEGATIVE_INFINITY;
+import static java.lang.Double.POSITIVE_INFINITY;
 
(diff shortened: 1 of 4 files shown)


moshi/src/main/java/com/squareup/moshi/ObjectJsonWriter.java
```diff
@@ -131,7 +140,20 @@ public Object root() {
   }
 
   @Override public JsonWriter value(double value) throws IOException {
-    return value(Double.valueOf(value));
+    long longValue = (long) value;
+    if (Double.compare(longValue, value) == 0) {
+      return value(longValue);
```
swankjesse: despite having prepared this PR I’m still not 100% convinced that this is the best behavior.

If a caller emits a sequence `[0.0d, 0.5d, 1.0d, 1.5d]` we’ll create JSON of `[0L, 0.5d, 1L, 1.5d]`.

It might be simpler to only do clever stuff in `value(Number)`, and to limit the clever stuff to promoting to a preferred type (Long over Integer, Double over Float). In either case this stuff is tricky.

swankjesse: The reason I care is that `0L` doesn’t equal `0.0d`, and that can be pretty annoying

JakeWharton: I'm inclined to agree here. Seems like if you have a double, even one without a floating part, then we should retain it as a double.

moshi/src/main/java/com/squareup/moshi/ObjectJsonWriter.java
```diff
@@ -144,20 +166,49 @@ public Object root() {
...
+    // Everything else gets converted to a BigDecimal, which itself might convert to a long.
+    BigDecimal bigDecimalValue = value instanceof BigDecimal
+        ? ((BigDecimal) value)
+        : new BigDecimal(value.toString());
+    try {
+      return value(bigDecimalValue.longValueExact());
+    } catch (ArithmeticException lossyConversion) {
+    }
+
+    // We're stuck with a big decimal because the value is too big or has a fractional part Put it
```
JakeWharton: nit: missing period between sentences

swankjesse:
Killed! Let’s do this instead:
https://github.com/square/moshi/pull/244

PR #246, opened 2017-02-02
Use BigDecimal to encode exotic number types.

swankjesse:
Previously we'd retain whatever types the caller passed in. This was
potentially problematic for non-immutable numeric types like AtomicInteger.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -18,6 +18,7 @@ package com.squareup.moshi;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.math.BigDecimal;
 import java.util.Set;
 import okio.Buffer;
 import okio.BufferedSink;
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -56,7 +57,16 @@ public final String toJson(T value) {
 
   /**
    * Encodes {@code value} as a Java model comprised of maps, lists, strings, numbers, booleans
-   * and nulls. The returned model is equivalent to calling {@link #toJson} to encode {@code value}
+   * and nulls.
+   * 
+   * <p>Values encoded using {@linkplain JsonWriter#value(double) value(double)} or {@linkplain
```
JakeWharton: Should these be normal `{@link .. ..}` since they refer to method signatures?

swankjesse: I wanna draw attention to the arguments not the method names and blue links seem distracting. might even be better as `{@code }`?

JakeWharton: Sure!

PR #247, opened 2017-02-03
Rename toJsonObject() to toJsonValue(), fromJsonObject() to fromJsonV…

swankjesse:
…alue().

Also rename the JsonReader and JsonWriter implementations.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUt8Writer.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUt8Writer.java
@@ -27,7 +27,7 @@ import static com.squareup.moshi.JsonScope.NONEMPTY_ARRAY;
 import static com.squareup.moshi.JsonScope.NONEMPTY_DOCUMENT;
 import static com.squareup.moshi.JsonScope.NONEMPTY_OBJECT;
 
-final class BufferedSinkJsonWriter extends JsonWriter {
+final class JsonUt8Writer extends JsonWriter {
 
   /*
    * From RFC 7159, "All Unicode characters may be placed within the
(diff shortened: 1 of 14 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUt8Writer.java
```diff
@@ -27,7 +27,7 @@
 import static com.squareup.moshi.JsonScope.NONEMPTY_DOCUMENT;
 import static com.squareup.moshi.JsonScope.NONEMPTY_OBJECT;
 
-final class BufferedSinkJsonWriter extends JsonWriter {
+final class JsonUt8Writer extends JsonWriter {
```
stevecstian: You meant `JsonUtf8Writer`, right?

moshi/src/test/java/com/squareup/moshi/JsonUt8WriterTest.java
```diff
@@ -21,7 +21,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public final class BufferedSinkJsonWriterTest {
+public final class JsonUt8WriterTest {
```
stevecstian: You meant `JsonUtf8WriterTest`, right?

moshi/src/main/java/com/squareup/moshi/JsonUt8Writer.java
swankjesse: YEP. Ouch.

swankjesse: https://github.com/square/moshi/pull/249

PR #248, opened 2017-02-03
Example adapter that recovers from JsonDataExceptions.

swankjesse:


examples/src/main/java/com/squareup/moshi/recipes/DefaultOnDataMismatchAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+    this.delegate = delegate;
+    this.defaultValue = defaultValue;
+    this.objectAdapter = objectAdapter;
+  }
+
+  @Override public T fromJson(JsonReader reader) throws IOException {
+    // Read the value first so that the reader will be in a known state even if there's an
+    // exception. Otherwise it may be awkward to recover: it might be between calls to
+    // beginObject() and endObject() for example.
+    Object jsonValue = objectAdapter.fromJson(reader);
```
swankjesse: ... this is where `reader.readJsonValue()` would slot in if we offered that API

JakeWharton: This definitely makes me want `reader.readJsonValue()`

PR #250, opened 2017-02-03
Add JsonReader.readJsonValue

NightlyNexus:
Moves `readJsonValue` up to `JsonReader`.
Just want to start the discussion.
The docs need improvement.

Is a `JsonWriter.writeJsonValue(Object value)` useful?

Also, I think I should move `defaultIfDataException` to a different change, if this is something that you'd like to add at all.


JakeWharton:
These should at the very least be separate commits, but more probably separate PRs for discussion.

swankjesse:
Nice.

PR #253, opened 2017-02-04
Update example with readJsonValue

NightlyNexus:


NightlyNexus:
oops, I didn't look first. https://github.com/square/moshi/pull/252

PR #254, opened 2017-02-04
Remove package API from README

NightlyNexus:
Types.createJsonQualifierImplementation() is not in the public API.


swankjesse:
Good eye. That's just me being sloppy!

Issue #218, opened 2016-11-23
Returning unknown enum values as null


edenman:
I have an enum that represents the various different types of errors i could get for a particular server endpoint.  If i get back a value that isn’t in that enum (say, the server adds a new error type and I’m on an old client that doesn’t know about it), `EnumJsonAdapter` blows up.  I'd like to add an option to have it just return null in that case.  Thoughts?  How would I go about adding this? (I don't see any customization of the `StandardJsonAdapters` but I might just be missing it)

edenman:
Or is the answer just "add your own enum adapter Factory"?

swankjesse:
That's the easiest. We should probably include that as a standard extra adapter.

edenman:
Cool, I'll just roll my own for now, thanks!

serj-lotutovici:
Sorry for bragging, but there is a [`@FallbackEnum` annotation](https://github.com/serj-lotutovici/moshi-lazy-adapters/blob/master/src/main/java/com/serjltt/moshi/adapters/FallbackEnum.java) in a small extension I've wrote. It allows to specify a fallback value (spares a null check). The used adapter is a fork of the internal one.

NightlyNexus:
for posterity:
https://github.com/square/moshi/pull/248

NightlyNexus:
@edenman and everybody else,
Is this okay to close, given the above sample? Is it alright not to have this fallback in the JsonAdapter API?

edenman:
I'm ok with the "roll your own"/"use the sample" approach, but it does still seem like something that is a common enough occurrence that it'd be good to build in (at least as an option).  Not a big deal either way though, I'm already using my own bespoke adapter.

swankjesse:
Yep. I think we want something specifically for enums.

naturalwarren:
FWIW I currently roll my own "enum fallback adapters". Would love to see this get added.

NightlyNexus:
@swankjesse Are you thinking something like `Moshi.Builder.add(Class<T> enumType, T defaultValue)`? That would be trivial to implement, but maybe there's a better API I'm not thinking of yet.

NightlyNexus:
Or, `Builder.enumFallback(T fallbackValue)` and skip the class parameter?

You wouldn't be able to fall back to null, then, but that might be okay. Force people who need a fallback to define a default or a special "UNKNOWN" type instead of null.

NightlyNexus:
[...]

PR #255, opened 2017-02-05
Add JsonAdapter.defaultIfDataException(T)

NightlyNexus:
https://github.com/square/moshi/issues/218
Enums are a good use case. Other cases are unfortunate workarounds, and I'm not yet confident this is great to generalize.

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
@@ -16,11 +16,14 @@
 package com.squareup.moshi;
 
 import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
```diff
@@ -164,4 +167,71 @@
...
+        return pizzaAdapter.defaultIfDataException(new Pizza(-1));
+      }
+    };
+    Moshi moshi = new Moshi.Builder().add(pizzaFactory).build();
+    String json = "{\"pizza\":{\"price\":\"eighty\"}, \"drink\":\"coke\"}";
+    JsonAdapter<MealDeal> adapter = moshi.adapter(MealDeal.class);
+    assertThat(adapter.fromJson(json)).isEqualTo(new MealDeal(new Pizza(-1), "coke"));
+  }
+
+  private static final class Pizza {
```
NightlyNexus: This feels bad to bring in more lines of machinery than lines of tests. I wonder if TestUtils should contain some test value objects.

NightlyNexus:
This is straightforward enough to do one's self, especially given the [sample](https://github.com/square/moshi/blob/af09de142d9bb57057a80362104dcf7ff051db64/examples/src/main/java/com/squareup/moshi/recipes/DefaultOnDataMismatchAdapter.java).

PR #256, opened 2017-02-05
Add Polymorphic sample

NightlyNexus:
~This made me realize I think we do need `JsonWriter.writeJsonValue(Object)`.~
Never mind, obviously just use the delegate to make calls to the JsonWriter.

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/Polymorphic.java
+++ b/examples/src/main/java/com/squareup/moshi/recipes/Polymorphic.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi.recipes;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonDataException;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.recipes.models.BlackjackHand;
+import com.squareup.moshi.recipes.models.Card;
+import com.squareup.moshi.recipes.models.Player;
+import com.squareup.moshi.recipes.models.Suit;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+final class Polymorphic {
+  public static void main(String[] args) throws Exception {
+    Moshi moshi = new Moshi.Builder().add(PolymorphicBlackjackJsonAdapter.FACTORY)
+        .add(new CardAdapter())
+        .build();
+    String handJson = "{\n"
+        + "    \"kind\": \"blackjack-hand\",\n"
+        + "    \"data\": {\n"
+        + "      \"hidden_card\": \"6S\",\n"
+        + "      \"visible_cards\": [\n"
+        + "        \"4C\",\n"
+        + "        \"AH\"\n"
+        + "      ]\n"
+        + "    }\n"
+        + "  }";
+    String playerJson = "{\n"
+        + "    \"kind\": \"player\",\n"
+        + "    \"data\": {\n"
+        + "      \"username\": \"jesse\",\n"
+        + "      \"lucky number\": 32\n"
+        + "    }\n"
+        + "  }";
+    JsonAdapter<BlackjackHand> handAdapter = moshi.adapter(BlackjackHand.class);
+    JsonAdapter<Player> playerAdapter = moshi.adapter(Player.class);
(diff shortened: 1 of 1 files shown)


examples/src/main/java/com/squareup/moshi/recipes/Polymorphic.java
```diff
@@ -0,0 +1,126 @@
...
+          throw new JsonDataException("Unexpected type: " + type);
+      }
+    }
+
+    @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      Class<?> cls = value.getClass();
+      writer.beginObject();
+      writer.name("kind");
+      String type;
+      JsonAdapter adapter;
```
JakeWharton: `JsonAdapter<Object>`

NightlyNexus: Will the type system let me assign to that?
I'm mobile, so that's just my quick glance.

JakeWharton: It should. Every JsonAdapter<Foo> is a JsonAdapter<Object>

NightlyNexus: I don't think I can assign a `JsonAdapter<Foo>` to the parameterized `JsonAdapter<Object>` type.
Kotlin can do that.

JayNewstrom: You should be able to use `JsonAdapter<?>`.

JayNewstrom: Well, if you do that, the `adapter.toJson` call doesn't compile. If you're trying to avoid using raw types, you can get around it with something similar to:
```java
  Class<?> cls = value.getClass();
  writer.beginObject();
  writer.name("kind");
  if (cls == BlackjackHand.class) {
      writeTypeAndValueJson(writer, "blackjack-hand", handAdapter, value);
  } else if (cls == Player.class) {
      writeTypeAndValueJson(writer, "player", playerAdapter, value);
  } else {
    throw new AssertionError(cls + " is not BlackJackHand.class or Player.class.");
  }
  writer.endObject();
}

private static <T> void writeTypeAndValueJson(JsonWriter writer, String type, JsonAdapter<T> adapter, Object value) throws IOException {
    writer.value(type);
    writer.name("data");
    //noinspection unchecked
    adapter.toJson(writer, (T) value);
}
```

NightlyNexus: sure.
Jake, any opinion on what's clearest here: put JsonWriter calls in the conditional branches or leave as-is?

NightlyNexus:
@JakeWharton Is there anything I can do to make the above clearer, or should I merge this?

examples/src/main/java/com/squareup/moshi/recipes/Polymorphic.java
```diff
@@ -0,0 +1,126 @@
...
+    private final JsonAdapter<Player> playerAdapter;
+
+    PolymorphicBlackjackJsonAdapter(JsonAdapter<BlackjackHand> handAdapter,
+        JsonAdapter<Player> playerAdapter) {
+      this.handAdapter = handAdapter;
+      this.playerAdapter = playerAdapter;
+    }
+
+    @Override public Object fromJson(JsonReader reader) throws IOException {
+      // noinspection unchecked
```
swankjesse: I wonder if we should do a checked cast here to set a good example. That way a bad JSON fails with a JsonDataException (expected) instead of a ClassCastException (unexpected)

examples/src/main/java/com/squareup/moshi/recipes/Polymorphic.java
```diff
@@ -0,0 +1,126 @@
...
+      Map<String, Object> value = (Map<String, Object>) reader.readJsonValue();
+      String type = (String) value.get("kind");
```
swankjesse: similarly, defensive programming suggests we don’t know if this is going to be null or not.

I guess the main challenge is we want to simultaneously demonstrate best practices while also being simple and easy to understand. Tricky.

NightlyNexus:
Getting something in the adapters module would be great. I totally agree that it is easy to slip up, and it is tedious to write tests for every polymorphic adapter.

The abstraction might be even trickier.
Not every structure looks like this one.
Some people have to check the type of everything read in from `readJsonValue` (like: `["string", {}, "string2", {}]`).

The adapter could specifically only cover fairly simple cases like the one here, perhaps.

Maybe this should go in a new GH issue; close this?

swankjesse:
Unsure. What do you think of the way Gson does it?
https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java#L35

NightlyNexus:
I think that is reasonable and covers a lot of use cases.

A lot of the docs there reference subtyping, but that's not really a requirement. (I've played with the term "polymorphic" loosely.)
Other than that, it sounds like a good idea.

I'm unsure what you have in mind for the adapters module.
Is a versioned general utilities artifact a good idea as opposed to focused sample code or focused separate libraries?

swankjesse:
I think it’d be a thing where you configure names + types and that’s it.

NightlyNexus:
How about a `@RuntimeType` qualifier in place of sub types?

swankjesse:
Hmmm. I'm unsure. You'd need to specify the type in two places.

NightlyNexus:
I'm leaning back towards requiring inheritance.

The problem is nobody could qualify type arguments without Java 8.
The people who need to do List<@RuntimeType Object> because the types are final platform or library types will either have to roll this manually and use Java 8 or extend ArrayList. (I've had to do the the latter in the past with Gson.)

We'll have to make people do List<ParentMarkerType>, as far as I see.

NightlyNexus:
Moving to https://github.com/square/moshi/pull/264

PR #257, opened 2017-02-05
Fix README version number

NightlyNexus:
Oops, I missed this yesterday.


swankjesse:
It was I who missed. Thanks!

PR #258, opened 2017-02-06
Add DeferredAdapter recursion test

NightlyNexus:
I couldn't find an existing test that demonstrated DeferredAdapter's usage; let me know if I missed one.

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/MoshiTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/MoshiTest.java
@@ -995,6 +995,39 @@ public final class MoshiTest {
     }
   }
 
+  @Test public void deferredAdapterRecursesInfinitely() throws Exception {
+    JsonAdapter.Factory faulty = new JsonAdapter.Factory() {
+      @Override
+      public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+        if (type != Boolean.class) return null;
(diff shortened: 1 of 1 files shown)


moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -995,6 +995,39 @@
     }
   }
 
+  @Test public void deferredAdapterRecursesInfinitely() throws Exception {
```
swankjesse: Deferring StackOverflowErrors isn’t really what we’re going for with this feature.

Instead imagine you’re building an adapter for a binary tree:
```
class TreeNode {
  String value;
  TreeNode left;
  TreeNode right;
}
```
The standard reflective adapter for this is going to recursively look itself up, and it’s going to mostly just work.

swankjesse: And I think we cover this in CircularAdaptersTest. I think those tests will fail if you comment out the deferred adapter code.

NightlyNexus: Right.
Sorry, I was reasoning through to myself a few days ago what would happen in this case, as someone was talking to me about not being able to use `nextAdapter` in his factory.
This doesn't make sense as a test.

NightlyNexus: Oh, I'll give those tests a try. thanks!

NightlyNexus:
👍  CircularAdaptersTest tests.

PR #260, opened 2017-02-14
Fix error-prone warning

NightlyNexus:
I think this is what error-prone is upset about.


JakeWharton:
Added to my PR which was based on a really old SHA.

NightlyNexus:
Ha, I noticed that! 👍

PR #264, opened 2017-02-25
Add a RuntimeJsonAdapterFactory

NightlyNexus:
Wrote up a quick thing to get discussion started. The names and messages probably need improvement, this should probably have tests somewhere, etc.

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/RuntimeJsonAdapterFactory.java
+++ b/examples/src/main/java/com/squareup/moshi/recipes/RuntimeJsonAdapterFactory.java
@@ -1,0 +1,129 @@
+package com.squareup.moshi.recipes;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonDataException;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.Types;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+public final class RuntimeJsonAdapterFactory implements JsonAdapter.Factory {
+  private final Map<Class<?>, String> subtypeToLabel = new LinkedHashMap<>();
+  private final Class<?> baseType;
+  private final String labelKey;
+
+  public RuntimeJsonAdapterFactory(Class<?> baseType, String labelKey) {
+    if (baseType == null) {
+      throw new NullPointerException("baseType == null");
+    }
+    if (labelKey == null) {
+      throw new NullPointerException("labelKey == null");
+    }
+    this.baseType = baseType;
+    this.labelKey = labelKey;
+  }
+
+  public RuntimeJsonAdapterFactory registerSubtype(Class<?> subtype, String label) {
+    if (subtype == null) {
+      throw new NullPointerException("subtype == null");
+    }
+    if (label == null) {
+      throw new NullPointerException("label == null");
+    }
+    if (!baseType.isAssignableFrom(subtype)) {
+      throw new IllegalArgumentException(subtype + " must be a " + baseType);
+    }
+    subtypeToLabel.put(subtype, label);
+    return this;
+  }
+
+  @Override
+  public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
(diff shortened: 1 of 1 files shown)


examples/src/main/java/com/squareup/moshi/recipes/RuntimeJsonAdapterFactory.java
```diff
@@ -0,0 +1,82 @@
...
+              + label.getClass());
+        }
+        JsonAdapter<?> delegate = labelToDelegate.get(label);
+        if (delegate == null) {
+          throw new JsonDataException("Type not registered for label: " + label);
+        }
+        return delegate.fromJsonValue(value);
+      }
+
+      @Override public void toJson(JsonWriter writer, Object value) throws IOException {
```
NightlyNexus: Maybe it'd be nice if this would set the value of the label on the value. But, I think the best way would include creating the JSON value Map and then serializing the Map. That wouldn't be worth it at all.

NightlyNexus: Actually, it might be worth it for consistent behavior with the RuntimeTypeAdapterFactory for Gson.

examples/src/main/java/com/squareup/moshi/recipes/RuntimeJsonAdapterFactory.java
```diff
@@ -0,0 +1,129 @@
...
+
+  public RuntimeJsonAdapterFactory registerSubtype(Class<?> subtype, String label) {
+    if (subtype == null) {
+      throw new NullPointerException("subtype == null");
+    }
+    if (label == null) {
+      throw new NullPointerException("label == null");
+    }
+    if (!baseType.isAssignableFrom(subtype)) {
+      throw new IllegalArgumentException(subtype + " must be a " + baseType);
```
NightlyNexus: Or use generics to make this a compile-time failure. (`Class<? extends T> subtype`). Not sure.

examples/src/main/java/com/squareup/moshi/recipes/RuntimeJsonAdapterFactory.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override public Object fromJson(JsonReader reader) throws IOException {
+      Object raw = reader.readJsonValue();
+      if (!(raw instanceof Map)) {
+        throw new JsonDataException(
+            "Value must be a JSON object but had a value of " + raw + " of type " + raw.getClass());
+      }
+      @SuppressWarnings("unchecked") // This is a JSON object.
+          Map<String, Object> value = (Map<String, Object>) raw;
+      Object label = value.remove(labelKey);
```
spotlight-david: Would you mind it if get was used instead of remove here? In my use case I want to keep the labelkey to use it later.

JakeWharton: It's just a sample so you're free to modify it as you wish.

NightlyNexus: Yep, this is only to emulate Gson's RuntimeTypeAdapterFactory: https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java#L202

Copy, paste, modify to your needs.

NightlyNexus:
I'm closing in favor of blog posts.

robstoll:
@NightlyNexus where do I find the blog post?

outlying:
@NightlyNexus bump, I was looking for this support, it's strange that example not in the repository anymore

robstoll:
@outlying I came up with the following solution: 
https://github.com/loewenfels/dep-graph-releaser/blob/master/dep-graph-releaser-serialization/src/main/kotlin/ch/loewenfels/depgraph/serialization/PolymorphicAdapterFactory.kt

Notice, that it is faster than the solution usually showed here but relies on the order of the properties.

NightlyNexus:
You probably want to cache the adapters instead of looking them up on every fromJson call.

NightlyNexus:
The extra hashmap lookup is unnecessary work.

robstoll:
Well, I would need to administer a HashMap as well, because the runtime types are not known during compilation. Thus I prefer there is just one cache.

outlying:
@robstoll thanks, I will look at this solution, I think that there will be no problem to take into account the different order of properties, it is generally strange for me that Moshi does not offer / suggests solving the problem of deserializing polymorphic data

PR #265, opened 2017-03-03
expecting EncodingException with beginArray() when malformed json

danglotb:
Hi,

I propose to had the following changes to expect a `JsonEncodingException` while the flag lenient is false.


NightlyNexus:
The tests in JsonUtf8ReaderTest cover this pretty well (like https://github.com/square/moshi/blob/81aed974c6b2ebf98e9ade500cd5c4e435b20ec1/moshi/src/test/java/com/squareup/moshi/JsonUtf8ReaderTest.java#L554), although they don't check the exception message. Maybe update one or so of them with the exception message instead?

PR #268, opened 2017-03-09
mUser class factories

cbruegg:


cbruegg:
Disregard this PR, it was sent by accident.

PR #269, opened 2017-03-20
Add kotlin-module with support for Kotlin data classes

cbruegg:
This PR adds support for properly deserializing Kotlin data classes. Instead of assigning fields, the `KotlinJsonAdapter` invokes the primary constructor as proposed [here](https://github.com/square/moshi/pull/266#issuecomment-287724888).

To activate this behavior, a library user will need to add a dependency on `moshi-kotlin` and add the `KotlinJsonAdapterFactory` to Moshi. Thus, no existing code using Moshi will break after this PR is merged.

Thanks for your patience. I haven't contributed to many open source projects yet, so any feedback is appreciated.

### DIFF
--- a/kotlin/pom.xml
+++ b/kotlin/pom.xml
@@ -11,10 +11,6 @@
 
   <artifactId>moshi-kotlin</artifactId>
 
-  <properties>
-    <kotlin.version>1.1.1</kotlin.version>
-  </properties>
-
   <dependencies>
     <dependency>
       <groupId>com.squareup.moshi</groupId>
@@ -34,17 +30,14 @@
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
       <artifactId>kotlin-stdlib</artifactId>
-      <version>${kotlin.version}</version>
     </dependency>
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
       <artifactId>kotlin-reflect</artifactId>
-      <version>${kotlin.version}</version>
     </dependency>
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
       <artifactId>kotlin-test</artifactId>
-      <version>${kotlin.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
@@ -54,7 +47,6 @@
       <plugin>
         <groupId>org.jetbrains.kotlin</groupId>
         <artifactId>kotlin-maven-plugin</artifactId>
-        <version>${kotlin.version}</version>
         <executions>
           <execution>
             <id>compile</id>
@@ -94,6 +86,5 @@
       </plugin>
     </plugins>
   </build>
-
 
 </project>
--- a/kotlin/pom.xml
+++ b/kotlin/pom.xml
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.squareup.moshi</groupId>
+    <artifactId>moshi-parent</artifactId>
+    <version>1.5.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>moshi-kotlin</artifactId>
+
+  <properties>
+    <kotlin.version>1.1.1</kotlin.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.squareup.moshi</groupId>
+      <artifactId>moshi</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-stdlib</artifactId>
+      <version>${kotlin.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-reflect</artifactId>
+      <version>${kotlin.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-test</artifactId>
+      <version>${kotlin.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.jetbrains.kotlin</groupId>
+        <artifactId>kotlin-maven-plugin</artifactId>
+        <version>${kotlin.version}</version>
+        <executions>
+          <execution>
+            <id>compile</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>test-compile</id>
+            <phase>test-compile</phase>
+            <goals>
+              <goal>test-compile</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>compile</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>testCompile</id>
+            <phase>test-compile</phase>
+            <goals>
+              <goal>testCompile</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+
+</project>
--- a/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
@@ -2,9 +2,8 @@
 
 package com.squareup.moshi
 
-import com.squareup.moshi.ClassJsonAdapter.includeField
-import com.squareup.moshi.ClassJsonAdapter.isPlatformType
-import java.lang.reflect.Modifier
+import com.squareup.moshi.ClassJsonAdapter.*
+import java.lang.reflect.Field
 import java.lang.reflect.Type
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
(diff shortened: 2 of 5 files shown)


kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * - Any properties not found in the primary constructor must be transient, including delegated properties.
+ * - The primary constructor may not contain transient properties unless a default value is provided.
+ * - If the class is a platform type (kotlin.*), it may not contain private fields.
+ */
+class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+  override fun create(type: Type, annotations: Set<Annotation>, moshi: Moshi): JsonAdapter<*>? {
+    val kclass = (type as? Class<*>)?.kotlin ?: return null
+    if (!kclass.isData) return null
+    // This gives us a few guarantees:
```
swankjesse: nice

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+    constructor.isAccessible = true
+
+    val concretePropertyByName = kclass.declaredMemberProperties
+        .filter { it.javaField != null }
+        .associateBy { it.name }
+    val constructorParameterByName = constructor.parameters.associateBy { it.name }
+    for ((name, property) in concretePropertyByName) {
+      if (name !in constructorParameterByName.keys &&
+          !Modifier.isTransient(property.javaField!!.modifiers)) {
+        throw IllegalArgumentException("Property $name is not transient. " +
```
swankjesse: (this can remain as an IllegalArgumentException. JsonDataException is only for mismatches between data and schema)

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+    val notIncludedParams = constructorParameterByName
+        .filterNot { it.value.isOptional }
+        .keys.map { concretePropertyByName[it]!! }
+        .filter { !includeField(isPlatformType, it.javaField!!.modifiers) }
+    if (notIncludedParams.isNotEmpty()) {
+      throw IllegalArgumentException("Primary constructor contains parameters " +
+          "not found in serialized JSON: $notIncludedParams")
+    }
+
+    val toJsonAdapter = ClassJsonAdapter.FACTORY.create(type, annotations, moshi)
```
swankjesse: Not sure how I feel about delegating here. Seems like it might be simpler to just do it all in Kotlin

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+    val propertyByParam = constructor.parameters.associate { it to concretePropertyByName[it.name]!! }
+    val adapterByParam = constructor.parameters.associate { param ->
+      val javaField = propertyByParam[param]!!.javaField!!
+      val paramType = Types.resolve(javaField.type, Types.getRawType(javaField.type), javaField.genericType)
+      val annotations = param.annotations.toSet()
+
+      param to moshi.adapter<Any>(paramType, annotations)
```
swankjesse: off topic: implicit returns are weird

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+    private val toJsonAdapter: JsonAdapter<T>,
+    private val propertyByParam: Map<KParameter, KProperty1<out Any, Any?>>,
+    private val adapterByParam: Map<KParameter, JsonAdapter<*>>
+    ) : JsonAdapter<T>() {
+
+  private val jsonNames = constructor.parameters
+      .map { propertyByParam[it]!! }
+      .map { it.javaField!!.getAnnotation(Json::class.java)?.name ?: it.name }
+      .toTypedArray()
+  private val options = JsonReader.Options.of(*jsonNames)
```
swankjesse: I need to learn what `*` does

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+    return KotlinJsonAdapter(constructor, toJsonAdapter as JsonAdapter<Any?>, propertyByParam, adapterByParam)
+  }
+
+}
+
+@Suppress("LoopToCallChain")
+internal class KotlinJsonAdapter<T>(
+    private val constructor: KFunction<T>,
+    private val toJsonAdapter: JsonAdapter<T>,
+    private val propertyByParam: Map<KParameter, KProperty1<out Any, Any?>>,
```
swankjesse: Ideally for performance we avoid map lookups in the decode/encode paths. Otherwise this will be slower than Java and nobody wants that

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+  private val options = JsonReader.Options.of(*jsonNames)
+
+  override fun fromJson(reader: JsonReader): T {
+    try {
+      val valuesByParam = mutableMapOf<KParameter, Any?>()
+
+      reader.beginObject()
+      while (reader.hasNext()) {
+        val index = reader.selectName(options)
+        val param = if (index != -1) {
```
swankjesse: I don’t think I like this style

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+          continue
+        }
+
+        valuesByParam[param] = adapterByParam[param]!!.fromJson(reader)
+      }
+      reader.endObject()
+
+      for (param in constructor.parameters) {
+        if (!param.isOptional && param !in valuesByParam) {
+          throw IllegalArgumentException("JSON is missing value for ${param.name}")
```
swankjesse: JsonDataException

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
JakeWharton: s/weird/awful/

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
JakeWharton: explodes an array to varargs

kotlin/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+  <parent>
+    <groupId>com.squareup.moshi</groupId>
+    <artifactId>moshi-parent</artifactId>
+    <version>1.5.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>moshi-kotlin</artifactId>
+
+  <properties>
+    <kotlin.version>1.1.1</kotlin.version>
```
JakeWharton: This should really be in the root `pom.xml`

kotlin/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-stdlib</artifactId>
+      <version>${kotlin.version}</version>
```
JakeWharton: These three deps should be in `<dependencyManagement>` of the root `pom.xml` and declared without a version here.

kotlin/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.jetbrains.kotlin</groupId>
+        <artifactId>kotlin-maven-plugin</artifactId>
+        <version>${kotlin.version}</version>
```
JakeWharton: Similarly, in `<pluginManagement>`.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
cbruegg: In the ClassJsonAdapter, it basically was like this:

```
val param: KParameter
if (index != -1) {
  param = constructor.parameters[index]
} else {
  reader.nextName()
  reader.skipValue()
  continue
}
```

I'm not sure if that'd be preferable over `val param = if ...`. Seeing [this discussion](https://github.com/yole/kotlin-style-guide/issues/11) in the WIP Kotlin style guide, I think using the `if` expression should be fine. Let me know what you think, I can change it if you still don't like it.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,141 @@
...
+ * Moshi will call its primary constructor to create the object. This allows for delegated properties
+ * to be initialized correctly. If the JSON does not contain a value for a parameter with a default value
+ * the default value is used. If no default value is present an [IllegalArgumentException] will be
+ * thrown during deserialization.
+ *
+ * Data classes to be (de-)serialized must conform to the following rules:
+ *
+ * - Any properties not found in the primary constructor must be transient, including delegated properties.
+ * - The primary constructor may not contain transient properties unless a default value is provided.
+ * - If the class is a platform type (kotlin.*), it may not contain private fields.
```
cbruegg: > Any properties not found in the primary constructor must be transient, including delegated properties.
Thinking about this, I don't think it's a necessary restriction. Instead, these fields could be deserialized/serialized just like the `ClassJsonAdapter` does currently using `FieldBinding`s. I'll lift this restriction.

> The primary constructor may not contain transient properties unless a default value is provided.

If it contained a transient property `p` without a default value, Moshi couldn't call the constructor since it may not provide values for `p` due to the `Transient` annotation. Thus, in order for the constructor call to succeed even though the `valuesByParam` don't contain an entry `(p, ...)`, the constructor will need to contain a default value for `p`.

> If the class is a platform type (kotlin.*), it may not contain private fields.

Same restriction as in the `ClassJsonAdapter`.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
cbruegg: As mentioned above, `FieldBinding`s will be required anyway for properties that weren't defined in the primary constructor, so yup, I'll replace the delegation.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
cbruegg: Nevermind, this had to be changed for supporting non-transient fields not defined in the primary constructor anyway.

cbruegg:
Ok, I've implemented the requested changes. The adapter now also supports non-transient properties not defined in the primary constructor, including those from superclasses, which data classes may have since Kotlin 1.1.

I've eliminated the delegation to `ClassJsonAdapter`, though I've used its factory's method `createFieldBindings`, which is now static.

cbruegg:
Are there any other changes you'd like me to make to make this ready to merge?

swankjesse:
Sorry for dragging on this. Hoping to take a more thorough look soon.

swankjesse:
Folded into this PR: https://github.com/square/moshi/pull/281

PR #272, opened 2017-03-26
Allow easy delegates in adapter methods.

### LINKED ISSUE #271: Allow delegates for non-qualified types into adapter methods.
```java
final class PointJsonAdapter {
  @FromJson Point fromJson(JsonReader reader, JsonAdapter<Point> delegate) throws Exception {
    reader.beginObject();
    reader.nextName();
    Point point = delegate.fromJson(reader);
    reader.endObject();
    return point;
  }
}
```
For the above to work, I would need a `@JsonQualifier` on all of my application-layer `Point`s, or I would need to write the `JsonAdapter.Factory` manually.

Should Moshi support the above?

NightlyNexus:
Closes #271.

Are there any more test cases I might not be thinking of?

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
@@ -55,8 +55,8 @@ final class AdapterMethodsFactory implements JsonAdapter.Factory {
       delegate = null;
     }
 
-    if (toAdapter != null) toAdapter.bind(moshi);
-    if (fromAdapter != null) fromAdapter.bind(moshi);
+    if (toAdapter != null) toAdapter.bind(moshi, this);
+    if (fromAdapter != null) fromAdapter.bind(moshi, this);
 
     return new JsonAdapter<Object>() {
       @Override public void toJson(JsonWriter writer, Object value) throws IOException {
@@ -170,8 +170,8 @@ final class AdapterMethodsFactory implements JsonAdapter.Factory {
           parameterTypes.length, 1, nullable) {
         private JsonAdapter<Object> delegate;
 
-        @Override public void bind(Moshi moshi) {
-          super.bind(moshi);
+        @Override public void bind(Moshi moshi, JsonAdapter.Factory factory) {
+          super.bind(moshi, factory);
           delegate = moshi.adapter(returnType, returnTypeAnnotations);
         }
 
@@ -235,8 +235,8 @@ final class AdapterMethodsFactory implements JsonAdapter.Factory {
           parameterTypes.length, 1, nullable) {
         JsonAdapter<Object> delegate;
 
-        @Override public void bind(Moshi moshi) {
-          super.bind(moshi);
+        @Override public void bind(Moshi moshi, JsonAdapter.Factory factory) {
+          super.bind(moshi, factory);
           delegate = moshi.adapter(parameterTypes[0], qualifierAnnotations);
         }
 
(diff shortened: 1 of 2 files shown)


NightlyNexus:
tangential: Should we update the IllegalArgumentException messages in AdapterMethodsFactory to state that JsonAdapters can be used?

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
```diff
@@ -304,15 +343,15 @@
...
+    @FromJson int jsonToInt(JsonReader reader) throws IOException {
       if (reader.peek() == JsonReader.Token.NULL) {
         reader.nextNull();
         return -1;
       }
       return reader.nextInt();
     }
 
-    @ToJson void jsonToInt(JsonWriter writer, int value) throws IOException {
+    @ToJson void intToJson(JsonWriter writer, int value) throws IOException {
```
NightlyNexus: sneaking these name fixes in, if that's alright.

NightlyNexus:
You've always got to have the JsonReader and JsonWriter, so you can't do something like:
```
final class PointJsonAdapter {
  @FromJson Point fromJson(JsonAdapter<Point> delegate, String json) throws IOException {
    return delegate.fromJson(json);
  }
}
```
That's a good thing, right?

NightlyNexus:
>tangential: Should we update the IllegalArgumentException messages in AdapterMethodsFactory to state that JsonAdapters can be used?

Closed with #273.

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
```diff
@@ -285,14 +285,18 @@ public AdapterMethod(Type type, Set<? extends Annotation> annotations, Object ad
...
       if (jsonAdapters.length > 0) {
         Type[] parameterTypes = method.getGenericParameterTypes();
         Annotation[][] parameterAnnotations = method.getParameterAnnotations();
         for (int i = adaptersOffset, size = parameterTypes.length; i < size; i++) {
-          jsonAdapters[i - adaptersOffset] = moshi.adapter(
-              ((ParameterizedType) parameterTypes[i]).getActualTypeArguments()[0],
-              jsonAnnotations(parameterAnnotations[i]));
+          Type type = ((ParameterizedType) parameterTypes[i]).getActualTypeArguments()[0];
+          Set<? extends Annotation> jsonAnnotations = jsonAnnotations(parameterAnnotations[i]);
+          if (this.type == type && annotations.equals(jsonAnnotations)) {
```
swankjesse: this should probably be Types.equal(). In general neither `==` nor `.equals()` works on `java.lang.reflect.Type`. (for example, `String[].class` isn’t equal to the parameter of `List<String[]>.class`)

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
```diff
@@ -285,14 +285,18 @@ public AdapterMethod(Type type, Set<? extends Annotation> annotations, Object ad
...
+            jsonAdapters[i - adaptersOffset] = moshi.nextAdapter(factory, type, jsonAnnotations);
```
swankjesse: would this work better as a ternary assignment?

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
```diff
@@ -79,6 +79,45 @@
...
+
+  @Test public void toAndFromWithDelegate() throws Exception {
+    Moshi moshi = new Moshi.Builder().add(new PointJsonAdapterWithDelegate()).build();
+    JsonAdapter<Point> adapter = moshi.adapter(Point.class);
+    Point point = new Point(5, 8);
+    assertThat(adapter.toJson(point)).isEqualTo("{\"x\":5,\"y\":8}");
+    assertThat(adapter.fromJson("{\"x\":5,\"y\":8}")).isEqualTo(point);
+  }
+
+  @Test public void toAndFromWithDelegateWithQualifier() throws Exception {
```
swankjesse: would these tests pass if the behavior is wrong?

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
swankjesse: Yes please.

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
swankjesse: Maybe just customize the behavior somehow? Maybe wrap one of ’em in an array?

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
NightlyNexus: Oops, yep.
Is there a good way to add a test for this?

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
NightlyNexus: Yep!

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
NightlyNexus: 👍

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
swankjesse: Creating a JsonAdapter for `String[]` should do it. The reflection API is hazardous.

swankjesse:
Looks good, I’m merging. The `String[]` test is no big deal; the code is right.

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
NightlyNexus: 👍 #274.

PR #273, opened 2017-03-27
Update adapter methods ISE message.

NightlyNexus:
Lemme know if you have a better wording for the message.


swankjesse:
Nice!

moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java
```diff
@@ -186,6 +186,8 @@ static AdapterMethod toAdapter(Object adapter, Method method) {
       throw new IllegalArgumentException("Unexpected signature for " + method + ".\n"
           + "@ToJson method signatures may have one of the following structures:\n"
           + "    <any access modifier> void toJson(JsonWriter writer, T value) throws <any>;\n"
+          + "    <any access modifier> void toJson(JsonReader jsonReader,"
```
ChristianBecker: `JsonWriter`?

NightlyNexus: Dang.
Yep! Thanks.
It's also not totally complete, since you can have `JsonWriter, T, JsonAdapter, ...)`, too.

NightlyNexus: #275.

PR #279, opened 2017-04-17
Fix Types.equals for arrays.

NightlyNexus:
I get the feeling this was not supposed to work at some point, or I am misunderstanding something.


swankjesse:
This is perfect. Thanks!

PR #280, opened 2017-04-18
Minor improvements

cbruegg:
1. As `ABSENT_VALUE` is a singleton with its own type, it can be made an `object` ([docs](https://kotlinlang.org/docs/reference/object-declarations.html#object-declarations)).
2. Performance: `var allAnnotations = property.annotations` where `property.annotations` isn't mutable, calling `allAnnotations += ...` will create a copy of the immutable list. I've made it a `val` now by creating a mutable copy of `property.annotations`.

### DIFF
--- a/kotlin/pom.xml
+++ b/kotlin/pom.xml
@@ -86,5 +86,4 @@
       </plugin>
     </plugins>
   </build>
-
 </project>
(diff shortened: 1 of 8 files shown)


swankjesse:
Folded in! https://github.com/square/moshi/pull/281

PR #281, opened 2017-04-19
KotlinJsonAdapter

swankjesse:


kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,203 @@
...
+    override fun containsKey(key: KParameter) = parameterValues[key.index] !== ABSENT_VALUE
+
+    override fun get(key: KParameter): Any? {
+      val value = parameterValues[key.index]
+      return if (value !== ABSENT_VALUE) value else null
+    }
+  }
+
+  companion object {
+    val FACTORY = Factory { type, annotations, moshi ->
```
JakeWharton: Needs `@JvmField` for Java consumers wanting to add this via `Class.FACTORY`

kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
```diff
@@ -0,0 +1,434 @@
...
+    assertThat(jsonAdapter.serializeNulls().toJson(encoded)).isEqualTo("{\"a\":null,\"b\":5}")
+
+    val decoded = jsonAdapter.fromJson("{\"a\":null,\"b\":6}")
+    assertThat(decoded.a).isEqualTo(null)
+    assertThat(decoded.b).isEqualTo(6)
+  }
+
+  class ExplicitNull(var a: Int?, var b: Int?)
+
+  // TODO(jwilson): if a nullable field is absent, just do the obvious thing instead of crashing?
```
JakeWharton: :+1:

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,203 @@
...
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.findAnnotation
+import kotlin.reflect.full.memberProperties
+import kotlin.reflect.full.primaryConstructor
+import kotlin.reflect.jvm.isAccessible
+import kotlin.reflect.jvm.javaField
+import kotlin.reflect.jvm.javaType
+
+/** Classes annotated with this are eligible for this adapter. */
+private val KOTLIN_METADATA = Class.forName("kotlin.Metadata") as Class<out Annotation>
```
JakeWharton: Sad being that it's not a public type

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    // Read each value into its slot in the array.
+    val values = Array<Any?>(bindings.size) { ABSENT_VALUE }
+    reader.beginObject()
+    while (reader.hasNext()) {
+      val index = reader.selectName(options)
+      val binding = if (index != -1) bindings[index] else null
+
+      if (binding == null) {
+        reader.nextName()
```
swankjesse: Yeah, but only because selectName returned -1

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
swankjesse: Yep. Apparently it’s unlikely to change.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
swankjesse: Fixed.

PR #282, opened 2017-04-19
Fix Factory visibility

cbruegg:
Since `FACTORY` was defined inside the companion object of the `internal` class `KotlinJsonAdapter`, it was inaccesible from Kotlin modules other than `moshi-kotlin`. I've fixed this by making it a top-level singleton `object`.

### DIFF
--- a/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
@@ -43,7 +43,7 @@ private object ABSENT_VALUE
  * This class encodes Kotlin classes using their properties. It decodes them by first invoking the
  * constructor, and then by setting any additional properties that exist, if any.
  */
-internal class KotlinJsonAdapter<T> constructor(
+internal class KotlinJsonAdapter<T>(
     val constructor: KFunction<T>,
     val bindings: List<Binding<T, Any?>?>,
     val options: JsonReader.Options) : JsonAdapter<T>() {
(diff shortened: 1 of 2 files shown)


cbruegg:
I wonder why this didn't trigger Travis CI as other PRs did?

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -143,61 +144,61 @@ internal class KotlinJsonAdapter<T> private constructor(
       return if (value !== ABSENT_VALUE) value else null
     }
   }
+}
 
-  companion object {
-    @JvmField val FACTORY = Factory { type, annotations, moshi ->
-      if (!annotations.isEmpty()) return@Factory null
+object KotlinJsonAdapterFactory : JsonAdapter.Factory {
```
JakeWharton: This has ugly implications for Java callers. We need `@file:JvmName` to get a decent named enclosing class and then we want this to be exposed ideally as a `FACTORY` field.

cbruegg: It's accessed using `KotlinJsonAdapterFactory.INSTANCE` from Java in this PR.

Using `@file:JvmName("Something")` and naming it `FACTORY` would mean it's accessed like this from Java: `Something.FACTORY` and simply `FACTORY` from Kotlin. The latter seems to be worse than the downside of appearing as `KotlinJsonAdapterFactory.INSTANCE` from Java caused by this PR.

There's another option: Make `KotlinJsonAdapter` and thus the companion object public, but all of its non-inherited members `internal`. That may be error-prone though.

JakeWharton: Ah yeah it's a type not a function. I suppose that's an acceptable trade-off

cbruegg:
Merged `master` into this to fix the build.

swankjesse:
Thanks!

PR #285, opened 2017-04-20
Treat negative zero as a number, not a long.

NightlyNexus:
Updates logic from https://github.com/google/gson/issues/1053

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -464,7 +464,8 @@ final class JsonUtf8Reader extends JsonReader {
     }
 
     // We've read a complete number. Decide if it's a PEEKED_LONG or a PEEKED_NUMBER.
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -464,7 +464,8 @@ private int peekNumber() throws IOException {
     }
 
     // We've read a complete number. Decide if it's a PEEKED_LONG or a PEEKED_NUMBER.
-    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative)) {
+    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative) && (value
+        != 0 || !negative)) {
```
JakeWharton: This is weird wrapping. Our styles do this? I'd prefer it to wrap on the boolean operator.

NightlyNexus: oops, agreed. fixed!
(looks like i reset my style preferences.)

Issue #96, opened 2015-10-03
Recipe for buffered value


adriancole:
I've a scenario where I need to know a peer in order to properly interpret the other. In this case, there's a `type` field and a `value` field. The `value` could be a short, int, long, or double. I don't want to depend on field declaration order.

I was thinking about how to defer the decision since `JsonReader` smartly handles numbers.

Ex.

``` java
JsonReader valueReader = null;
switch (reader.nextName()) {
  case "type":
    type = Type.fromValue(reader.nextString());
    break;
  case "value":
    valueReader = reader.bufferValue(); // similar to peek, except consumes the next value
    break;
--snip--

switch (type) {
  case I64:
    value = valueReader.nextLong();
    break;
  case DOUBLE:
    value = valueReader.nextDouble();
    break;
--snip--
```

swankjesse:
Interesting. What I like is that we have some tech that makes this relatively straightforward.

NightlyNexus:
@adriancole
Will [`JsonReader.readJsonValue()`](https://github.com/square/moshi/blob/b4c43ae77193185e30fb062b29006ab03843ece3/moshi/src/main/java/com/squareup/moshi/JsonReader.java#L403) now do what you want?

adriancole:
slight glitch is it makes a thing I might need as a long a double.  but
close enough I think!

adriancole:
actually another glitch is quoted numbers.. In a perfect world, I could snap off the leaf and the poke at it. Even without this change, it is still fine. Thanks though!

NightlyNexus:
Oh, cool! Thank you for clarifying for me.
This seems straightforward to implement now.

I'll try it out this weekend and see if it's something others want to support.

NightlyNexus:
Btw, the alternative is to delegate to the built-in adapters to do the conversions for you (because `JsonAdapter.fromJsonValue` creates the `JsonValueReader`).

```java
final JsonAdapter<Double> doubleDelegate;
final JsonAdapter<Long> longDelegate;
fromJson(JsonReader reader) {
  Map<String, Object> object = reader.readJsonValue();
  String type = object.get("type");
  Object value = object.get("value");
  switch (type) {
    case I64:
      value = longDelegate.fromJsonValue(value);
      break;
    case DOUBLE:
      value = doubleDelegate.fromJsonValue(value);
      break;
    default:
      throw new JsonDataException("Unknown type: " + type);
  }
}
```

swankjesse:
[...]

PR #287, opened 2017-04-21
Add JsonReader.bufferValue().

NightlyNexus:
Closes https://github.com/square/moshi/issues/96 if Moshi should indeed add this.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -445,6 +445,16 @@ public abstract class JsonReader implements Closeable {
   }
 
   /**
+   * Consumes the value of the next token and returns a JsonReader to defer reading the value.
+   *
+   * @throws JsonDataException if the next token is not a literal value, if a JSON object has a
+   * duplicate key.
+   */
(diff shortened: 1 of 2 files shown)


adriancole:
Agreed. This handles case when you want to read a leaf differently based on
data you might read late (ex a type key indicating whether the value is a
number or a string)

If this is generally useful would be happy to have this merged as I would
use it.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -445,6 +445,16 @@ public final Object readJsonValue() throws IOException {
   }
 
   /**
+   * Consumes the value of the next token and returns a JsonReader to defer reading the value.
+   *
+   * @throws JsonDataException if the next token is not a literal value, if a JSON object has a
+   * duplicate key.
+   */
+  public final JsonReader bufferValue() throws IOException {
+    return new JsonValueReader(readJsonValue());
```
NightlyNexus: This might be better exposed simply as JsonReader.of(Object). Awkwardly, any API here will render JsonAdapter.fromValue(Object) redundant.

PR #289, opened 2017-04-24
Throw NPE for null indent string in factory method.

NightlyNexus:
Fail when creating the JsonAdapter rather than when using it.


JakeWharton:
Test?

NightlyNexus:
Yep, it should.
Wasn't sure how much noise it would be.
Will update when not mobile.

PR #292, opened 2017-04-29
Make Types.equals(Type, Type) public.

serj-lotutovici:


serj-lotutovici:
There's only one direct test for this method. Mostly it's tested indirectly by other test. Do you think it's sufficient?

swankjesse:
This is good enough for now. I don’t expect tests to discover any surprises.

PR #297, opened 2017-05-06
Import jsr305 and use it to mark @Nullable stuff.

swankjesse:


moshi/src/test/java/com/squareup/moshi/JsonWriterTest.java
```diff
@@ -286,6 +286,15 @@
         + "3.141592653589793238462643383]");
   }
 
+  @Test public void nullNumbers() throws IOException {
```
swankjesse: A bug that revealed itself by being explicit about nulls.

swankjesse:
BTW, adding `@Nullable` in Java is not source-compatible with Kotlin. This is the best time to make this change!

```
[ERROR] /home/travis/build/square/moshi/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt: (46, 10) Class 'KotlinJsonAdapter' must be declared abstract or implement abstract base class member public abstract fun toJson(p0: JsonWriter!, @Nullable p1: T?): Unit defined in com.squareup.moshi.JsonAdapter

[ERROR] /home/travis/build/square/moshi/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt: (98, 3) 'toJson' overrides nothing
```

JakeWharton:
Looks like the package level non-null annotation isn't working from
Kotlin's side though. That's a shame.

swankjesse:
Because the return type is now strictly `@Nullable`, I needed to bang a bunch of tests.

```
    val decoded = jsonAdapter.fromJson("{\"a\":4,\"b\":6}")!!
    assertThat(decoded.a).isEqualTo(4)
    assertThat(decoded.b).isEqualTo(6)
```

swankjesse:
I think this is actually an improvement, since the input string could have been `"null"` which would have returned a null instance.

Issue #299, opened 2017-05-07
Kotlin adapter should omit properties with no backing field?


swankjesse:
I had a val that just computed a thing, and it broke my ability to use the Kotlin adapter.

PR #300, opened 2017-05-07
Support more kinds of properties in KotlinJsonAdapter

swankjesse:
This makes it possible to have synthetic properties that have no
state.

Also test properties that are synthetic and have no backing field.

Closes: https://github.com/square/moshi/issues/299

### DIFF
--- a/kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
+++ b/kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
@@ -445,20 +445,68 @@ class KotlinJsonAdapterTest {
     }
   }
 
-  @Test fun unsettableProperty() {
+  @Test fun unsettablePropertyIgnored() {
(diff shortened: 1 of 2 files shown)


kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
```diff
@@ -445,22 +445,70 @@ class KotlinJsonAdapterTest {
...
-          "val ${UnsettableProperty::class.qualifiedName}.a: kotlin.Int")
-    }
+    val jsonAdapter = moshi.adapter(UnsettableProperty::class.java)
+
+    val encoded = UnsettableProperty()
+    encoded.b = 5
+    assertThat(jsonAdapter.toJson(encoded)).isEqualTo("{\"b\":5}")
+
+    val decoded = jsonAdapter.fromJson("{\"a\":4,\"b\":6}")!!
+    assertThat(decoded.a).isEqualTo(-1)
```
JakeWharton: I think this is correct, but may be surprising

PR #314, opened 2017-05-25
JsonReader.readJsonValue: parse number as int/long/dbl

simon04:
Attempt an exact conversion to integer or long prior before returning a double.

This addresses https://github.com/influxdata/influxdb-java/issues/276.

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
@@ -927,7 +927,7 @@ public final class JsonReaderTest {
   @Test public void readJsonValueInt() throws IOException {
     JsonReader reader = newReader("1");
     Object value = reader.readJsonValue();
-    assertThat(value).isEqualTo(1.0);
+    assertThat(value).isEqualTo(1);
   }
 
   @Test public void readJsonValueMap() throws IOException {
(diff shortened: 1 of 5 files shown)


moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -297,7 +297,7 @@ JsonReader newReader(String json) throws IOException {
         + "2.718281828459045]";
     JsonReader reader = newReader(json);
     reader.beginArray();
-    assertThat(reader.nextDouble()).isEqualTo(-0.0);
+    assertThat(reader.nextDouble()).isIn(0.0, -0.0);
```
simon04: The distinction ±0.0 is lost in the `JsonValueReader` since `BigDecimal` does not distinguish between those.

swankjesse:
Thanks for sending this. Unfortunately, I don’t think we want to change the _default_ behavior here.

In particular I think these two JSONs should be equal when converted to Java.
`[1,2,3]`, `[1.0,2.0,3.0]`

Unfortunately, a Java `Long` is not equal to the corresponding `Double`, so we need to either pick the data type based on the encoding, or pick one data type universally.

For Moshi’s defaults we’ve chosen to use a single data type, `double`, to use everywhere.

That said, it is possible to get this behavior with a JsonAdapter. Lemme try to supply that...

simon04:
Thanks for your answer. The [JSON standard](https://tools.ietf.org/html/rfc7159#section-6) is pretty lose on the interpretation of numbers. So there's no clear _correct_ or _wrong_.

I'm highly interested in the mentioned `JsonAdapter`.

swankjesse:
@simon04 here’s a JSON adapter that does what we’re talking about. Unfortunately it didn’t work immediately as I was hoping so I submitted [this PR](https://github.com/square/moshi/pull/316). You’ll need a snapshot or the next release before this adapter works as you want.

https://gist.github.com/swankjesse/0331f24df1aa25454eaae04e49fe2f19

PR #316, opened 2017-05-27
Change the adapter for Object.class to delegate.

swankjesse:
Previously if we ever had an opaque Object, the content of this object
would always only use the built-in adapters for its members.

This changes the built-in Object adapter to do one layer of type checking
and then to delegate to user-supplied adapters.

The big upside of this is that application code can now change the default
numeric type to use when decoding an untyped object. Typically this will
be used to replace our default of Double with a user-specified numeric type
like BigDecimal.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
+++ b/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
@@ -20,6 +20,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -1006,6 +1011,107 @@
     assertThat(adapter.fromJson(json)).isEqualTo(new Pizza(5, true));
   }
 
+  /**
+   * Confirm that the built-in adapter for Object delegates to user-supplied adapters for JSON value
+   * types like strings.
+   */
+  @Test public void objectAdapterDelegatesStringNamesAndValues() throws Exception {
```
NightlyNexus: Any reason to put them in MoshiTest instead of ObjectAdapterTest?

swankjesse: Yep, that’s better. Fixed.

PR #321, opened 2017-06-09
proguard config updates

OleksandrKucherenko:
updated proguard configuration for custom Json Adapters @FromJson/@ToJson cases


swankjesse:
Thanks.

PR #328, opened 2017-07-04
Allow configuration of maximum nesting depth.

programingjd:
I ran into the max depth limit with a (very) complex elastic query.
I think the max depth should be configurable.

I've added a Config object that is used to configure the maximum allowed nesting depth before a JsonDataException is thrown.
The JsonReaders and JsonWriters are using this variable for sizing their internal arrays rather than hardcoding the length to 32.

I experimented with adding additional implementations of JsonReader and JsonWriter, but I couldn't get something clean because of the factory methods. This static Config seemed like the best way to do it.
There's no change to the existing APIs at all, juste an extra Config object to increase (or decrease the max depth allowed).

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Config.java
+++ b/moshi/src/main/java/com/squareup/moshi/Config.java
@@ -1,0 +1,9 @@
+package com.squareup.moshi;
+
+public class Config {
+
+  private Config() {}
+
+  public static int maxDepth = 32;
+
+}
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -175,13 +175,14 @@ import okio.ByteString;
  * of this class are not thread safe.
  */
 public abstract class JsonReader implements Closeable {
-  // The nesting stack. Using a manual array rather than an ArrayList saves 20%. This stack permits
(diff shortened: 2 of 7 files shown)


programingjd:
Sorry about the build failure. It looks like there's a line going over the 100 chars limit.

JakeWharton:
What depth would you need?

programingjd:
Around 50 I think.

JakeWharton:
As is we can't accept this global method of configuring the library. As a rule adding a configuration option is usually close to last if not last. Perhaps we could increase the limit though.

programingjd:
Yes, I understand, that's why I tried to have different implementations of JsonReader and JsonWriter before doing the static config. I would go with 64 max depth to be on the safe side. Elastic is quite verbose and it's relatively easy to go over 32, but I think 64 should be fine.
I saw the comment about the 20% savings achieved by using arrays rather than ArrayLists. Is the 20% savings really worth it?

swankjesse:
Would you like your JSON parser to be 20% faster?

programingjd:
I am not sure 20% faster justifies having a limited nesting depth.

programingjd:
How about adding some api similar to setLenient() on the reader/writer. Would that be acceptable?
I agree that we don't want to give up the extra speed if we don't have to.

PR #329, opened 2017-07-04
setMaxDepth(int) on readers and writers.

programingjd:
Second attempt at adding support for deeper nesting.
I've added setMaxDepth(int maxDepth) methods on JsonReader and JsonWriter, and the maxDepth(int maxDepth) method on JsonAdapter.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -169,6 +169,25 @@ public abstract class JsonAdapter<T> {
     };
   }
 
+  /** Returns a JSON adapter equal to this, but allows the specified maximum depth when
+   *  reading and writing. */
+  public final JsonAdapter<T> maxDepth(final int maxDepth) {
+    final JsonAdapter<T> delegate = this;
+    return new JsonAdapter<T>() {
+      @Override public @Nullable T fromJson(JsonReader reader) throws IOException {
+        reader.setMaxDepth(maxDepth);
+        return delegate.fromJson(reader);
+      }
+      @Override public void toJson(JsonWriter writer, @Nullable T value) throws IOException {
+        writer.setMaxDepth(maxDepth);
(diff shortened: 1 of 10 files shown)


swankjesse:
I don’t think this is a good fit for Moshi. We could change the code to statically allocate a larger initial buffer (say, 64) or we could grow the buffer on-demand to some limit (say, 64). But making the user pick their depth is too awkward.

programingjd:
The idea was that only the user knows how deep he needs the nesting to be.
If you never run into the JsonDataException with the too deeply nested error message, then you don't have to do anything. And if you do, then that made the fix easy to find.

I can see that growing the buffer automatically when needed could be better, but it would change the current behavior (it might be ok).
If you want to make this optional, then where would you put the call to toggle the option on (and how would you call it)?
I put the setMaxDepth next to the other options (setLenient for instance) to be consistent.
Maybe we can just leave it there but change it to remove the configurable max depth and grow it to a fixed size (64 or even 128).

programingjd:
I am closing this, but I still think this needs to be addressed. I can't be the only one running into the limitation, and right now there's no way to get around it other than using a fork or another lib.
I am open to suggestions for working further on this towards something that would be accepted.

swankjesse:
I wanna fix this by making it scale automatically. In the meantime, if you wanna send a PR that bumps the limit to 64, please do. You’ll probably need to fix a few tests.

programingjd:
By reallocating bigger arrays in pushScope?
Something similar to what I have done [here](https://github.com/programingjd/okjson/blob/master/src/main/java/info/jdavid/ok/json/JsonWriter.java)?

swankjesse:
I think we just want to manually have a max depth of 64, non-resizable.

nicusorflorin:
any chance of this being addressed? "Nesting too deep" error is rather a pest to get around

swankjesse:
Fixed by dynamically resizing up to 256 levels.

PR #332, opened 2017-07-18
Resolve generic types in kotlin.

### LINKED ISSUE #309: Generic Kotlin data classes cannot be (de)serialized
It seems that Moshi 1.5.0 has a problem with (de)serialization of generic data classes.

```kotlin
data class Test<out T>(val item: T)

data class Test2(val test: Test<String>)
```

The following code

```kotlin
val json = """{test:{item:"hello"}}"""
val moshi = Moshi.Builder()
        .build()
val data = moshi.adapter(Test2::class.java).fromJson(json)
```

produces the exception

```
java.lang.IllegalArgumentException: Expected a Class, ParameterizedType, or GenericArrayType, but <null> is of type null
  at com.squareup.moshi.Types.getRawType(Types.java:167)
  at com.squareup.moshi.ClassJsonAdapter$1.createFieldBindings(ClassJsonAdapter.java:83)
  at com.squareup.moshi.ClassJsonAdapter$1.create(ClassJsonAdapter.java:75)
  at com.squareup.moshi.Moshi.adapter(Moshi.java:100)
  at com.squareup.moshi.KotlinJsonAdapterFactory.create(KotlinJsonAdapter.kt:184)
  at com.squareup.moshi.Moshi.adapter(Moshi.java:100)
  at com.squareup.moshi.KotlinJsonAdapterFactory.create(KotlinJsonAdapter.kt:184)
  at com.squareup.moshi.Moshi.adapter(Moshi.java:100)
  at com.squareup.moshi.Moshi.adapter(Moshi.java:62)
  at com.svenjacobs.test.components.main.view.MainActivity.onCreate(MainActivity.kt:24)
```

This was working in Moshi 1.4.0 and before.

serj-lotutovici:
Closes #309

### DIFF
--- a/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
@@ -15,9 +15,11 @@
  */
 package com.squareup.moshi
 
+import java.lang.reflect.GenericArrayType
 import java.lang.reflect.Modifier
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
+import java.lang.reflect.WildcardType
 import java.util.AbstractMap.SimpleEntry
 import kotlin.collections.Map.Entry
 import kotlin.reflect.KFunction
@@ -149,7 +151,7 @@ internal class KotlinJsonAdapter<T>(
 }
 
 class KotlinJsonAdapterFactory : JsonAdapter.Factory {
-  override fun create(type: Type?, annotations: MutableSet<out Annotation>, moshi: Moshi)
+  override fun create(type: Type, annotations: MutableSet<out Annotation>, moshi: Moshi)
       : JsonAdapter<*>? {
     if (!annotations.isEmpty()) return null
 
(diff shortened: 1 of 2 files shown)


kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -16,13 +16,11 @@
...
 import java.lang.reflect.Modifier
+import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import java.util.AbstractMap.SimpleEntry
 import kotlin.collections.Map.Entry
-import kotlin.reflect.KFunction
-import kotlin.reflect.KMutableProperty1
-import kotlin.reflect.KParameter
-import kotlin.reflect.KProperty1
+import kotlin.reflect.*
```
JakeWharton: Revert

swankjesse:
I suspect this is too simple. For example, I don’t think this is going to handle cases where the generic types’ type arguments aren’t used directly.

```
class Multitable<R, C, V> {
  Map<Pair<R, C>, List<V>> cells;
}
```

serj-lotutovici:
@swankjesse sorry missed that. I've pushed a small fix to expand on the idea how more complex generics would be handled. If you consider this a good approach, I'll wrapup and fix the todos.
Don't know what to do with wildcard types though.

swankjesse:
Yeah, in Guice and Gson there are big ugly recursive methods to do generic type resolution. I think we probably need that here too.

serj-lotutovici:
>Yeah, in Guice and Gson there are big ugly recursive methods to do generic type resolution. I think we probably need that here too.

Will take a look...

serj-lotutovici:
After some time investigating, I've got lost in the usage of `getGenericSupertype` and how to apply it in the kotlin adapter factory. Also, the generics seam to be stripped away from the java byte code, and that method becomes usless when trying to retrive the right type from a kotlin class.
I completry uderstand that I may be missing something, but this is as far as my brain power can take me at the moment.

Btw, after reading [this](https://kotlinlang.org/docs/reference/generics.html) more carefully I've come to realize that kotlin doesn't use java's representation of wildcard types, so `out` and `in` are backed up only by the compiler.

NightlyNexus:
#475 fixed this, as far as i see.

Issue #315, opened 2017-05-26
Moshi 1.5 ignores @Json annotation on Kotlin class properties


consp1racy:
```
@Keep
class RestUserSubscriptionPackage(
        val id: Int,
        @Json(name = "package_id") val package_prototype_id: Int,
        @Json(name = "package") val package_prototype: RestSubscriptionPackage?,
        val job_credit: JobCredit,
        val starts_at: OffsetDateTime,
        val ends_at: OffsetDateTime
)
```

`package_prototype_id` is always 0, `package_prototype` is always `null`.

Works out-of-the-box in Moshi 1.4.

Workaround: `@field:Json`

swankjesse:
Try adding `moshi-kotlin` as a dependency and including it when you configure your `Moshi` instance.

```
    val moshi = Moshi.Builder()
        .add(KotlinJsonAdapterFactory())
        .build()
```

consp1racy:
Not an option at the time, it crashes at runtime with `Reflection on built-in Kotlin types is not yet fully supported. No metadata found for ` [compareTo(other: something): Int]. When I pinpoint what's happening, I'll open another issue.

swankjesse:
Please do. Yeah, we wanna make sure this stuff all works well.

ligi:
Dam - moshi-kotlin is not yet an option on Android as for the size of the kotlin-reflect dependency - makes me need to go back to 1.4.0 for this reason.

consp1racy:
@ligi Have you tested after proguard?

ligi:
@consp1racy  I did not really test - was reading https://medium.com/square-corner-blog/kotlins-a-great-language-for-json-fcd6ef99256b
>If you’re using JSON, Moshi and Kotlin help you to build better models with less code. Note that moshi-kotlin uses kotlin-reflect for property binding. That dependency is large by Android standards (1.7 MiB / 11,500 methods). We’re thinking of creative ways to address that!

jaredsburrows:
@consp1racy Thanks for posting this. I didn't want to roll back to `1.4.0`. I have tested `1.4.0` and it does work out the box. Thanks for the `@field:Json` work around.

I completely agree with @ligi, the `moshi-kotlin` dependency is too large. We should not have to add another dependency that includes transitive dependencies in order to solve this issue. See the dependency count here: http://www.methodscount.com/?lib=com.squareup.moshi%3Amoshi-kotlin%3A1.5.0.
[...]

PR #336, opened 2017-07-26
Add moshi-kotlin documentation

### LINKED ISSUE #320: Document Kotlin module ProGuard rules.
Something like
```
    public <methods>;
}
```

Jawnnypoo:
Resolves #320 and the general fact that moshi-kotlin is undocumented, highlighted somewhat in issue #315 

@swankjesse I could not predict the wording and order of where the documentation should go in the README, or the wording, so just let me know and I can tweak as needed.

I think it's important to document the existence of moshi-kotlin right now even if it is not as ideal as it could be with the reliance on `kotlin-reflect`. Side-note: I believe @Takhion is creating a library that could potentially be a drop in solution for reading the Kotlin metadata here: https://github.com/Takhion/kotlin-metadata But I suppose that is somewhat unrelated to this PR 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,6 +472,14 @@ public final class BlackjackHand {
 }
 ```
 
+### Kotlin Support
+
+Kotlin classes work with Moshi out of the box, with the exception of annotations. If you need to annotate your Kotlin classes with an `@Json` annotation or otherwise, you will need to use the `moshi-kotlin` artifact, and set up Moshi to use its converter factory:
+```kotlin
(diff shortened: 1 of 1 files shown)


Jawnnypoo:
Ahh, good point @serj-lotutovici I didn't realize that was a feature of `moshi-kotlin`. 

Also, would this documentation be better off in the `kotlin` folder, similar to https://github.com/square/moshi/pull/337 ?

swankjesse:
Thanks.

PR #337, opened 2017-07-26
Document `moshi-adapters` artifact

Jawnnypoo:
Title says it all! Took the format from https://github.com/square/retrofit explanation of converters

### DIFF
--- a/adapters/README.md
+++ b/adapters/README.md
@@ -1,0 +1,37 @@
+Adapters
+===================
+
+Prebuilt Moshi `JsonAdapter`s for various things, such as `Rfc3339DateJsonAdapter` for parsing `java.util.Date`s
+
+To use, supply an instance of your desired converter when building your `Moshi` instance.
+
+```java
+Moshi moshi = new Moshi.Builder()
+    .addConverterFactory(new Rfc3339DateJsonAdapter())
+    //etc
+    .build();
+```
+
+Download
+--------
+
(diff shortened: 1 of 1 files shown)


adapters/README.md
```diff
@@ -0,0 +1,37 @@
+Adapters
+===================
+
+Prebuilt Moshi `JsonAdapter`s for various things, such as `Rfc3339DateJsonAdapter` for parsing `java.util.Date`s
+
+To use, supply an instance of your desired converter when building your `Moshi` instance.
+
+```java
+Moshi moshi = new Moshi.Builder()
+    .addConverterFactory(new Rfc3339DateJsonAdapter())
```
NightlyNexus: `.add(Date.class, new Rfc3339DateJsonAdapter())`

Jawnnypoo: Ahh, good catch

NightlyNexus:
>JsonAdapters for various things

I'm also curious here: would this adapters artifact be used for anything other than the date adapter?

Jawnnypoo:
Great question, I could see if being something similar to Retrofit where each package would get its own artifact or something like that. So this would be `moshi-adapters` since it does not contain any other dependency other than Java itself, and something like [threetenbp](https://github.com/ThreeTen/threetenbp) would have its own: `moshi-adapters-threetenbp` etc. Was that the goal with this artifact, @swankjesse ?

swankjesse:
Yep, that’s the plan.

PR #340, opened 2017-08-01
Extract moshi instance creation in KotlinJsonAdapterTest

serj-lotutovici:
This will make new tests less verbose except for the cases when an additional adapter factory is being used.

### DIFF
--- a/kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
+++ b/kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
@@ -25,9 +25,12 @@ import java.util.SimpleTimeZone
 import kotlin.annotation.AnnotationRetention.RUNTIME
 
 class KotlinJsonAdapterTest {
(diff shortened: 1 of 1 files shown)


kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
```diff
@@ -25,9 +25,12 @@ import java.util.SimpleTimeZone
 import kotlin.annotation.AnnotationRetention.RUNTIME
 
 class KotlinJsonAdapterTest {
+  /** Returns a new instance of [Moshi] with [KotlinJsonAdapterFactory] added. */
+  val newMoshi: Moshi
```
serj-lotutovici: Although it's a `val` in bytecode this will be called as a getter method, so we can be sure that tests are not affecting each other.

swankjesse:
This is on purpose.

In application code, I definitely prefer to reuse as much as possible. DRY.

In test code, self-contained tests are worth the repetition. All of the context that tells the story of the test is in one place.

serj-lotutovici:
Got it. Closing.

Issue #313, opened 2017-05-25
Moshi Adapter and Proguard


mirono:
We wrote an adapter that maps the same class to itself (i.e. String -> String):

_File: MultiLineString.java_
```
@JsonQualifier
@Retention(RetentionPolicy.RUNTIME)
public @interface MultilineString {}
```

_File: MultiStringAdpater.java_
```
public class MultilineStringAdapter {
    private static final Pattern TO_PATTERN = Pattern.compile("\n", Pattern.LITERAL);
    private static final Pattern FROM_PATTERN = Pattern.compile("~~~", Pattern.LITERAL);

    @Nullable
    @ToJson String toJson(@MultilineString String value) {
        if(Strings.isNullOrEmpty(value)) {
            return "";
        }
        return TO_PATTERN.matcher(value).replaceAll("~~~");
    }

    @Nullable
    @FromJson @MultilineString String fromJson(String value) {
        if(Strings.isNullOrEmpty(value)) {
            return "";
        }
        return FROM_PATTERN.matcher(value).replaceAll("\n");
    }
}
```
Everything worked fine in dev builds. On production we were getting StackOverflow exception in: 
```
java.lang.StackOverflowError: stack size 1038KB
     at com.squareup.moshi.AdapterMethodsFactory$5.fromJson(Unknown Source)
     at com.squareup.moshi.AdapterMethodsFactory$1.fromJson(Unknown Source)
     at com.squareup.moshi.Moshi$DeferredAdapter.fromJson(Unknown Source)
     at com.squareup.moshi.AdapterMethodsFactory$5.fromJson(Unknown Source)
...
```
It seems the adapters factory initialization is running into an endless loop. Being only in production we suspected Proguard. We tried several options but the one that worked was:
```
@Keep
@JsonQualifier
@Retention(RetentionPolicy.RUNTIME)
public @interface MultilineString {}
```
We suspect the Proguard removed the attribute @Retention(RetensionPolicy.RUNTIME) causing the the run-time infinite mapping of this adapter to itself.

Maybe adding some documentaion regarding the recommended Proguard setting would help?

Miron

OleksandrKucherenko:
```
-keepclasseswithmembers class * {
    @com.squareup.moshi.* <methods>;
}
```

mirono:
This is not enough as the annotation for the interface is removed by proguard, thus it is not retained and Moshi will not find the adapter at run time. The documentation should state the the interface should be annotated with @Keep

[...]

PR #342, opened 2017-08-04
Add ProGuard rule for JsonQualifier

Jawnnypoo:
Hopefully, this will save others from the 3+ hours of ProGuard hell I just dredged through. 

Looks like annotations meant for `@JsonQualifier` were getting stripped by ProGuard, which does not cause builds to fail, so it is extremely subtle. It also would not cause runtime errors all the time, just for specific cases, such as if you had two `java.util.Date` formats, like so:
```
new Moshi.Builder()
                .add(new DashDateAdapter())
                .add(new SlashDateAdapter())
                .build();
```
It would presumably fail to find the annotations, and would just use DashDateAdapter for all `java.util.Date`s, since it is added to Moshi first. Therefore the error would typically come from that adapter trying to parse a date that is meant for another adapter entirely. 

The rule is a derivation of one I found for [LoganSquare](https://github.com/bluelinelabs/LoganSquare) which appears as:
`-keep @com.bluelinelabs.logansquare.annotation.JsonObject class *`

I believe the new rule is specific enough, since it will force keep all annotation classes that are annotated with `@JsonQualifier` but honestly, ProGuard rules can be hard for me to grasp. Please let me know if there is a more restrictive rule that would suit this case better. Regardless, this one fixed the problem for me, and I believe is the general rule that will work for everyone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -523,6 +523,7 @@ If you are using ProGuard you might need to add the following options:
 -keepclasseswithmembers class * {
     @com.squareup.moshi.* <methods>;
 }
+-keep @com.squareup.moshi.JsonQualifier class *
 ```
 Additional rules are needed if you are using the Kotlin artifact:
 ```
--- a/README.md
+++ b/README.md
@@ -523,7 +523,7 @@ If you are using ProGuard you might need to add the following options:
 -keepclasseswithmembers class * {
     @com.squareup.moshi.* <methods>;
 }
--keep @com.squareup.moshi.JsonQualifier class *
+-keep @com.squareup.moshi.JsonQualifier interface *
 ```
 Additional rules are needed if you are using the Kotlin artifact:
 ```


Jawnnypoo:
Resolves https://github.com/square/moshi/issues/313

README.md
```diff
@@ -523,6 +523,7 @@ If you are using ProGuard you might need to add the following options:
 -keepclasseswithmembers class * {
     @com.squareup.moshi.* <methods>;
 }
+-keep @com.squareup.moshi.JsonQualifier class *
```
NightlyNexus: lazyweb: does `class` here cover interfaces, too?

Jawnnypoo: Seems like that is the case. I have my annotations in their own class files though, let me see if it matters to have them as inner classes of the adapter or something else like that.

Jawnnypoo: Yep, seems to work on inner classes of the adapter too. But yeah, I dunno if there is some ProGuard keyword for just interfaces

NightlyNexus: `interface *` i think is just for interfaces? not that i know i'm talking about.

JakeWharton: It would be `-keep interface`.

JakeWharton: https://www.guardsquare.com/en/proguard/manual/usage#classspecification

Jawnnypoo: So, in full:
`-keep @com.squareup.moshi.JsonQualifier interface *` ?

JakeWharton: Yep

Jawnnypoo:
Thanks for the feedback @NightlyNexus and @JakeWharton . Updated for just interfaces instead of all classes :+1:

PR #352, opened 2017-09-24
Add an example custom Adapter which delegates

emmaguy:
When writing a custom adapter recently, I was a bit confused about how to go about making one which does some logic and then delegates the rest as usual - so thought it would be worth contributing a sample.

I reused the example which I also [blogged about](https://medium.com/@emmaguy/creating-a-custom-type-adapter-for-moshi-ae7e1cf469a9). I named the file based on what it is, not sure if there's a convention for them though.

I have signed the CLA! 💪


swankjesse:
Nice! Thank you!

PR #354, opened 2017-09-25
Add CheckReturnValue for toJson's string result.

NightlyNexus:
This helps Error Prone and the IDE find accidental usages of toJson(value) instead of toJson(writer, value).


NightlyNexus:
(just failing because #353 fails Error Prone, yay.)

NightlyNexus:
downsides: clutter and also needing ProGuard people to include `-dontwarn javax.annotation.*` or whatever to their configurations.

swankjesse:
We should do this throughout our libraries!

NightlyNexus:
@swankjesse Are you thinking for every method with no side effects, or a select few that look ambiguous?

JakeWharton:
In SQL Brite we do it for every method

PR #360, opened 2017-10-04
Allow delegates for intermediates in adapters.

### LINKED ISSUE #346: Support for String Type Adapter
Not sure if it is "out of scope" but since Type Adapter are super powerful, it would be handy to be able to add a Type Adapter for Strings.

Often I have to remove html tags from json. Adding a Type Adapter that does that would be nice, but this might be out of scope since ideally the backend doesn't provide such kind of data (html tags in json) or you move the task of removing html tags one step higher in your business logic (not part of serializing / deserializing).

Are you interested in investigating in this topic? If yes, i could take a closer look into the source code and try to create a pull request for this.

Right now, it seems that we run in a StackOverflow because recursively calling the string adapter:
```kotlin
class HtmlStringAdapter {

  @ToJson
  fun toString(to: String?): String? = to

  @FromJson
  fun fromString(str: String?): String? {
    return if (str == null)
      return null
    else
      Html.fromHtml(str).toString() // quick & dirty proof of concept  workaround to remove html tags ... don't try this at home
  }
}
```

Stacktrace:
```
java.lang.StackOverflowError: stack size 1038KB
                                                                 at com.squareup.moshi.Moshi$DeferredAdapter.fromJson(Moshi.java:234)
                                                                 at com.squareup.moshi.AdapterMethodsFactory$5.fromJson(AdapterMethodsFactory.java:248)
                                                                 at com.squareup.moshi.AdapterMethodsFactory$1.fromJson(AdapterMethodsFactory.java:87)
                                                                 at com.squareup.moshi.Moshi$DeferredAdapter.fromJson(Moshi.java:234)
                                                                 at com.squareup.moshi.AdapterMethodsFactory$5.fromJson(AdapterMethodsFactory.java:248)
                                                                 at com.squareup.moshi.AdapterMethodsFactory$1.fromJson(AdapterMethodsFactory.java:87)
                                                                 at com.squareup.moshi.Moshi$DeferredAdapter.fromJson(Moshi.java:234)
```

NightlyNexus:
Closes #346.
Builds on #272.

I don't have a real use case for this, but it could be surprising that some delegation works and some doesn't.

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.moshi;
 
+import com.squareup.moshi.MoshiTest.Uppercase;
+import com.squareup.moshi.MoshiTest.UppercaseAdapterFactory;
 import java.io.IOException;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
```diff
@@ -130,6 +132,36 @@
     assertThat(adapter.fromJson("[{\"x\":5,\"y\":8}]")).isEqualTo(point);
   }
 
+  @Test public void toAndFromWithIntermediate() throws Exception {
```
swankjesse: nice

swankjesse:
I love it.

PR #363, opened 2017-10-09
make ClassJsonAdapter not final

lfarkas:
in this PR i made following changes:

- remove final from ClassJsonAdapter to be able to subclass it
- set includeField as static and moe out from the factory same as isPlatformType since they have totally similar purpose
- move out from the factory and make static the createFieldBindings
- factor out getRawType from the JsonAdapter.Factory.create same as all the above includeField, isPlatformType and createFieldBindings

the reason for these cahnges to be able to subclass from ClassJsonAdapter and easily implement other adapter classes for which i'll send another PR.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -37,7 +37,9 @@ public abstract class JsonAdapter<T> {
   }
 
   public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    try (Buffer buffer = new Buffer()) {
+    	return fromJson(buffer.writeUtf8(string));
+    }
   }
 
   public abstract void toJson(JsonWriter writer, @Nullable T value) throws IOException;
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -41,10 +41,21 @@ import javax.annotation.Nullable;
  *   <li>scala.*
  * </ul>
  */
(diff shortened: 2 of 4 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -37,7 +37,9 @@
   }
 
   public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    try (Buffer buffer = new Buffer()) {
+    	return fromJson(buffer.writeUtf8(string));
+    }
```
NightlyNexus: writing to a Buffer does not IO, and the changes outside of ClassJsonAdapter are unnecessary and unrelated to the change you are trying to make. will you revert these changes please?

lfarkas: this way an accident with other code cleanup i like to submit later..

lfarkas:
i was on vacation and i assume i can do it during my holiday so I'll create a new PR only for this and home not to mix different codes.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
lfarkas: anyway Buffer is anAutoClosabe and these should have to be close otherwise code checkers are argue about it

swankjesse:
I don't think we want this. This leaks implementation details and prevents potential future optimizations.

If you want to subclass this, please just copy and paste it into your own project. It isn't that big!

lfarkas:
i write another PR you may close that #364 :-(
but i still not agree with you since eg: isPlatformType should have to be used in all other places otherwise the code becomes inconsistent. the same apply to includeField and getRawType. may be the createFieldBindings can be copied but imho refactoring this function into a separate function makes the code more cleaner.

Issue #357, opened 2017-10-01
@JsonFormat(shape=JsonFormat.Shape.ARRAY) support in moshi


lfarkas:
In jackson one can write:
```
@JsonPropertyOrder({"x", "y"})
public class Point {
  public int x, y;
}
```
by adding annotation
```
@JsonFormat(shape=JsonFormat.Shape.ARRAY)
```
and get serialization like:
```
[12,35]
```
instead of
```
{ "x":12,"y":35}
```

How can I do same with moshi? 
Is it possible to add this feature to moshi?
Thanks in advance.

swankjesse:
You’ll need to use `@ToJson` and `@FromJson` to do this.

lfarkas:
OK but I wouldn't like to do it all the time...that's the reason why there is an annotation for this. eg I've got a few such class where I'd like this behaviour. Is there any way to write a common adapter, base class, annotation etc which can i use for all of my such class?

swankjesse:
Definitely. Take a look at JsonAdapter.Factory.
http://square.github.io/moshi/1.x/moshi/com/squareup/moshi/JsonAdapter.Factory.html

It can be used to implement `@JsonPropertyOrder` completely in your own code.

lfarkas:
is there any such example? wouldn't be useful add such to moshi (at least those which are in jackson) at least JsonFormat and JsonPropertyOrder or something like this eg:
```
@JsonFormat(shape=JsonFormat.Shape.ARRAY)
@JsonPropertyOrder({"x", "y"})
```
since these 2 line used to used together may one adapter or something would be useful.

swankjesse:
Here's one example. The factory lets you inspect each type, and return a JsonAdapter for the types you want to support.

You'd need to look over all of the fields to get delegate adapters, then return an adapter that calls them in sequence to build an array.

https://github.com/square/moshi/blob/master/examples/src/main/java/com/squareup/moshi/recipes/Unwrap.java

NightlyNexus:
>I wouldn't like to do it all the time...that's the reason why there is an annotation for this

Moshi uses JsonQualifier annotations to differ behavior within the same Java types.
You can add a `@Retention(RUNTIME) @JsonQualifier @interface JsonArray {}`.

Combined with JsonAdapter.Factory, JsonQualifiers should let you implement what you want here.
Feel free to post on StackOverflow with the `moshi` tag if you still need help after the example above.

PR #364, opened 2017-10-09
make ClassJsonAdapter not final

lfarkas:
in this PR i made following changes:

- remove final from ClassJsonAdapter to be able to subclass it
- set includeField as static and move out from the factory same as isPlatformType since they have totally similar purpose
- move out from the factory and make static the createFieldBindings
- factor out getRawType from the JsonAdapter.Factory.create same as all the above includeField, isPlatformType and createFieldBindings

the reason for these changes to be able to subclass from ClassJsonAdapter and easily implement other adapter classes for which i'll send another PR (described in #357)


swankjesse:
I’m sorry; this class is not a public API and not intended for subclassing. If you want this behavior you should copy the entire class into your project.

PR #366, opened 2017-10-12
Fix compile warnings

lfarkas:
- fix generic warnings 
- fix autocloseable resource release right after use (not just at gc run)


moshi/src/test/java/com/squareup/moshi/JsonCodecFactory.java
```diff
@@ -29,8 +29,10 @@
       Buffer buffer;
 
       @Override public JsonReader newReader(String json) {
-        Buffer buffer = new Buffer().writeUtf8(json);
-        return JsonReader.of(buffer);
+        try (Buffer buffer = new Buffer()) {
```
JakeWharton: This is incorrect as it will call `close()` on the `Buffer` before returning a `JsonReader` that wraps it and intends to then read from it.

JakeWharton: It doesn't break anything because `Buffer` ignores `close()`, but it's still semantically wrong.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -37,7 +37,9 @@
   }
 
   public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    try (Buffer buffer = new Buffer()) {
```
JakeWharton: This is also weird and seems unnecessary. Calling `close()` on a `Buffer` [is a no-op][1] so there's nothing gained by doing this.

 [1]: https://github.com/square/okio/blob/20e259c08af6c96508387540a5ea53f89b1d01f5/okio/src/main/java/okio/Buffer.java#L1530-L1531

PR #367, opened 2017-10-12
implement array json adapter

lfarkas:
as described in #357

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/AsArrayJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/AsArrayJsonAdapter.java
@@ -1,0 +1,74 @@
+package com.squareup.moshi;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class AsArrayJsonAdapter<T> extends ClassJsonAdapter<T> {
+	public static final Factory FACTORY = new Factory() {
+		@Override
+		public @Nullable JsonAdapter<?> create(@Nonnull Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+			Class<?> rawType = getRawType(type, annotations);
+			if (rawType == null)
+				return null;
+			ClassFactory<Object> classFactory = ClassFactory.get(rawType);
+			Map<String, FieldBinding<?>> fields = new TreeMap<>();
+			for (Type t = type; t != Object.class; t = Types.getGenericSuperclass(t))
+				createFieldBindings(moshi, t, fields);
+			return new AsArrayJsonAdapter<>(classFactory, fields).nullSafe();
+		}
+	};
+
+	AsArrayJsonAdapter(ClassFactory<T> classFactory, Map<String, FieldBinding<?>> fieldsArray) {
+		super(classFactory, fieldsArray);
+	}
+
+	@Override
+	public T fromJson(JsonReader reader) throws IOException {
+		T result;
+		try {
+			result = classFactory.newInstance();
+		} catch (InstantiationException e) {
+			throw new RuntimeException(e);
+		} catch (InvocationTargetException e) {
+			Throwable targetException = e.getTargetException();
+			if (targetException instanceof RuntimeException) throw (RuntimeException) targetException;
+			if (targetException instanceof Error) throw (Error) targetException;
+			throw new RuntimeException(targetException);
+		} catch (IllegalAccessException e) {
+			throw new AssertionError();
+		}
+
+		try {
+			reader.beginArray();
+			for (FieldBinding<?> fieldBinding : fieldsArray) {
+				if (!reader.hasNext())
+					throw new RuntimeException("Array has too much element");
+				fieldBinding.read(reader, result);
+			}
+			reader.endArray();
+			return result;
+		} catch (IllegalAccessException e) {
+			throw new AssertionError();
+		}
+	}
+
+	@Override
+	public void toJson(JsonWriter writer, Object value) throws IOException {
+		try {
+			writer.beginArray();
+			for (FieldBinding<?> fieldBinding : fieldsArray)
+				fieldBinding.write(writer, value);
+			writer.endArray();
+		} catch (IllegalAccessException e) {
+	      throw new AssertionError();
+	    }
+	}
+}
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -41,10 +41,21 @@ import javax.annotation.Nullable;
  *   <li>scala.*
  * </ul>
  */
-final class ClassJsonAdapter<T> extends JsonAdapter<T> {
+class ClassJsonAdapter<T> extends JsonAdapter<T> {
   public static final JsonAdapter.Factory FACTORY = new JsonAdapter.Factory() {
     @Override public @Nullable JsonAdapter<?> create(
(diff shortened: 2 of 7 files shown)


NightlyNexus:
Thank you for your code and tests.

This functionality is very specific to a small use case.
If you need this behavior, you can copy ClassJsonAdapter into your project or maybe use JsonQualifiers.

I don't think Moshi wants to support this, but feel free to ask any question that arise when working on your project code on StackOverflow using the `moshi` tag.

tdbgamer:
Wait so the solution here was "everyone copy and paste this code every time you need it" for something that's a pretty common usecase? I wanted to switch to mochi since it looked interesting, but I guess I'll be sticking with Jackson then... This makes no sense guys.

JakeWharton:
> something that's a pretty common usecase?

[citation needed]

Also as stated you can already use a `@JsonQualifier` and `@ToJson`/`@FromJson` to replicate this without much trouble. I agree it doesn't seem worth supporting with a specific library feature.

tdbgamer:
> [citation needed]

[Simple github search turns up a few results](https://github.com/search?q=%22%40JsonFormat%28shape%22+%22%3D%22+%22JsonFormat.Shape.ARRAY%29%22&type=Code)

>Also as stated you can already use a @JsonQualifier and @ToJson/@FromJson to replicate this without much trouble

It does seem strange to me for a serialization library to force users to implement basic serialization themselves. So we just keep duplicating the same unnecessary code ad infinitum because "it doesn't seem worth it?" I'm sorry but this line of reasoning makes no sense to me.

JakeWharton:
We prefer smaller libraries with extensible primitives. If you need a library that wants to include every feature for every user then Jackson seems like a good fit for your need. Otherwise, we offer the tools to achieve this behavior for the rare case it's needed.

tdbgamer:
I agree that libraries should cover basic cases and offer extensibility for the rare cases. I just disagree that this is a rare scenario.

> If you need a library that wants to include every feature for every user then Jackson

Does a library have to either be so small as to push work off onto the user or so large as to encompass every possible usecase? I don't see why there can't be a middle ground. Specifying the "shape" (whether that shape be an array, string, etc.) an object will take after serialization in general seems like a pretty common usecase to me, if you guys disagree then that fine. I'm just offering my 2 cents.

swankjesse:
@tdbgamer  you can use AsArrayJsonAdapter without it being built-in. Moshi lets you plug-in custom behavior. 

I don‘t include this feature not only to keep Moshi small, but also because I think it’s a bad feature. The rules and policy on how to handle not-enough array elements or too many array elements is too confusing. I think `@ToJson` is a much better fit here.

```
object PizzaAdapter {
  @ToJson fun pizzaToJson(pizza: Pizza) = listOf<Any>(pizza.diameter, pizza.extraCheese)
  @FromJson fun jsonToPizza(list: List<Any>) = Pizza(list[0] as Int, list[1] as Boolean)
}
```

swankjesse:
Also the citation suggests this feature isn’t actually that common at all.

screen shot 2019-01-25 at 9 49 38 pm
screen shot 2019-01-25 at 9 49 17 pm

tdbgamer:
>The rules and policy on how to handle not-enough array elements or too many array elements is too confusing.

How this is confusing? I am interested to hear your reasoning. Other libraries seem to have solved this problem.

>Also the citation suggests this feature isn’t actually that common at all.

I think it was pretty clear that was a three second courtesy search showing people using it and was by no means a sophisticated or robust way of finding usage (doesn't even take into account differences in spacing, argument ording, etc.). You seem to have missed the point though.
The main point being:
>Specifying the "shape" (whether that shape be an array, string, etc.) an object will take after serialization in general seems like a pretty common usecase

Now let's search how many people use `@JsonFormat` to reshape their data.
image

That's quite a bit more. We could go further, but there's really no point. I doubt that was the part you actually cared about though, so let's get to the real point.

```kotlin
data class Pizza(val diameter: Int, val extraCheese: Boolean)

object PizzaAdapter {
  @ToJson fun pizzaToJson(pizza: Pizza) = listOf<Any>(pizza.diameter, pizza.extraCheese)
  @FromJson fun jsonToPizza(list: List<Any>) = Pizza(list[0] as Int, list[1] as Boolean)
}
```

vs

```kotlin
@JsonFormat(shape = JsonFormat.Shape.ARRAY)
data class Pizza(val diameter: Int, val extraCheese: Boolean)
```

This is clearly duplicating code, introducing the possibility for bugs and mistakes, increasing the refactor surface area for structures changes, etc. and are all problems for the adapter that are handily solved with a single annotation.

Now take a look at the code changes to add `pepperoni` to the pizza:
```kotlin
data class Pizza(val diameter: Int, val extraCheese: Boolean, val pepperoni: Boolean)

object PizzaAdapter {
  @ToJson fun pizzaToJson(pizza: Pizza) = listOf<Any>(pizza.diameter, pizza.extraCheese, val pepperoni: Boolean)
  @FromJson fun jsonToPizza(list: List<Any>) = Pizza(list[0] as Int, list[1] as Boolean, list[2] as Boolean)
}
```

Now let's remove diameter cause we only sell one size pizza:
```kotlin
data class Pizza(val diameter: Int, val extraCheese: Boolean, val pepperoni: Boolean)

object PizzaAdapter {
  @ToJson fun pizzaToJson(pizza: Pizza) = listOf<Any>(pizza.extraCheese, val pepperoni: Boolean)
  @FromJson fun jsonToPizza(list: List<Any>) = Pizza(list[1] as Boolean, list[2] as Boolean)
}
```

Oh no! We made a fatal mistake. The first element was deleted, but we didn't reindex every field after it and now deserializing will cause either IndexOutOfBounds on the last element, or a ClassCastException on a middle element of a differing type. Yet more time wasted debugging trivial serialization code and more mental overhead thrust upon the user.

One might also add a field to the `fromJson` and forget the `toJson` in which case serializing would "work" by producing bad data and the deserializer would fail, thereby obfuscating the error further (was the field the wrong type or were the wrong number of fields passed?).

I'm just scraping the surface here. Surely you can see how maintaining this code is a waste of time, not DRY, and will simply cause more problems than it solves. If you still insist that this is a "bad feature" and copy/pasting code is a better, cleaner solution, more power to you man.

This argument doesn't really matter to me anymore as my team has chosen to go with Jackson. I was simply surprised that someone offering up their free time to improve Moshi was glibly denied in favor of copy/pasting.

swankjesse:
> glibly

This reads like its intended to shame. Please don’t bring that here.

We intend to keep Moshi small and extensible. We discuss why here:
https://jakewharton.com/an-open-source-advantage/

If you want a large everything-included project please enjoy Jackson!

Finally – the reason I think this is a bad feature is because you can't add pepperoni or remove diameter without breaking compatibility. I think that's a bad trade-off.

tdbgamer:
>This reads like its intended to shame. Please don’t bring that here.

Not sure how that was interpreted as shame. Just seemed very strange to me. Maybe came across wrong over text.

>If you want a large everything-included project please enjoy Jackson!

[I already answered that line of reasoning earlier](https://github.com/square/moshi/pull/367#issuecomment-456496464)

Also doesn't seem super relevant to this particular feature. Does adding it make the library now huge and all-encompassing? It looks as though it hasn't even changed the scope of the project as Moshi already provides builtin adapters for the user. It seems to me that people tend to throw out the "I want my library to be small" argument only when it's a feature they don't personally like, and not because it will actually increase the scope of the project.

>Finally – the reason I think this is a bad feature is because you can't add pepperoni or remove diameter without breaking compatibility. I think that's a bad trade-off.

People will make breaking changes to their structures. Even without this feature any new fields have to be optional to not break, which isn't always an option. As long as the user understands the trade-offs, it is up to the user what is appropriate.

Imagine if Linux kernel said, "Synchronous IO is slow, so from now on all IO syscalls are now asynchronous because I think Synchronous IO has bad trade-offs." Seems like more of user concern, no?

Again as I said earlier, it's your library and your decision. I was simply expressing my opinion on it. If you disagree that's perfectly fine with me man. Overall I think your library is very cool and that you guys did good work. I am only saying these things because I wanted to use your project and was a little disappointed it didn't match my usecase.

PR #373, opened 2017-10-27
Respect nullability of collection element types.

NightlyNexus:
For Kotlin class properties and constructor parameters that are collections, respect the declared nullability of the elements.

just a proof of concept.
needs more tests because I get the feeling I'm missing something.
Also, I'm not sure if there's maybe a more general solution that's less of a workaround.

### DIFF
--- a/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
@@ -23,6 +23,7 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KParameter
 import kotlin.reflect.KProperty1
+import kotlin.reflect.KType
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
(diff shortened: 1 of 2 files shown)


kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -203,4 +204,55 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+    val collectionElementType = type.arguments[0].type!!
+    val elementAdapter = jsonAdapter(collectionElementType, moshi, Util.NO_ANNOTATIONS)
+    return collectionJsonAdapter(
+        elementAdapter, create, collectionElementType.isMarkedNullable) as JsonAdapter<Any?>
+  }
+
+  private inline fun <T> collectionJsonAdapter(elementAdapter: JsonAdapter<T?>,
+      crossinline create: () -> MutableCollection<T?>, nullableElements: Boolean):
+      JsonAdapter<Collection<T?>> {
+    return object : JsonAdapter<Collection<T?>>() {
```
NightlyNexus: needs fixing: creates a new JsonAdapter for every collection property instead of using the Moshi cache.

NightlyNexus:
i think i got ahead of myself.
more of Moshi should know about nullable types first before this.
going to return to this later.

fengdai:
When will Moshi support this feature?

PR #376, opened 2017-10-29
Clarify error for non-null Kotlin properties.

NightlyNexus:
Instead of throwing an InvocationTargetException.


swankjesse:
Nice tests.

Issue #293, opened 2017-04-30
KotlinJsonAdapter can get confused by parameter vs. property names


cbruegg:
Consider the following test, which fails with `java.lang.AssertionError: Expected <1>, actual <2>.`:

```kotlin
open class NonValParamSuper(val a: Int)
class NonValParam(a: Int, notA: Int = 2): NonValParamSuper(a = notA)

@Test fun kotlinNonValParam() {
  val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory).build()
  val adapter = moshi.adapter(NonValParam::class.java)

  val json = """{"a": 1, "b": 2}"""
  val deserialized = adapter.fromJson(json)
  assertEquals(1, deserialized.a)
}
```
The reason why this happens is because the `KotlinJsonAdapter` [assumes](https://github.com/square/moshi/blob/cd1542363db05ba00a53ec354d1f969fd997a0ab/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt#L176) that the constructor parameter name equals the property name, which isn't necessarily true. This may be a little of an edge case, but I thought it would be worth sharing.

When the `KotlinJsonAdapter` is documented, it should probably contain a warning about this. Another option would be asking JetBrains whether they could extend Kotlin reflection to allow access to an associated property of a constructor parameter in case it has one.

swankjesse:
Yeah, I think this important to track, but unlikely to be a problem in practice.

We should probably confirm that the types match.

swankjesse:
@NightlyNexus any further action on this?

PR #377, opened 2017-11-01
Crash earlier for property type conflicts.

NightlyNexus:
IllegalArgumentException was the best exception type I could think of.

### DIFF
--- a/kotlin/src/main/java/com/squareup/moshi/kotlin/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/kotlin/KotlinJsonAdapter.kt
@@ -206,6 +206,11 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
         continue
       }
 
+      if (parameter != null && parameter.type != property.returnType) {
+        throw IllegalArgumentException("'${property.name}' has a constructor parameter of type " +
+            "${parameter.type} but a property of type ${property.returnType}.")
+      }
+
       if (property !is KMutableProperty1 && parameter == null) continue
 
       property.isAccessible = true
--- a/kotlin/src/test/java/com/squareup/moshi/kotlin/KotlinJsonAdapterTest.kt
+++ b/kotlin/src/test/java/com/squareup/moshi/kotlin/KotlinJsonAdapterTest.kt
@@ -346,6 +346,21 @@ class KotlinJsonAdapterTest {
     var b: Int = -1
   }
 
+  @Test fun constructorParametersAndPropertiesWithSameNamesMustHaveSameTypes() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+    try {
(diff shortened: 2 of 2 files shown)


kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
```diff
@@ -296,6 +296,39 @@ class KotlinJsonAdapterTest {
...
+  }
+
+  @Test fun constructorParametersAndPropertiesWithSameNamesMustHaveSameTypes() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+    try {
+      moshi.adapter(ConstructorParameterWithSameNameAsPropertyButDifferentType::class.java)
+      fail()
+    } catch (expected: IllegalArgumentException) {
+      assertThat(expected).hasMessage(
+          "a has a constructor parameter of type kotlin.Int but property of type kotlin.String.")
```
JakeWharton: Probably worth single-quoting names. "a has a" reads funny.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -167,6 +167,12 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       if (Modifier.isTransient(property.javaField?.modifiers ?: 0)) continue
 
       val parameter = parametersByName[property.name]
+
+      if (parameter != null && parameter.type != property.returnType) {
```
NightlyNexus: I think this is insufficient. We have to compare type qualifiers, too.

kotlin/src/test/java/com/squareup/moshi/KotlinJsonAdapterTest.kt
```diff
@@ -296,6 +296,39 @@ class KotlinJsonAdapterTest {
...
+    val encoded = ConstructorParameterWithSameNameAsProperty(5)
+    assertThat(jsonAdapter.toJson(encoded)).isEqualTo("{\"a\":-1,\"b\":5}")
+
+    val decoded = jsonAdapter.fromJson("{\"a\":5}")!!
+    assertThat(decoded.a).isEqualTo(-1)
+    assertThat(decoded.b).isEqualTo(5)
+  }
+
+  class ConstructorParameterWithSameNameAsProperty(a: Int) {
+    var a = -1
```
NightlyNexus: I have a few concerns about this behavior that I want to check. Like what happens with type qualifiers or transient?
I'll look into it, ask questions, and follow up.

kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
NightlyNexus: no, never mind me.

NightlyNexus:
tracking/noting other limitations in: #293

PR #382, opened 2017-11-05
Fix ignoring @Json on supertype constructors.

### LINKED ISSUE #372: Document Kotlin types shouldn’t subclass Java types and vice-versa
I'm annotating a field in a Kotlin class (using `moshi-kotlin`) with `@Json(name = "xyz")` and serializing produces the expected result, but when I try to serialize a subclass the annotation from the parent class is not applied.

Example:
```
class Container(@Json(name = "parent") var abc1: Parent = Child(),
                @Json(name = "child") var abc2: Child = Child())

open class Parent(@Json(name = "xyz") var abc: String = "")

class Child : Parent()
```

...results in:
```
{
  "parent": {
    "xyz": ""
  },
  "child": {
    "abc": ""
  }
}
```

expected result:
```
{
  "parent": {
    "xyz": ""
  },
  "child": {
    "xyz": ""
  }
}
```

Is this the intended behavior?

NightlyNexus:
Closes #372.

Probably needs some eyes to see if this logic can be cleaned up (or maybe clarity would be worth the cost of recursion here).

### DIFF
--- a/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
+++ b/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
@@ -26,6 +26,7 @@ import kotlin.reflect.KProperty1
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
+import kotlin.reflect.full.superclasses
 import kotlin.reflect.jvm.isAccessible
 import kotlin.reflect.jvm.javaField
 import kotlin.reflect.jvm.javaType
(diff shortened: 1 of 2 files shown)


swankjesse:
Not sure about this. At a minimum whatever solution we find should also honor `@JsonQualifier` annotations on superclass constructor parameters if we’re honoring `@Json` there.

I think we’re better off requesting callers target their `@Json` annotation properly. And we should figure out a better way to make that easier.

NightlyNexus:
Yeah, I like Egor's idea about an annotation that targets properties.
I'm going to upload a fix for the qualifiers for future travelers and close this out.

PR #383, opened 2017-11-05
Add @CheckReturnValue to appropriate public APIs.

NightlyNexus:
Moshi's a good small library to try this out on.
If this looks good, we could file issues on other libraries to get these in.  If this doesn't look good, no time was lost.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -30,13 +30,13 @@ import okio.BufferedSource;
  * Converts Java values to JSON, and JSON values to Java.
  */
 public abstract class JsonAdapter<T> {
-  public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
+  @CheckReturnValue public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
 
-  public final @Nullable T fromJson(BufferedSource source) throws IOException {
+  @CheckReturnValue public final @Nullable T fromJson(BufferedSource source) throws IOException {
     return fromJson(JsonReader.of(source));
   }
 
-  public final @Nullable T fromJson(String string) throws IOException {
+  @CheckReturnValue public final @Nullable T fromJson(String string) throws IOException {
     return fromJson(new Buffer().writeUtf8(string));
   }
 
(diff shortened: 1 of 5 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -30,13 +30,13 @@
  * Converts Java values to JSON, and JSON values to Java.
  */
 public abstract class JsonAdapter<T> {
-  public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
+  @CheckReturnValue public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
```
NightlyNexus: There may be some use case for having a delegate adapter that throws away a value you don't care about, but that seems weird enough that a suppression in app code with an explanation is warranted.

adriancole:
maybe for the lazy c&p what exactly checkReturnValue implies?

NightlyNexus:
@adriancole
It will give a hint in the IDE if you forget to use assign the result of the method call. (It's usually applied to factory methods and methods with no side effects.)

There's also an Error Prone check: http://errorprone.info/bugpattern/CheckReturnValue

Here's a real example where `@CheckReturnValue` prevents mistakes: https://github.com/square/moshi/pull/353

NightlyNexus:
>This annotation is used to denote a method whose return value should always be checked after invoking the method.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
swankjesse: definitely. `reader.skipValue()` is better.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -143,7 +143,7 @@ public final void toJson(BufferedSink sink, @Nullable T value) throws IOExceptio
   }
 
   /** Returns a JSON adapter equal to this, but is lenient when reading and writing. */
-  public final JsonAdapter<T> lenient() {
+  @CheckReturnValue public final JsonAdapter<T> lenient() {
```
swankjesse: these ones are particularly useful

adriancole:
thx for the details. looks very useful (though seems like could apply to
hordes of methods)

PR #384, opened 2017-11-05
Update to Error Prone 2.1.2.

NightlyNexus:


NightlyNexus:
probably have to clear the cache to update the compiler plugin, but i really just want EP updated. just going to revert.

PR #390, opened 2017-11-09
Let JsonValueReader.nextString read numbers.

### LINKED ISSUE #388: JsonValueReader.nextString() does not coerce numbers to strings.
```java 
@Test public void numberToStringCoersion() throws Exception {
  JsonReader reader = newReader("[0, 9223372036854775807, 1.5]");
  reader.beginArray();
  assertThat(reader.nextString()).isEqualTo("0");
  assertThat(reader.nextString()).isEqualTo("9223372036854775807");
  assertThat(reader.nextString()).isEqualTo("1.5");
  reader.endArray();
}
```

This test passes with a JsonUtf8Reader and fails (`com.squareup.moshi.JsonDataException: Expected STRING but was 0.0, a java.lang.Double, at path $[0]`) with a JsonValueReader.

NightlyNexus:
This adds parity with JsonUtf8Reader and lets big number literals in JSON be read in as strings in Java.

PR note:
```java
JsonAdapter<String> adapter = moshi.adapter(String.class);
JsonReader reader = JsonReader.of(new Buffer().writeUtf8("9223372036854775808"));
String bigMoney = adapter.fromJsonValue(reader.readJsonValue());
```
This change will make the above work (instead of throwing, expecting a double), so that you can use big numbers in polymorphic adapters where you need to read in the map.

Closes #388.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
@@ -152,9 +152,19 @@ final class JsonValueReader extends JsonReader {
   }
 
   @Override public String nextString() throws IOException {
-    String peeked = require(String.class, Token.STRING);
-    remove();
-    return peeked;
+    Object peeked = (stackSize != 0 ? stack[stackSize - 1] : null);
+    if (peeked instanceof String) {
+      remove();
+      return (String) peeked;
+    }
+    if (peeked instanceof Number) {
+      remove();
+      return peeked.toString();
+    }
+    if (peeked == JSON_READER_CLOSED) {
+      throw new IllegalStateException("JsonReader is closed");
+    }
+    throw typeMismatch(peeked, Token.STRING);
   }
 
   @Override public int selectString(Options options) throws IOException {
--- a/moshi/src/test/java/com/squareup/moshi/JsonUtf8ReaderTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonUtf8ReaderTest.java
@@ -336,6 +336,18 @@ public final class JsonUtf8ReaderTest {
     assertEquals("-0", reader.nextString());
   }
 
+  @Test public void numberToStringCoersion() throws Exception {
+    JsonReader reader = newReader("[0, 9223372036854775807, 2.5, 3.010, \"a\", \"5\"]");
+    reader.beginArray();
+    assertThat(reader.nextString()).isEqualTo("0");
+    assertThat(reader.nextString()).isEqualTo("9223372036854775807");
+    assertThat(reader.nextString()).isEqualTo("2.5");
+    assertThat(reader.nextString()).isEqualTo("3.010");
+    assertThat(reader.nextString()).isEqualTo("a");
+    assertThat(reader.nextString()).isEqualTo("5");
+    reader.endArray();
+  }
+
   @Test public void quotedNumberWithEscape() throws IOException {
     JsonReader reader = newReader("[\"12\u00334\"]");
     reader.setLenient(true);
--- a/moshi/src/test/java/com/squareup/moshi/JsonValueReaderTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonValueReaderTest.java
@@ -425,7 +425,8 @@ public final class JsonValueReaderTest {
   }
 
   @Test public void numberToStringCoersion() throws Exception {
-    JsonReader reader = new JsonValueReader(Arrays.asList(0, 9223372036854775807L, 2.5d, 3.010f, "a", "5"));
+    JsonReader reader =
+        new JsonValueReader(Arrays.asList(0, 9223372036854775807L, 2.5d, 3.01f, "a", "5"));
     reader.beginArray();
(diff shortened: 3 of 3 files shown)


JakeWharton:
Looking on my computer now I guess it's hard to do both because of the difference in initialization.

JakeWharton:
Looking on my computer now I guess it's hard to do both because of the difference in initialization.

NightlyNexus:
Yeah, but the differences make a good point:
Since JsonValueReader is only exposed via `readJsonValue()`, doubles are the only numbers.
I don't think this approach really helps users.

NightlyNexus:
Er, via JsonAdapter.fromJsonValue, but I don't know how many people are using that without JsonReader.readJsonValue(). (though i might be biased in my use cases.)

NightlyNexus:
there's probably no use in getting strings, since you'll have lose precision in a value reader (like in my PR comment example), no matter what you do.

NightlyNexus:
Sorry for the delay here.
Also, sorry, I conflated too many things here.
The context is that I'd like to blog about polymorphic deserialization and unusual JSON formats.
</br>
</br>
I would like this code
```java
// "Buffered" code.
JsonAdapter<Object> objectAdapter = moshi.adapter(Object.class);
JsonAdapter<String> numberAdapter = moshi.adapter(String.class);
String number = numberAdapter.fromJsonValue(objectAdapter.fromJson("17"));
```
to result the same as this code
```java
// "Streaming" code.
JsonAdapter<String> numberAdapter = moshi.adapter(String.class);
String number = numberAdapter.fromJson("17");
```

This PR does that. (The code isn't quite right but could be fixed.)
</br>
</br>
There's a twist. This might be out of the scope of discussion here.

There cannot be analogous "buffered" code for the "streaming" code below because we prefer doubles for number types. (Gson had its own `LazilyParsedNumber`.) Moshi requires registering a custom adapter to get the same behavior across the "streaming" code and "buffered" code.
```
// "Streaming" code.
JsonAdapter<String> stringAdapter = moshi.adapter(String.class);
String number = stringAdapter.fromJson("9223372036854775808");
```
</br>
</br>

2 options I see:
- Close this issue. Leave the first and second example code above broken. Recommend using a custom adapter.
- Patch to make the first code example work. Recommend using a custom adapter if expecting big numbers.

moshi/src/test/java/com/squareup/moshi/JsonValueReaderTest.java
```diff
@@ -423,6 +424,18 @@
     }
   }
 
+  @Test public void numberToStringCoersion() throws Exception {
+    JsonReader reader = new JsonValueReader(Arrays.asList(0, 9223372036854775807L, 2.5d, 3.010f, "a", "5"));
```
swankjesse: is there any reason to have the trailing `0` in `3.010f` ?

swankjesse:
I like this as-is.

As a general rule, I don’t want to encourage people to put numeric values into JSON that can’t be represented as a `double`. That is asking for trouble.

moshi/src/test/java/com/squareup/moshi/JsonValueReaderTest.java
NightlyNexus: nope, it's copy-paste from the other test case. will remove.

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -152,9 +152,19 @@
...
-    remove();
-    return peeked;
+    Object peeked = (stackSize != 0 ? stack[stackSize - 1] : null);
+    if (peeked instanceof String) {
+      remove();
+      return (String) peeked;
+    }
+    if (peeked instanceof Number) {
+      remove();
+      return peeked.toString();
```
NightlyNexus: this should probably be `BigDecimal(peeked).toPlainString()`.

NightlyNexus:
>As a general rule, I don’t want to encourage people to put numeric values into JSON that can’t be represented as a double. That is asking for trouble.

Thanks, I'm going to note this when writing about this. 👍

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
NightlyNexus: no, dunno what i was thinking.

moshi/src/test/java/com/squareup/moshi/JsonUtf8ReaderTest.java
```diff
@@ -336,6 +336,18 @@ private void assertNotANumber(String s) throws IOException {
     assertEquals("-0", reader.nextString());
   }
 
+  @Test public void numberToStringCoersion() throws Exception {
+    JsonReader reader = newReader("[0, 9223372036854775807, 2.5, 3.010, \"a\", \"5\"]");
+    reader.beginArray();
+    assertThat(reader.nextString()).isEqualTo("0");
+    assertThat(reader.nextString()).isEqualTo("9223372036854775807");
+    assertThat(reader.nextString()).isEqualTo("2.5");
+    assertThat(reader.nextString()).isEqualTo("3.010");
```
swankjesse: The retained precision here is interesting!

PR #393, opened 2017-11-15
Added url to POJO generation plugin

robohorse:
Hi, i added support to Moshi. I think, that it can may be helpful for someone :)
https://plugins.jetbrains.com/plugin/8634-robopojogenerator
https://github.com/robohorse/RoboPOJOGenerator


swankjesse:
I don't think this is on topic for Moshi.

PR #397, opened 2017-11-24
Remove trailing zeros and exponent notation from JsonValueReader.next…

NightlyNexus:
…String.

I remembered what I was worried about [here](https://github.com/square/moshi/pull/390#discussion_r152438517). sorry about that.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
@@ -159,7 +159,7 @@ final class JsonValueReader extends JsonReader {
     }
     if (peeked instanceof Number) {
       remove();
-      return peeked.toString();
+      return new BigDecimal(peeked.toString()).stripTrailingZeros().toPlainString();
     }
     if (peeked == JSON_READER_CLOSED) {
       throw new IllegalStateException("JsonReader is closed");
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -159,7 +159,7 @@
     }
     if (peeked instanceof Number) {
       remove();
-      return peeked.toString();
+      return new BigDecimal(peeked.toString()).stripTrailingZeros().toPlainString();
```
swankjesse: Hmmm . . . I’m not sure this is worth the effort? Also interesting for values like 23.45e-1. What do our users want?

NightlyNexus: Good point! `23.45e-1` will result in the string `2.345` in both these cases. It seems best for users not to care about the string format.

I had something like this, and I was concerned about my `value` string being `1.0E7`, but whatever it ends up getting used for shouldn't care about the string format.

```java
JsonAdapter<Object> objectAdapter = moshi.adapter(Object.class);
JsonAdapter<String> adapter = moshi.adapter(String.class);
Object jsonElement = objectAdapter.fromJson("10000000");
String value = adapter.fromJsonValue(jsonElement);
```

Thanks for the help.

PR #399, opened 2017-11-27
Make selectString consistent across JsonReaders.

NightlyNexus:
Make JsonValueReader.selectString return -1 for non-strings instead of throwing.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
@@ -168,9 +168,18 @@ final class JsonValueReader extends JsonReader {
   }
 
   @Override public int selectString(Options options) throws IOException {
-    String peeked = require(String.class, Token.STRING);
+    Object peeked = (stackSize != 0 ? stack[stackSize - 1] : null);
+
+    if (!(peeked instanceof String)) {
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -168,9 +168,20 @@
   }
 
   @Override public int selectString(Options options) throws IOException {
-    String peeked = require(String.class, Token.STRING);
+    Object peeked = (stackSize != 0 ? stack[stackSize - 1] : null);
+    String peekedString;
+
+    if ((peeked instanceof String)) {
```
swankjesse: nit: extra parens?

swankjesse: or, if you invert this if, `if (!(peeked instance of String)) { ... }` you can avoid separate declaration & assignment for peekedString

swankjesse:
Good find!

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
NightlyNexus: Yep, thanks!

PR #401, opened 2017-11-29
Update kotlin to 1.2.10

StefMa:


StefMa:
@NightlyNexus can you pls take a look into it? 🙃

swankjesse:
Upgrading our major Kotlin version requires all of our downstream users to upgrade. I’m not sure whether we want to force that on them at this time.

JakeWharton:
That's not strictly true:
https://kotlinlang.org/docs/reference/compatibility.html#compatibility-guarantees-for-kotlin-releases

Provided you don't use any new APIs, of course!

swankjesse:
Of course. I assume that if we depend on Kotlin 1.2, we’re implicitly saying that it’s our minimum.

StefMa:
Where the problem to saying that it is the minimum of moshi? 🤔
Since 1.2 is backward compatible to 1.1 it don't prevent users to update
their kotlin dependency.

swankjesse:
We don’t want to force users to update their Kotlin dependency in order to update their Moshi dependency. It’s quite possible that upgrading from Kotlin 1.1 to Kotlin 1.2 is difficult or dangerous, and since Moshi itself works with both we can leave that to our users.

You could imagine a webpage saying “You need Chrome 63+ or Firefox 57+ to view this page” but doesn’t actually require features from those new browsers. Upgrading is great! But the requirement to upgrade is unnecessary.

swankjesse:
Yep, exactly.

swankjesse:
I’m gonna close this PR, though we will probably require Kotlin 1.2 soon.

PR #406, opened 2017-12-02
Don't handle WildcardTypes in ClassJsonAdapter.

### LINKED ISSUE #404: Better error message for supertypes/lower-bounded wildcards.
```java
new Moshi.Builder().build().adapter(Types.supertypeOf(String.class));
```
will crash because the ClassJsonAdapter factory will try to create an adapter and fail checking the supertype. ([Here's a case of this accident.](https://github.com/square/moshi/issues/400#issuecomment-347610491))

This should fail with a clear message (there's nothing obvious to do with `? super Foo`), and the ClassJsonAdapter factory shouldn't try to handle it.

Note that the ClassJsonAdapter also tries to handle subtypes, which results in an error while serializing.
For subtypes, maybe the upper bound type should be looked up.

Maybe `Moshi.adapter(...)` should check for wild card types and handle this logic.

NightlyNexus:
Fixes #404.

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/MoshiTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/MoshiTest.java
@@ -533,6 +533,26 @@ public final class MoshiTest {
     assertThat(adapter.toJson(null)).isEqualTo("null");
   }
 
+  @Test public void upperBoundedWildcardsAreNotHandled() {
+    Moshi moshi = new Moshi.Builder().build();
+    try {
+      moshi.adapter(Types.subtypeOf(String.class));
+      fail();
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessage("No JsonAdapter for ? extends java.lang.String annotated []");
+    }
+  }
+
+  @Test public void lowerBoundedWildcardsAreNotHandled() {
+    Moshi moshi = new Moshi.Builder().build();
+    try {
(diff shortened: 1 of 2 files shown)


moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -533,6 +533,26 @@
     assertThat(adapter.toJson(null)).isEqualTo("null");
   }
 
+  @Test public void upperBoundedWildcardsAreNotHandled() {
+    Moshi moshi = new Moshi.Builder().build();
+    try {
+      moshi.adapter(Types.subtypeOf(String.class));
```
NightlyNexus: There could be a delegating factory that gives the string JsonAdapter for this type, but I don't see a use case.

Would somebody with a `List<?> objects` expect the object JsonAdapter to handle this? I doubt it, and I haven't seen anybody ask for it.

swankjesse:
I like this. Hopefully it doesn’t cause too much grief for users with collections like `List<? extends Taco>`.

NightlyNexus:
@swankjesse We'll need to call this out in the changelog.
```java
class Box<T> {
  T data;
}
```
used to work because the field binding for `data` would use JsonAdapter for Object, due to ClassJsonAdapter's implementation.

It now fails to find a matching adapter for the type.

This is not what a user intends to do, but somebody might have this code.

PR #422, opened 2018-01-10
Fix ClassJsonAdapter to handle ParameterizedTypes.

NightlyNexus:
I broke this in https://github.com/square/moshi/pull/406

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -20,6 +20,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -45,8 +46,10 @@
   public static final JsonAdapter.Factory FACTORY = new JsonAdapter.Factory() {
     @Override public @Nullable JsonAdapter<?> create(
         Type type, Set<? extends Annotation> annotations, Moshi moshi) {
-      if (!(type instanceof Class)) return null;
-      Class<?> rawType = (Class<?>) type;
+      if (!(type instanceof Class || type instanceof ParameterizedType)) {
```
swankjesse: I find this kind of expression difficult to read; like a double negative. For whatever reason (maybe just me) I find the contrapositive easier.

```
if (!(type instanceof Class) && !(type instanceof ParameterizedType)) {
  return null;
}
```

NightlyNexus: totally. thanks for point that out. #444

NightlyNexus: totally. thanks for pointing that out. #444

PR #423, opened 2018-01-11
Improve error message for local classes.

NightlyNexus:


moshi/src/test/java/com/squareup/moshi/ClassJsonAdapterTest.java
```diff
@@ -341,8 +341,23 @@ public MagicConstructorWithDefaultField(Void argument) {
...
+          + "com.squareup.moshi.ClassJsonAdapterTest$1Local");
+    }
+  }
+
+  interface Interface {
+  }
+
   @Test public void interfaceNotSupported() throws Exception {
-    assertThat(ClassJsonAdapter.FACTORY.create(Runnable.class, NO_ANNOTATIONS, moshi)).isNull();
+    assertThat(ClassJsonAdapter.FACTORY.create(Interface.class, NO_ANNOTATIONS, moshi)).isNull();
```
NightlyNexus: sneaking this in to avoid confusion with a platform type. can revert, if preferable.

PR #424, opened 2018-01-11
Disallow irregular Kotlin classes.

NightlyNexus:


kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt
```diff
@@ -162,17 +162,32 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
 
     val rawType = Types.getRawType(type)
+    if (rawType.isInterface) return null
     if (rawType.isEnum) return null
     if (!rawType.isAnnotationPresent(KOTLIN_METADATA)) return null
     if (ClassJsonAdapter.isPlatformType(rawType)) return null
 
-    val constructor = rawType.kotlin.primaryConstructor ?: return null
+    if (rawType.isLocalClass) {
+      throw IllegalArgumentException("Cannot serialize local class or object expression ${rawType.name}")
```
NightlyNexus: I looked earlier today for a way to differentiate these types, but I'm not sure you can, given the way expressions are implemented.

swankjesse: works for me

PR #434, opened 2018-02-07
Move modules into their own packages.

swankjesse:
This sets the Automatic-Module-Name for moshi, moshi-adapters, and moshi-kotlin.
It moves moshi-adapters into its own .adapters package and forwards the existing
adapter. It moves the moshi-kotlin into its own .kotlin package and forwards the
existing adapter.

I'm not certain this is necessary or sufficient, but I think it's the right idea
for JPMS compatibility.


adapters/src/main/java/com/squareup/moshi/adapters/Rfc3339DateJsonAdapter.java
```diff
@@ -0,0 +1,38 @@
+/*
```
swankjesse: no new code here, just GitHub not tracking a copy

kotlin/src/main/java/com/squareup/moshi/kotlin/KotlinJsonAdapter.kt
```diff
@@ -0,0 +1,245 @@
+/*
```
swankjesse: no new code here either.

JakeWharton:
Not sure if the `Util` will work without real module definitions.

Issue #438, opened 2018-02-13
Does not fail gracefully


robstoll:
In contrast, it parses all most everything and does not provide me the safety I expected :(
Consider the following (Kotlin):
```
data class A(val a: String)

fun main(vararg args: String) {
    val moshi = Moshi.Builder().build()
    val test = moshi.adapter(A::class.java).fromJson("""{"notEvenAPropertyOfA": "1.2.0"} <!-- comment in JSON? what about the superfluous } --> }""")
}
```
I end up with an `A` whose `a` is null
Do I need to set a configuration or something so that it fails [as describe in the documentation](https://github.com/square/moshi#fails-gracefully)?

*Edit*: for one, see next comment, this explains why the validation for `a` should not be `null` was not carried out. Secondly, it is actually good, that additional fields are ignored (makes it easier to evolve the JSON). However, the comment (which is illegal in JSON as far as I know) and the superfluous `}` should cause an Exception

robstoll:
I figured out why I do not get any complaints about an absent `a`. I need to use `KotlinJsonAdapterFactory` so that I get the complaint. Yet, the documentation states, I only need it, if I use annotations. That should be fixed IMO. Nevertheless, even if I provide an `a` as follows, the rest is still mall-formed and I expect an exception:
```
fun main(vararg args: String) {
    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
    val test = moshi.adapter(A::class.java).fromJson("""{"a": "1", "notEvenAPropertyOfA": "1.2.0"} <!-- comment in JSON? what about the superfluous } --> }""")
}
```

NightlyNexus:
The adapter doesn't fail on the trailing bracket because adapters read only their part of the document. 

Many adapters read from the same document and JsonReader, so a user may have another adapter that continues reading (which would fail in your example because of the extra bracket).

If I have the JSON `["hello", "world"]`, I might have a string adapter that reads the elements, but the string adapter should not consider the next token after reading.

If you want to validate that the document is consumed:
```java
JsonReader reader = JsonReader.of(new Buffer().writeUtf8(json));
Foo value = fooAdapter.fromJson(reader);
boolean consumed = reader.peek() == JsonReader.Token.END_DOCUMENT;
```

robstoll:
[...]

PR #439, opened 2018-02-13
mention that KotlinJsonAdapterFactory validates

robstoll:
See #438 for the use case. The documentation was not accurate enough IMO. Thus, mention that KotlinJsonAdapterFactory is required for validation.


swankjesse:
Thanks.

PR #441, opened 2018-02-14
JsonAdapter.fromJson(String) must fully consume.

NightlyNexus:
I hope this doesn't surprise anybody with existing broken adapters.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -30,6 +30,16 @@ import okio.BufferedSource;
  * Converts Java values to JSON, and JSON values to Java.
  */
 public abstract class JsonAdapter<T> {
+  final boolean lenient;
+
+  public JsonAdapter() {
+    lenient = false;
+  }
+
+  private JsonAdapter(boolean lenient) {
+    this.lenient = lenient;
+  }
+
   @CheckReturnValue public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
 
   @CheckReturnValue public final @Nullable T fromJson(BufferedSource source) throws IOException {
@@ -37,7 +47,12 @@ public abstract class JsonAdapter<T> {
   }
 
   @CheckReturnValue public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    JsonReader reader = JsonReader.of(new Buffer().writeUtf8(string));
+    T result = fromJson(reader);
+    if (!lenient && reader.peek() != JsonReader.Token.END_DOCUMENT) {
+      throw new JsonDataException("JSON document was not fully consumed.");
+    }
+    return result;
   }
 
   public abstract void toJson(JsonWriter writer, @Nullable T value) throws IOException;
@@ -96,7 +111,7 @@ public abstract class JsonAdapter<T> {
    */
   @CheckReturnValue public final JsonAdapter<T> serializeNulls() {
     final JsonAdapter<T> delegate = this;
-    return new JsonAdapter<T>() {
+    return new JsonAdapter<T>(delegate.lenient) {
       @Override public @Nullable T fromJson(JsonReader reader) throws IOException {
         return delegate.fromJson(reader);
       }
@@ -121,7 +136,7 @@ public abstract class JsonAdapter<T> {
    */
   @CheckReturnValue public final JsonAdapter<T> nullSafe() {
     final JsonAdapter<T> delegate = this;
-    return new JsonAdapter<T>() {
+    return new JsonAdapter<T>(delegate.lenient) {
       @Override public @Nullable T fromJson(JsonReader reader) throws IOException {
         if (reader.peek() == JsonReader.Token.NULL) {
           return reader.nextNull();
@@ -145,7 +160,7 @@ public abstract class JsonAdapter<T> {
   /** Returns a JSON adapter equal to this, but is lenient when reading and writing. */
   @CheckReturnValue public final JsonAdapter<T> lenient() {
     final JsonAdapter<T> delegate = this;
-    return new JsonAdapter<T>() {
+    return new JsonAdapter<T>(true) {
       @Override public @Nullable T fromJson(JsonReader reader) throws IOException {
         boolean lenient = reader.isLenient();
         reader.setLenient(true);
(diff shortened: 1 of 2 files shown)


NightlyNexus:
References #438

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -37,7 +37,12 @@
   }
 
   @CheckReturnValue public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    JsonReader reader = JsonReader.of(new Buffer().writeUtf8(string));
+    T result = fromJson(reader);
+    if (reader.peek() != JsonReader.Token.END_DOCUMENT) {
+      throw new JsonDataException("JSON document was not fully consumed.");
```
robstoll: I would include the next token as additional hint

robstoll: usually a JsonEncodingException is thrown for invalid data

NightlyNexus: Sounds good, as long as it's consistent with the check in Retrofit's converter.
@swankjesse Any objections?

NightlyNexus:
>Worth disabling the check if the adapter is lenient?

I'm not sure. JsonReaders can be lenient to accept more input. This check seems like a different type of leniency?

On an implementation note, adapters don't currently know if they are lenient or not. They could with a few more lines of code, though. Just let me know if you'd like to see it.

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
```diff
@@ -182,4 +183,22 @@
     serializeNulls.toJson(writer, Collections.<String, String>singletonMap("a", null));
     assertThat(factory.json()).isEqualTo("{\"a\":null}");
   }
+
+  @Test public void stringDocumentMustBeFullyConsumed() throws IOException {
+    JsonAdapter<String> brokenAdapter = new JsonAdapter<String>() {
+      @Override public String fromJson(JsonReader reader) throws IOException {
+        return "Forgot to call reader.nextString().";
```
swankjesse: neat

swankjesse:
I think I’d still like `fromJson(string)` to not fail if the JsonAdapter is lenient. But this is an improvement regardless.

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
```diff
@@ -182,4 +183,22 @@
...
+      }
+
+      @Override public void toJson(JsonWriter writer, @Nullable String value) throws IOException {
+        throw new AssertionError();
+      }
+    };
+    try {
+      brokenAdapter.fromJson("\"value\"");
```
swankjesse: This is actually somewhat of a surprising test. I’d expect a regular JsonAdapter and input of like `true true` or something

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -37,7 +37,12 @@
   }
 
   @CheckReturnValue public final @Nullable T fromJson(String string) throws IOException {
-    return fromJson(new Buffer().writeUtf8(string));
+    JsonReader reader = JsonReader.of(new Buffer().writeUtf8(string));
+    T result = fromJson(reader);
+    if (reader.peek() != JsonReader.Token.END_DOCUMENT) {
```
NightlyNexus: good call on making lenient adapters skip this check. we would need the lenient state, anyway, due to to the reading here.

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
NightlyNexus: do you have a test case you'd like to see?
making the new check only for non-lenient adapters will mean `true true` will either (in a strict adapter) fail as invalid JSON or (in a lenient adapter) won't be checked.

swankjesse: Yeah I think that's exactly how the test should look.

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
```diff
@@ -182,4 +183,36 @@
...
+  @Test public void lenientAdapterDoesNotCheckStringDocumentConsumption() throws IOException {
+    JsonAdapter<String> brokenAdapter = new JsonAdapter<String>() {
+      @Override public String fromJson(JsonReader reader) throws IOException {
+        return "Forgot to call reader.nextString().";
+      }
+
+      @Override public void toJson(JsonWriter writer, @Nullable String value) throws IOException {
+        throw new AssertionError();
+      }
+    }.lenient();
```
swankjesse: I think this is broken. Try lenient().nullSafe().

NightlyNexus: ahaha, thanks for the check. i'm going to try my original thought and make an overridable implementation of fromJson(String).

swankjesse: Maybe just a package-private method, `isLenient()` that returns false by default, returns true for lenient adapters, and delegates for other wrappers?

NightlyNexus: yep, i don't see a way to let the wrappers override fromJson(String) correctly.
isLenient is easier.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -223,12 +240,19 @@ public final void toJson(BufferedSink sink, @Nullable T value) throws IOExceptio
...
+      @Override boolean isLenient() {
+        return delegate.isLenient();
+      }
       @Override public String toString() {
         return delegate + ".indent(\"" + indent + "\")";
       }
     };
   }
 
+  boolean isLenient() {
```
NightlyNexus: I did this with a field for fewer lines of code, but changed it to a method in 4c4c71910381f9dad7871475cc7cbb0c1e95468a because it seems clearer than using a constructor parameter.

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
NightlyNexus: roger. I left this test because it's the only case where the message `"JSON document was not fully consumed."` can ever be seen. I updated with `true true` tests.

swankjesse:
Perfect!

luciofm:
> I hope this doesn't surprise anybody with existing broken adapters.

Hahaha, just surprised me using a snapshot build on Rocket.Chat

PR #442, opened 2018-02-14
Add example for custom qualifier with an element.

NightlyNexus:
It'd be neat to have an example with a complex JsonQualifier.

This thing I hacked together might be distracting in the implementation. I can try some other ideas for an example, too...


NightlyNexus:
Yep, good thought. It would be surprising to a user when that doesn't work.
I think the class field name being the outermost key is awkward, too.

I don't have any creative ideas to make this generalizable.

PR #447, opened 2018-02-22
Made Types.mapKeyAndValueTypes public

nedtwigg:
Helps a lot for adding third-party adapters for e.g. ImmutableMap.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Types.java
+++ b/moshi/src/main/java/com/squareup/moshi/Types.java
@@ -360,7 +360,7 @@ public final class Types {
    * Returns a two element array containing this map's key and value types in positions 0 and 1
    * respectively.
    */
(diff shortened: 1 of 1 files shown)


JakeWharton:
I'm not sure exposing this is the right thing to do. It's not sufficiently general (and neither is the existing collection one). What if I want to serialize Guava's `Table` which has three generic params?

nedtwigg:
Fair enough :)

JakeWharton:
File a bug at least so we can track it?

nedtwigg:
Will do once I have a good story on the complete usecase.  I may have other ways to solve my problem - just saw that similar methods in this class were public and figured it might have been an oversight that this one wasn't as well.

nedtwigg:
I have described the issue better in #449.  And, as you correctly noted, the approach I took here was too specific, and there is a better, more general method, which I made public in #448.

PR #448, opened 2018-02-22
Made Types::getSupertype public.

nedtwigg:
This makes it easier to create JsonAdapters for generic types
that delegate to other existing JsonAdapters.


nedtwigg:
Makes #449 possible.

swankjesse:
I’m _slightly_ reluctant to do this because it opens the doors to us being a more complete reflection library. We mostly want to be a JSON library though we need _some_ reflection stuff to get by.

nedtwigg:
As pointed out by @swankjesse in #449, Guava's TypeToken works great for this usecase.

PR #450, opened 2018-02-23
Fail earlier for some incorrect owner types.

NightlyNexus:


swankjesse:
Does this ever happen?

swankjesse:
Seems like too much code for a case we don’t even care about.

NightlyNexus:
```kotlin
class Foo {
  class Box<T>
}

val stringBoxFactory = Factory { type, annotations, moshi ->
  if (Types.equals(type, Types.newParameterizedTypeWithOwner(Any::class.java, Box::class.java, String::class.java))) {
    StringBoxAdapter()
  } else null
}
```
In the above, the equality check will silently fail currently (from the wrong owner type), and you'll never get your StringBoxAdapter. With the change, the broken `newParameterizedTypeWithOwner` usage will fail.

`@nedtwigg` was asking about why `newParameterizedTypeWithOwner` was "working" with the wrong owner type over on the auto-value-moshi issue, so I posted this.

Does somebody ever enter a wrong owner type? Probably very rarely, yeah.
Happy to close this if it's never going to happen in reality.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -503,10 +503,17 @@ static void checkNotPrimitive(Type type) {
 
     ParameterizedTypeImpl(@Nullable Type ownerType, Type rawType, Type... typeArguments) {
       // Require an owner type if the raw type needs it.
-      if (rawType instanceof Class<?>
-          && (ownerType == null) != (((Class<?>) rawType).getEnclosingClass() == null)) {
-        throw new IllegalArgumentException(
-            "unexpected owner type for " + rawType + ": " + ownerType);
```
NightlyNexus: fwiw, change with less refactoring could be:
```java
if (rawType instanceof Class<?>
    && ((ownerType == null) != (((Class<?>) rawType).getEnclosingClass() == null)
    || (ownerType != null && ((Class<?>) rawType).getEnclosingClass() != ownerType))) {
  throw new IllegalArgumentException(
      "unexpected owner type for " + rawType + ": " + ownerType);
}
```

swankjesse:
LGTM!

PR #460, opened 2018-03-10
Disallow null annotation set in adapter lookup.

NightlyNexus:
This just happened to me because I wasn't null checking the result of `Types.nextAnnotations`. might as well fail here instead of in Moshi.cacheKey.


swankjesse:
Nice.

PR #462, opened 2018-03-12
Split the MoshiKotlinCodeGenProcessor into multiple types.

swankjesse:
Move more behavior into the types: Adapter and Property now have many
instance methods and are now mutable value classes rather than immutable
data classes.

This changes how some of the bookkeeping code works. Previously there were
some pretty tricky types: maps with pairs as keys, indexed and non-indexed
lists of properties. With this change more of the logic operates directly
on the properties.

### DIFF
--- a/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
+++ b/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
@@ -1,0 +1,276 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi
+
+import com.squareup.kotlinpoet.ARRAY
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FileSpec
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.KModifier
+import com.squareup.kotlinpoet.NameAllocator
+import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.ParameterizedTypeName
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.TypeVariableName
+import com.squareup.kotlinpoet.asClassName
+import com.squareup.kotlinpoet.asTypeName
+import org.jetbrains.kotlin.serialization.ProtoBuf
+import java.lang.reflect.Type
+import javax.lang.model.element.Element
+import javax.lang.model.util.Elements
+
+/** Generates a JSON adapter for a target type. */
+internal class AdapterGenerator(
+  val fqClassName: String,
+  val packageName: String,
+  val propertyList: List<PropertyGenerator>,
+  val originalElement: Element,
+  name: String = fqClassName.substringAfter(packageName)
+      .replace('.', '_')
+      .removePrefix("_"),
+  val hasCompanionObject: Boolean,
+  val visibility: ProtoBuf.Visibility,
+  val elements: Elements,
+  val genericTypeNames: List<TypeVariableName>?
+) {
+  val nameAllocator = NameAllocator()
+  val adapterName = "${name}JsonAdapter"
+  val originalTypeName = originalElement.asType().asTypeName()
+
+  val moshiParam = ParameterSpec.builder(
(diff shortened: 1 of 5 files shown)


hzsweers:
:+1:

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
```diff
@@ -0,0 +1,276 @@
...
+import com.squareup.kotlinpoet.TypeVariableName
+import com.squareup.kotlinpoet.asClassName
+import com.squareup.kotlinpoet.asTypeName
+import org.jetbrains.kotlin.serialization.ProtoBuf
+import java.lang.reflect.Type
+import javax.lang.model.element.Element
+import javax.lang.model.util.Elements
+
+/** Generates a JSON adapter for a target type. */
+internal class Adapter(
```
JakeWharton: `AdapterGenerator`? `AdapterWriter`? This isn't _just_ an `Adapter`.

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
```diff
@@ -0,0 +1,276 @@
...
+      .build()
+  val readerParam = ParameterSpec.builder(
+      nameAllocator.newName("reader"),
+      JsonReader::class)
+      .build()
+  val writerParam = ParameterSpec.builder(
+      nameAllocator.newName("writer"),
+      JsonWriter::class)
+      .build()
+  val valueParam = ParameterSpec.Companion.builder(
```
JakeWharton: nit: drop `Companion`

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
```diff
@@ -0,0 +1,276 @@
...
+      nameAllocator.newName("value"),
+      originalTypeName.asNullable())
+      .build()
+  val jsonAdapterTypeName = ParameterizedTypeName.Companion.get(
```
JakeWharton: nit: drop `Companion`

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
```diff
@@ -0,0 +1,276 @@
...
+      JsonAdapter::class.asClassName(), originalTypeName)
+
+  // selectName() API setup
+  val optionsProperty = PropertySpec.Companion.builder(
```
JakeWharton: nit: drop `Companion`

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Property.kt
```diff
@@ -0,0 +1,146 @@
...
+import com.squareup.kotlinpoet.ParameterizedTypeName
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.TypeVariableName
+import com.squareup.kotlinpoet.WildcardTypeName
+import com.squareup.kotlinpoet.asTypeName
+import javax.lang.model.element.AnnotationMirror
+
+/** Generates functions to encode and decode a property as JSON. */
+internal class Property(
```
JakeWharton: Same suffix comment

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Property.kt
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Returns a suggested variable name derived from a type name.
+ */
+private fun TypeName.toVariableName(): String {
+  return when (this) {
+    is ClassName -> simpleName().decapitalize()
+    is ParameterizedTypeName -> {
+      rawType.simpleName().decapitalize() + if (typeArguments.isEmpty()) "" else "__" + typeArguments.toVariableNames()
```
JakeWharton: This underscore thing is super weird. The whole point of name allocator is to avoid tricks like this, no? Can fix in follow up.

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
swankjesse: Fixed!

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
swankjesse: Fixed!

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
swankjesse: Fixed!

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Adapter.kt
swankjesse: Fixed!

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Property.kt
swankjesse: Fixed. PropertyGenerator.

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/Property.kt
swankjesse: Will punt to follow-up. This PR doesn’t change behavior.

PR #464, opened 2018-03-25
Fold the kotlin-codegen-runtime into Moshi itself.

swankjesse:
Rename @MoshiSerializable to @JsonClass. Like @Json, I'm anticipating
a future where there are other interesting properties on this annotation.
Perhaps a future feature where Moshi is strict and only adapts types that
have a '@JsonClass' annotation.

Also rename MoshiKotlinCodeGenProcessor to JsonClassCodeGenProcessor. We
may later support other ways of generating code here; perhaps for regular
Java types.


moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -215,6 +223,31 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
+      if (type instanceof ParameterizedType) {
+        return adapterClass.getDeclaredConstructor(Moshi.class, Type[].class)
+            .newInstance(moshi, ((ParameterizedType) type).getActualTypeArguments());
+      } else {
+        return adapterClass.getDeclaredConstructor(Moshi.class).newInstance(moshi);
+      }
+    } catch (ClassNotFoundException | NoSuchMethodException e) {
+      throw new RuntimeException(
+          "Failed to find the generated the JsonAdapter for " + rawType, e);
+    } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
```
JakeWharton: Doesn't this widen to ReflectiveOperationException which is only available on API 19+ / Java 7+?

swankjesse: The compiler can’t do that automatically because it doesn’t want to catch other, unknown subtypes of ReflectiveOperationException.

JakeWharton: That contrasts with AS warnings and animal sniffer failures that I've received in the past.

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -53,6 +55,12 @@ private StandardJsonAdapters() {
       if (type == Object.class) return new ObjectJsonAdapter(moshi).nullSafe();
 
       Class<?> rawType = Types.getRawType(type);
+
+      JsonClass jsonClass = rawType.getAnnotation(JsonClass.class);
+      if (jsonClass != null && jsonClass.generateAdapter()) {
+        return generatedAdapter(moshi, type, rawType);
```
NightlyNexus: this is so cool. was skeptical, but i like it now.

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -215,6 +223,31 @@ static int rangeCheckNextInt(JsonReader reader, String typeMessage, int min, int
...
+   * uses the same naming conventions as {@code JsonClassCodeGenProcessor}.
+   */
+  static JsonAdapter<?> generatedAdapter(Moshi moshi, Type type, Class<?> rawType) {
+    String adapterClassName = rawType.getName().replace("$", "_") + "JsonAdapter";
+    try {
+      @SuppressWarnings("unchecked") // We generate types to match.
+      Class<? extends JsonAdapter<?>> adapterClass = (Class<? extends JsonAdapter<?>>)
+          Class.forName(adapterClassName, true, rawType.getClassLoader());
+      if (type instanceof ParameterizedType) {
+        return adapterClass.getDeclaredConstructor(Moshi.class, Type[].class)
```
hzsweers: Want to fold in the defensive accessibility check from #465 while you're at it?

swankjesse: Can do

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
swankjesse: Confirmed you are correct. Gross! Fixed to have 5 catch clauses. Gross!

Issue #461, opened 2018-03-12
Kotlin Codegen Release Blockers


swankjesse:
The code’s checked in but there’s some follow-ups to do before we can invite users to use it:

 - [x] kotlin-metadata dependency released in mavencentral (https://github.com/Takhion/kotlin-metadata/issues/5)
 - [x] constructor signature
 - [x] @JsonQualifier annotations
 - [x] @MoshiSerializable annotation and factory potentially in core
 - [x] ”Generated code; do not edit“ header

ChrisAckerman:
Add jcenter to your pom, and there’s no need for kotlin-metadata to be in maven central.

```
<repositories>
    <repository>
      <id>jcenter</id>
      <url>https://jcenter.bintray.com/</url>
    </repository>
</repositories>
```

ChrisAckerman:
https://bintray.com/takhion/maven/kotlin-metadata

JakeWharton:
Sonatype OSSRH prevents release of artifacts to Maven central which point to arbitrary repositories.

hzsweers:
We should probably generated `@Generated` annotations onto the adapters via https://github.com/google/auto/blob/master/common/src/main/java/com/google/auto/common/GeneratedAnnotationSpecs.java#L46-L61

This could possibly just replace the "generated code, do not edit" header bit in the list. 

This would also allow for more fine-grained proguard `keepnames` retentions, which otherwise would have to keep all of moshi's built-in `JsonAdapter`-suffixed classes as well due to `**JsonAdapter`

```proguard
-keepnames @com.squareup.moshi.<wherever this ends up>.MoshiSerializable class *

# Java < 9
-keepnames @javax.annotation.Generated class **JsonAdapter

#  Java 9+
-keepnames @javax.annotation.processing.Generated class **JsonAdapter
```

hzsweers:
Another one for the list: proguard will make seemingly unused constructors `private`, so we should defensively check in `MoshiSerializableFactory` if the `Constructor` is accessible and make it so if it's not

swankjesse:
I’m looking into this. We need something like compile-testing for Kotlin. Otherwise we can’t test that unsupported inputs like abstract classes are handled appropriately. Right now we’re only testing inputs that can successfully generate an adapter.

hzsweers:
@takhion has actually mentioned working on something like that recently,
not sure where it stands.




On Mar 28, 2018 at 7:39 PM, Jesse Wilson <notifications@github.com> wrote:


[...]

PR #465, opened 2018-03-27
Defensively check if generated constructor is accessible and fix if not

hzsweers:
Part of #461

Proguard will make unused constructors private if possible, so this defensively makes them accessible if they're not already.


hzsweers:
Let me know if/how you want to test this

hzsweers:
Whoops, just saw #464. Will propose adding this into that

PR #466, opened 2018-03-27
Generate `@Generated` annotation onto adapters when possible

hzsweers:
Part of #461

This leverages AutoCommon's `GeneratedAnnotations#generatedAnnotation` API to generate a `@Generated` annotation where possible. This keeps with conventions in other code gen tools, and also allows for more fine grained proguard rules for keeping generated adapter names, like so:

```proguard
-keepnames @com.squareup.moshi.<wherever this ends up>.MoshiSerializable class *

# Java < 9
-keepnames @javax.annotation.Generated class **JsonAdapter

#  Java 9+
-keepnames @javax.annotation.processing.Generated class **JsonAdapter
```

Generated annotation looks like this:

```kotlin
@Generated(
        value = ["com.squareup.moshi.MoshiKotlinCodeGenProcessor"],
        comments = "https://github.com/square/moshi"
)
```

I think this should be able to replace the `”Generated code; do not edit“ header` item in the linked issue as well.

This doooooes also replace `elements` in `AdapterGenerator` with a `ProcessingEnv`, but I figured that was less evil than polluting it with both `elements` and plumbing down an `env` separately simultaneously. This does also hit a weird ambiguity case due to `KotlinMetadataUtils`' repeat declaration, so a good reason for removing that in the future. Figured it best to punt on a better final place for this to another time.

### DIFF
--- a/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
+++ b/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
@@ -104,8 +104,8 @@ internal class AdapterGenerator(
 
     generatedAnnotation(env.elementUtils, env.sourceVersion).ifPresent {
       result.addAnnotation(AnnotationSpec.builder(it.asClassName())
-          .addMember("value = [%S]", MoshiKotlinCodeGenProcessor::class.java.canonicalName)
-          .addMember("comments = %S", "https://github.com/square/moshi")
+          .addMember("%S", MoshiKotlinCodeGenProcessor::class.java.canonicalName)
+          .addMember("%S", "https://github.com/square/moshi")
           .build())
     }
 
--- a/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
+++ b/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
@@ -15,7 +15,9 @@
  */
 package com.squareup.moshi
 
+import com.google.auto.common.GeneratedAnnotations.generatedAnnotation
 import com.squareup.kotlinpoet.ARRAY
+import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
@@ -30,8 +32,8 @@ import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
 import org.jetbrains.kotlin.serialization.ProtoBuf
 import java.lang.reflect.Type
+import javax.annotation.processing.ProcessingEnvironment
 import javax.lang.model.element.Element
-import javax.lang.model.util.Elements
 
 /** Generates a JSON adapter for a target type. */
 internal class AdapterGenerator(
@@ -41,7 +43,7 @@ internal class AdapterGenerator(
   val isDataClass: Boolean,
   val hasCompanionObject: Boolean,
   val visibility: ProtoBuf.Visibility,
-  val elements: Elements,
+  val env: ProcessingEnvironment,
   val genericTypeNames: List<TypeVariableName>?
 ) {
   val nameAllocator = NameAllocator()
@@ -99,6 +101,14 @@ internal class AdapterGenerator(
 
   private fun generateType(): TypeSpec {
     val result = TypeSpec.classBuilder(adapterName)
+
+    generatedAnnotation(env.elementUtils, env.sourceVersion).ifPresent {
+      result.addAnnotation(AnnotationSpec.builder(it.asClassName())
+          .addMember("value = [%S]", MoshiKotlinCodeGenProcessor::class.java.canonicalName)
+          .addMember("comments = %S", "https://github.com/square/moshi")
+          .build())
+    }
+
     result.superclass(jsonAdapterTypeName)
 
     genericTypeNames?.let {
--- a/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
+++ b/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
@@ -15,7 +15,6 @@
  */
 package com.squareup.moshi
 
-import com.google.auto.common.GeneratedAnnotations.generatedAnnotation
 import com.squareup.kotlinpoet.ARRAY
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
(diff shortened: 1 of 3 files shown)


kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
```diff
@@ -102,6 +104,14 @@ internal class AdapterGenerator(
 
   private fun generateType(): TypeSpec {
     val result = TypeSpec.classBuilder(adapterName)
+
+    generatedAnnotation(env.elementUtils, env.sourceVersion).ifPresent {
+      result.addAnnotation(AnnotationSpec.builder(it.asClassName())
+          .addMember("value = [%S]", MoshiKotlinCodeGenProcessor::class.java.canonicalName)
```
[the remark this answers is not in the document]
hzsweers: True! Missing brackets are only deprecated in named form, I was confused by that I think

d73e0a6

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/MoshiKotlinCodeGenProcessor.kt
```diff
@@ -173,7 +173,7 @@ class MoshiKotlinCodeGenProcessor : KotlinAbstractProcessor(), KotlinMetadataUti
         hasCompanionObject = hasCompanionObject,
         visibility = classProto.visibility!!,
         genericTypeNames = genericTypeNames,
-        elements = elementUtils)
+        env = (this as KotlinAbstractProcessor).processingEnv) // Gross workaround for property ambiguity
```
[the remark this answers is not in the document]
hzsweers: comment - 81d48f3

True. I thought about that actually, but piping down both a source version separately alongside elements seemed worse

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
```diff
@@ -102,6 +104,14 @@ internal class AdapterGenerator(
 
   private fun generateType(): TypeSpec {
     val result = TypeSpec.classBuilder(adapterName)
+
+    generatedAnnotation(env.elementUtils, env.sourceVersion).ifPresent {
```
swankjesse: the fact that this emits different things depending on what the env’s source version is, is disgusting. Can we just hardcode the `javax.annotation.Generated` one?

JakeWharton: That will prevent its use on Java 9 or newer.

JakeWharton: Of course, Moshi uses JSR 305 already so it's already prevented from being used on the module path so... that's actually not a terrible idea until we tear out JSR 305.

JakeWharton: Another advantage of using this is that it only emits if the annotation is available on the classpath which javax.annotation.Generated is not for Android projects.

swankjesse: Can’t we use this?
 https://docs.oracle.com/javase/9/docs/api/javax/annotation/Generated.html

JakeWharton: You can, but it's in a super weird module that must be explicitly depended on and the entire module is marked as `@Deprecated(forRemoval=true)` so eventually this will need changed. And then you still need to conditionally generate for Android.

JakeWharton: Some details in https://github.com/google/auto/issues/560

swankjesse: I dislike the premise that you get a different annotation based on what JDK you’re compiling with.

Can we just skip `@Generated`? Seems like the juice isn’t worth the squeeze.

JakeWharton: That sounds simple!

hzsweers: That is simpler, but this is also useful for static analysis tools to know it's generated code. I agree it's a bit awkward due to the change in java 9, but it's also inline with what Dagger, AutoValue, etc all do. I'd definitely want to have some way to at least optionally enable it, even if off by default. What do you think?

swankjesse: If we support anything it should be opt-in via one of the annotation processor args: `-Amoshi.generated=javax.annotation.Generated`

If we add this we should discourage its use because it prevents a single class from being used in multiple environments. Everyone's life is easier if their generated code works on all platforms: Android, Java 8, Java 9.

hzsweers: Done in 436f36f

Also reverted the elements -> processingEnv change in b454aaa

For the life of me I can't figure out how to specify apt options in maven kapt though :|

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/MoshiKotlinCodeGenProcessor.kt
```diff
@@ -53,9 +56,38 @@ import javax.tools.Diagnostic.Kind.ERROR
...
   private val annotationName = MoshiSerializable::class.java.canonicalName
+  private var generatedType: TypeElement? = null
+
+  override fun init(processingEnv: ProcessingEnvironment) {
+    super.init(processingEnv)
+    generatedType = processingEnv.options[OPTION_GENERATED]?.let {
+      if (it !in POSSIBLE_GENERATED_NAMES) {
+        throw IllegalArgumentException("Invalid option value for $OPTION_GENERATED. Found $it, allowable values are $POSSIBLE_GENERATED_NAMES.")
+      }
+      processingEnv.elementUtils.getTypeElement(it)
```
NightlyNexus: if this returns null, though, isn't that an illegal state? if you're specifying the annotation to use, and it's not available, would it be better to fail here?
maybe generatedType should just be lateinit?

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AdapterGenerator.kt
```diff
@@ -96,12 +98,20 @@ internal class AdapterGenerator(
...
+    result.addType(generateType(generatedOption))
     return result.build()
   }
 
-  private fun generateType(): TypeSpec {
+  private fun generateType(generatedOption: TypeElement?): TypeSpec {
     val result = TypeSpec.classBuilder(adapterName)
+
+    generatedOption?.let {
+      result.addAnnotation(AnnotationSpec.builder(it.asClassName())
```
NightlyNexus: nit: I find a simple null check easier to read.
```kotlin
if (generatedOption != null) {
  result.addAnnotation(AnnotationSpec.builder(generatedOption.asClassName())
  ...
}
```
but i dunno.

hzsweers: idk, I like the `let` ¯\_(ツ)_/¯

PR #471, opened 2018-03-31
Call the kotlin compiler from within a test case.

swankjesse:
This is a fragile first step.


swankjesse:
FYI @eugeniomarletti

swankjesse:
FYI @Takhion

hzsweers:
This is super neat

PR #473, opened 2018-04-03
Handle qualifiers, names, and transient in generated adapters

swankjesse:


kotlin-codegen/compiler/src/main/java/com/squareup/moshi/JsonClassCodeGenProcessor.kt
```diff
@@ -127,6 +130,17 @@ class JsonClassCodeGenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
       nameResolver.getString(it.name)
     }
 
+    // The compiler might emit methods just so it has a place to put annotations. Find these.
+    val annotatedElements = mutableMapOf<Property, ExecutableElement>()
+    for (enclosedElement in element.enclosedElements) {
+      if (enclosedElement is ExecutableElement) {
+        val property = classData.getPropertyOrNull(enclosedElement)
+        if (property != null) {
```
hzsweers: You could write this as

```kotlin
val annotatedElements = element.enclosedElements
    .filterIsInstance<ExecutableElement>()
    .mapNotNull { element -> 
      classData.getPropertyOrNull(element)?.let {  it to element } 
    }
    .associate { it }
```

hzsweers: start with asSequence to make it single iteration too

swankjesse: The implicit Pair in that is awkward. Procedural code is easy to step through in the debugger!

swankjesse: How’s this:
```
    // The compiler might emit methods just so it has a place to put annotations. Find these.
    val annotatedElements = mutableMapOf<Property, ExecutableElement>()
    for (enclosedElement in element.enclosedElements) {
      if (enclosedElement !is ExecutableElement) continue
      val property = classData.getPropertyOrNull(enclosedElement) ?: continue
      annotatedElements[property] = enclosedElement
    }
```

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/JsonClassCodeGenProcessor.kt
```diff
@@ -142,22 +153,33 @@ class JsonClassCodeGenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
...
       if (property.visibility != ProtoBuf.Visibility.INTERNAL
           && property.visibility != ProtoBuf.Visibility.PROTECTED
           && property.visibility != ProtoBuf.Visibility.PUBLIC) {
         messager.printMessage(ERROR, "property $name is not visible", enclosedElement)
         return null
       }
 
+      val hasDefault = parameter?.declaresDefaultValue ?: true
+
+      if (enclosedElement.modifiers.contains(Modifier.TRANSIENT)) {
```
JakeWharton: nit: could be `TRANSIENT in enclosedElement.modifiers`

PR #477, opened 2018-04-04
Hide Types.resolve.

NightlyNexus:
unfortunately requires moving most of the internal type utilities into the Util. couldn't think of anything clever to make the diff not a mess.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Types.java
+++ b/moshi/src/main/java/com/squareup/moshi/Types.java
@@ -15,10 +15,12 @@
  */
 package com.squareup.moshi;
 
+import com.squareup.moshi.internal.Util.GenericArrayTypeImpl;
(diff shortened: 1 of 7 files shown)


moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -49,7 +53,7 @@ private Types() {
    * Returns the subset of {@code annotations} without {@code jsonQualifier}, or null if {@code
    * annotations} does not contain {@code jsonQualifier}.
    */
-  @CheckReturnValue public static @Nullable Set<? extends Annotation> nextAnnotations(
+  public static @Nullable Set<? extends Annotation> nextAnnotations(
```
swankjesse: Why drop @CheckReturnValue?

NightlyNexus: Just moved it to the Types type. That applies it to all the methods.

PR #480, opened 2018-04-05
Add API for defaulting enum values.

NightlyNexus:
I had this ancient branch that I never submitted ha.

This API would be nice, as I've had to copy the slightly modified EnumJsonAdapter into a project before, but I'm not attached!

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -15,6 +15,7 @@
  */
 package com.squareup.moshi;
 
+import com.squareup.moshi.StandardJsonAdapters.EnumJsonAdapter;
 import com.squareup.moshi.internal.Util;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -227,6 +228,26 @@ public Builder add(Object adapter) {
...
 
+    /**
+     * @param fallback the default value to use when a string is deserialized
+     * and does not match an existing enum value.
+     */
+    public <T extends Enum<T>> Builder addEnumFallback(final T fallback) {
+      if (fallback == null) {
+        throw new NullPointerException("fallback == null");
+      }
+      final Class<T> enumType = fallback.getDeclaringClass();
```
NightlyNexus: or could accept the class as a parameter and allow null values. not sure. i tend to use non-null enum types and have another enum value for the sentinel value of unknown.

swankjesse:
I like the behavior but I don’t know if this is the way I‘d have us expose it. I think I’d prefer to expose this as a JsonAdapter in `moshi-adapters` and have the user install it.

```
JsonAdapter<Roshambo> roshamboAdapter = CustomEnumAdapter.create(Roshambo.class)
    .withFallback(Roshambo.PAPER);

new Moshi.Builder()
    .add(Roshambo.class, roshamboAdapter)
    .build()
```

This is more code, but it’s also a little bit more flexible. It works with qualifier annotations. It gives us the ability to customize things later.

```
JsonAdapter<Roshambo> roshamboAdapter = CustomEnumAdapter.create(Roshambo.class)
    .withJsonName(Roshambo.SCISSORS, "snip")
    .withFallback(Roshambo.PAPER);
```

NightlyNexus:
Ooh, neat. 👍

PR #481, opened 2018-04-05
Add example for custom qualifier with an element.

NightlyNexus:


swankjesse:
This is brilliant.

NightlyNexus:
Credit to @serj-lotutovici for the idea.

PR #482, opened 2018-04-05
JsonAdapter.nonNull() forbids explicit nulls in the JSON body

swankjesse:
This adapter modifier throws exceptions if an unexpected null is
encountered. This may pair nicely with Kotlin.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -153,6 +153,39 @@ public abstract class JsonAdapter<T> {
     };
   }
 
+  /**
+   * Returns a JSON adapter equal to this JSON adapter, but that refuses null values. If null is
+   * read or written this will throw a {@link JsonDataException}.
+   *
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -153,6 +153,39 @@ public final void toJson(BufferedSink sink, @Nullable T value) throws IOExceptio
     };
   }
 
+  /**
+   * Returns a JSON adapter equal to this JSON adapter, but that refuses null values. If null is
+   * read or written this will throw a {@link JsonDataException}.
+   *
+   * <p>Note that this adapter will not usually be invoked for absent values and so those must be
+   * handled elsewhere. This should only be used to fail on explicit nulls.
```
NightlyNexus: nice.
the thing that made me unsure before was whether or not the result of `delegate.fromJson(reader)` should be checked for null.

this sounds like the right call.

NightlyNexus:
yeah, and then the KotlinJsonAdapter can do `if (!property.returnType.isMarkedNullable) adapter = adapter.nonNull()` to get a better error message.

PR #499, opened 2018-04-11
Add apoptions support in KotlinCompilerCall

hzsweers:
This also adds a smoke test to check for generated type errors.

### DIFF
--- a/kotlin-codegen/compiler/src/test/java/com/squareup/moshi/KotlinCompilerCall.kt
+++ b/kotlin-codegen/compiler/src/test/java/com/squareup/moshi/KotlinCompilerCall.kt
@@ -20,11 +20,14 @@ import okio.Buffer
 import okio.Okio
 import org.jetbrains.kotlin.cli.common.CLITool
 import org.jetbrains.kotlin.cli.jvm.K2JVMCompiler
+import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.FileOutputStream
+import java.io.ObjectOutputStream
 import java.io.PrintStream
 import java.net.URLClassLoader
 import java.net.URLDecoder
+import java.util.Base64
 import java.util.zip.ZipEntry
 import java.util.zip.ZipOutputStream
 import kotlin.reflect.KClass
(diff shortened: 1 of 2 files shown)


hzsweers:
Note that I found something interesting while testing. All tests currently will have failed exit codes due to

```
ERROR: Write-unsafe context! Model changes are allowed from write-safe contexts only. Please ensure you're using invokeLater/invokeAndWait with a correct modality state (not "any"). See TransactionGuard documentation for details.
  current modality=ModalityState.NON_MODAL
  known modalities={ModalityState.NON_MODAL=true}
```

swankjesse:
Yeah, I’ve seen that error but it’s unclear what to do about it. We’re going to eventually need tests that successfully compile, so I’m going to have to figure that out. Unless somebody beats me to it!

kotlin-codegen/compiler/src/test/java/com/squareup/moshi/KotlinCompilerCall.kt
```diff
@@ -166,4 +176,20 @@ class KotlinCompilerCall(var scratchDir: File) {
     throw IllegalStateException("no kotlin-annotation-processing-embeddable jar on classpath:\n  " +
         "${classpathFiles().joinToString(separator = "\n  ")}}")
   }
-}
\ No newline at end of file
+
+  /**
+   * Base64 encodes a mapping of annotation processor args for kapt, borrowed from
+   * https://kotlinlang.org/docs/reference/kapt.html#apjavac-options-encoding
```
NightlyNexus: for fun, i looked at ObjectOutputStream and tried to make this work with Okio.
```kotlin
val body = Buffer()
body.writeInt(options.size)
for ((key, value) in options.entries) {
  body.writeShort(Utf8.size(key).toInt())
  body.writeUtf8(key)
  body.writeShort(Utf8.size(value).toInt())
  body.writeUtf8(value)
}

val result = Buffer()
result.writeShort(STREAM_MAGIC.toInt())
result.writeShort(STREAM_VERSION.toInt())
val bodySize = body.size()
when {
  bodySize <= 0xFF -> {
    result.writeByte(TC_BLOCKDATA.toInt())
    result.writeByte(body.size().toInt())
  }
  bodySize <= 0x7FFFFFFF -> {
    result.writeByte(TC_BLOCKDATALONG.toInt())
    result.writeInt(bodySize.toInt())
  }
  else -> TODO("More blocks.")
}
result.writeAll(body)

val readByteString = result.readByteString()
return readByteString.base64()
```

so much hidden knowledge in ObjectOutputStream!

PR #500, opened 2018-04-14
Add coherent error message for uneconded map keys.

### LINKED ISSUE #277: Error where map key doesn’t encode as a string is incoherent
When a map key doesn’t have a natural string encoding we crash with a difficult exception. [For example](https://bitbucket.org/robeden/moshi-test), when writing a `Map<Matcher,String> identifiers` we crash like so:

```
Exception in thread "main" java.lang.IllegalStateException: Nesting problem.
        at com.squareup.moshi.JsonUtf8Writer.beforeValue(JsonUtf8Writer.java:367)
        at com.squareup.moshi.JsonUtf8Writer.open(JsonUtf8Writer.java:102)
        at com.squareup.moshi.JsonUtf8Writer.beginObject(JsonUtf8Writer.java:89)
        at com.squareup.moshi.ClassJsonAdapter.toJson(ClassJsonAdapter.java:177)
        at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:134)
        at com.squareup.moshi.MapJsonAdapter.toJson(MapJsonAdapter.java:56)
```

Instead we should say “Matcher cannot be used as a map key in JSON.”

The fix for the above example is a JSON adapter, like so:

```
class MatcherJsonAdapter {
  @ToJson String matcherToJson(Matcher m) {
    return m.type + ":" + m.match;
  }

  @FromJson Matcher matcherFromJson(String s) {
    String[] parts = s.split(":", 1);
    return new Matcher(MatchType.valueOf(parts[0]), parts[1]);
  }
}
```

NightlyNexus:
Closes #277.

We could pass in the class name from the map adapter, but this error message is clear enough without adding more complexity. (Correct me if I'm wrong, of course.)

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
@@ -77,6 +77,10 @@ final class JsonUtf8Writer extends JsonWriter {
   }
 
   @Override public JsonWriter beginArray() throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "Array cannot be used as a map key in JSON at path " + getPath());
+    }
     writeDeferredName();
     return open(EMPTY_ARRAY, "[");
   }
@@ -86,6 +90,10 @@ final class JsonUtf8Writer extends JsonWriter {
   }
 
   @Override public JsonWriter beginObject() throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "Object cannot be used as a map key in JSON at path " + getPath());
+    }
     writeDeferredName();
     return open(EMPTY_OBJECT, "{");
   }
@@ -171,6 +179,10 @@ final class JsonUtf8Writer extends JsonWriter {
   }
 
   @Override public JsonWriter nullValue() throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "null cannot be used as a map key in JSON at path " + getPath());
+    }
     if (deferredName != null) {
       if (serializeNulls) {
         writeDeferredName();
@@ -186,6 +198,10 @@ final class JsonUtf8Writer extends JsonWriter {
   }
 
   @Override public JsonWriter value(boolean value) throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "Boolean cannot be used as a map key in JSON at path " + getPath());
+    }
     writeDeferredName();
     beforeValue();
     sink.writeUtf8(value ? "true" : "false");
--- a/moshi/src/test/java/com/squareup/moshi/MapJsonAdapterTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/MapJsonAdapterTest.java
@@ -141,6 +141,65 @@ public final class MapJsonAdapterTest {
     assertThat(jsonAdapter.fromJsonValue(jsonObject)).isEqualTo(map);
   }
 
+  @Test public void booleanKeyTypeHasCoherentErrorMessage() {
+    Map<Boolean, String> map = new LinkedHashMap<>();
+    map.put(true, "");
+    JsonAdapter<Map<Boolean, String>> adapter = mapAdapter(Boolean.class, String.class);
+    try {
+      adapter.toJson(map);
+      fail();
+    } catch (IllegalStateException expected) {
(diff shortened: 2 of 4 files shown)


moshi/src/test/java/com/squareup/moshi/MapJsonAdapterTest.java
```diff
@@ -141,6 +142,84 @@
...
 
+  @Test public void booleanKeyTypeHasCoherentErrorMessage() {
+    Map<Boolean, String> map = new LinkedHashMap<>();
+    map.put(true, "");
+    JsonAdapter<Map<Boolean, String>> adapter = mapAdapter(Boolean.class, String.class);
+    try {
+      adapter.toJson(map);
+      fail();
+    } catch (IllegalStateException expected) {
+      assertThat(expected).hasMessage("Boolean cannot be used as a map key in JSON.");
```
NightlyNexus: tiny thing: can't recall if we prefer `expected` or `e` when matching error messages instead of ignoring the error.

moshi/src/test/java/com/squareup/moshi/PromoteNameToValueTest.java
```diff
@@ -252,7 +252,7 @@
       writer.value(true);
       fail();
     } catch (IllegalStateException e) {
-      assertThat(e).hasMessage("Nesting problem.");
+      assertThat(e).hasMessage("Boolean cannot be used as a map key in JSON.");
```
Egorand: This catch block uses `e` instead of `expected`, but I'd actually prefer the latter, `expected` vs `ignored` seems like a good naming convention. I think it's a nit though

moshi/src/test/java/com/squareup/moshi/PromoteNameToValueTest.java
```diff
@@ -278,7 +279,7 @@
   @Test public void writerNullValue() throws Exception {
```
NightlyNexus: this can't actually happen in practice because the map adapter doesn't allow null keys, but i'll update the error message, too. be right back.

NightlyNexus: did it.

swankjesse:
I think this is actually harder to follow than your first commit. Sorry for the grind.

Perhaps the best approach will be fore `promoteNameToValue()` to take a `JsonReader.Token` that hints about what it was doing when it failed?

NightlyNexus:
Don't apologize! Thanks for reviewing!

Should the JsonReader.Token be a field in addition to the existing boolean? Hacking in a nullable field to represent the boolean is confusing.

swankjesse:
Is it feasible to have just the boolean field, and pass the token as a parameter?

NightlyNexus:
For JsonWriter? `JsonWriter.promoteValueToName(JsonReader.Token)`?

JsonWriter needs to cache the possible error message because JsonWriter doesn't throw in promoteValueToName; it throws if the next value (JsonWriter.value(...)) is not a number or string.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -91,7 +91,7 @@
   }
 
   @Override public JsonWriter endObject() throws IOException {
-    promoteValueToName = false;
+    promoteValueToNameFailureToken = null;
```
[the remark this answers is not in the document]
NightlyNexus: that's what i'm unsure about.

the options i see:
1. Leave the code unchanged (only use the boolean field, revert my last commit on this PR). The error message will be generic and won't have the JSON token.
2. Change the field to a nullable JsonReader.Token, where null is the existing `false`, and the non-null value can be used in the error message.

I'm leaning toward 1 because of the diminished complexity.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -77,6 +77,10 @@
   }
 
   @Override public JsonWriter beginArray() throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "Array cannot be used as a map key in JSON at path " + getPath());
+    }
     writeDeferredName();
```
NightlyNexus: could be `writeDeferredName("Array")` and writeDeferredName do the check, but then it felt weird to pass in a never used parameter from the callers in value(String) and value(Number).

Issue #65, opened 2015-07-20
Allow raw JSON values?


orip:
I have raw JSON on disk that I want to embed in side a wrapper JSON. It would be awesome to do something like (explicitly choosing horrible names and hook points before and after)

```
File fileWithJson = ...;
JsonWriter json = new JsonWriter(sink);
json.beginObject();
json.name("foo").aboutToAddExternalValue();
sink.writeAll(Okio.source(fileWithJson));
json.valueAddedExternallyTrustMeItsFine();
json.closeObject();
```

Or even:

```
json.name("foo").value(Okio.source(...))
```

JakeWharton:
`.aboutToAddExternalValue()` would just be `.writeDeferredName()` (private) and `.flush()`. A second method would not be needed.

I question the value something like would have in the public API.

orip:
Fair enough regarding exposing it in the normal API.
It's a good technique to construct JSON API requests while streaming from disk, I wonder whether there's a way to allow reusing JsonWriter without forking that is inconvenient enough that people don't shoot themselves in the foot.

JakeWharton:
We can probably come up with something. I definitely like the stream-from-disk intention vs a raw byte[] or String.

adamtanner:
I just did this in https://github.com/google/gson/pull/667 if you want to port it over.

mescortes:
CC @mescortes

tmm1:
@NightlyNexus thanks for implementing this!

Is there an equivalent for `JsonReader`? i.e if I want to consume the contents of a specific field as a json blob with no additional parsing?

JakeWharton:
That's #318

PR #503, opened 2018-04-14
Allow writing out raw JSON.

NightlyNexus:
taking a shot at #65.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
@@ -18,6 +18,7 @@ package com.squareup.moshi;
 import java.io.IOException;
 import javax.annotation.Nullable;
 import okio.BufferedSink;
+import okio.BufferedSource;
 import okio.Sink;
 
 import static com.squareup.moshi.JsonScope.DANGLING_NAME;
@@ -257,6 +258,18 @@ final class JsonUtf8Writer extends JsonWriter {
     writeDeferredName();
     beforeValue();
     sink.writeUtf8(string);
+    pathIndices[stackSize - 1]++;
+    return this;
+  }
+
+  @Override public JsonWriter value(BufferedSource source) throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "BufferedSource cannot be used as a map key in JSON at path " + getPath());
+    }
+    writeDeferredName();
+    beforeValue();
+    sink.writeAll(source);
     pathIndices[stackSize - 1]++;
     return this;
   }
--- a/moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
+import okio.BufferedSource;
 
 import static com.squareup.moshi.JsonScope.EMPTY_ARRAY;
 import static com.squareup.moshi.JsonScope.EMPTY_DOCUMENT;
@@ -201,6 +202,23 @@ final class JsonValueWriter extends JsonWriter {
       return name(bigDecimalValue.toString());
     }
     add(bigDecimalValue);
+    pathIndices[stackSize - 1]++;
+    return this;
+  }
+
+  @Override public JsonWriter value(BufferedSource source) throws IOException {
+    if (promoteValueToName) {
+      throw new IllegalStateException(
+          "BufferedSource cannot be used as a map key in JSON at path " + getPath());
+    }
+    Object value = JsonReader.of(source).readJsonValue();
+    boolean serializeNulls = this.serializeNulls;
+    this.serializeNulls = true;
+    try {
+      add(value);
+    } finally {
+      this.serializeNulls = serializeNulls;
+    }
     pathIndices[stackSize - 1]++;
     return this;
   }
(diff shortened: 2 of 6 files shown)


moshi/src/test/java/com/squareup/moshi/JsonWriterTest.java
```diff
@@ -571,4 +572,15 @@
...
+
+  @Test public void valueFromSource() throws IOException {
+    JsonWriter writer = factory.newWriter();
+    writer.beginObject();
+    writer.name("a");
+    writer.value(new Buffer().writeUtf8("[\"value\"]"));
+    writer.name("c");
+    writer.value(3);
+    writer.endObject();
+    assertThat(factory.json()).isEqualTo("{\"a\":[\"value\"],\"c\":3}");
```
NightlyNexus: might delete this. covered in the separated tests.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -244,6 +245,24 @@ private void writeDeferredName() throws IOException {
     return this;
   }
 
+  @Override public JsonWriter value(BufferedSource source) throws IOException {
+    if (promoteValueToName) {
```
swankjesse: I don’t think we should support this.

swankjesse: ... instead we should probably just fail. It’s weird.

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
```diff
@@ -182,6 +183,28 @@ public Object root() {
...
+  @Override public JsonWriter value(BufferedSource source) throws IOException {
+    Object value = JsonReader.of(source).readJsonValue();
+    if (value instanceof String) {
+      return value((String) value);
+    }
+    if (value instanceof Number) {
+      return value((Number) value);
+    }
+    if (promoteValueToName) {
+      throw new AssertionError("This is not a map key.");
```
swankjesse: same

swankjesse: Should be IllegalStateException

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
```diff
@@ -182,6 +183,28 @@ public Object root() {
...
+    }
+    boolean serializeNulls = this.serializeNulls;
+    this.serializeNulls = true;
```
swankjesse: nice

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -335,6 +336,8 @@ public final void setSerializeNulls(boolean serializeNulls) {
    */
   public abstract JsonWriter value(@Nullable Number value) throws IOException;
 
+  public abstract JsonWriter value(BufferedSource source) throws IOException;
```
swankjesse: This is a clever API. Nice.

moshi/src/test/java/com/squareup/moshi/JsonUtf8WriterTest.java
```diff
@@ -107,4 +107,20 @@
     // JsonWriter doesn't attempt to detect duplicate names
     assertThat(buffer.readUtf8()).isEqualTo("{\"a\":1,\"a\":2}");
   }
+
+  @Test public void valueFromSource() throws IOException {
```
swankjesse: nice

moshi/src/test/java/com/squareup/moshi/JsonValueWriterTest.java
```diff
@@ -239,6 +241,22 @@
     assertThat((List<?>) writer.root()).isEqualTo(numbers);
   }
 
+  @Test public void valueFromSource() throws IOException {
```
swankjesse: awesome

moshi/src/test/java/com/squareup/moshi/PromoteNameToValueTest.java
```diff
@@ -333,4 +335,19 @@
...
+    writer.promoteValueToName();
+    writer.value(new Buffer().writeUtf8("\"a\""));
+    writer.value("a value");
+    writer.promoteValueToName();
+    writer.value(new Buffer().writeUtf8("2"));
+    writer.value("2 value");
+    writer.name("c");
+    writer.value(3);
+    writer.endObject();
+    assertThat(factory.json()).isEqualTo("{\"a\":\"a value\",\"2.0\":\"2 value\",\"c\":3}");
```
swankjesse: the `2.0` really suggests to me that promoteNameToValue is not a good fit

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
NightlyNexus: Agree! Hurts my head to think about this usage. IllegalStateException is best.

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
NightlyNexus: I might document this explicitly in the Javadoc.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
NightlyNexus: I'll write up a Javadoc since the API looks good to you!

NightlyNexus: I considered valueRaw(String) as a convenience API for Buffer.writeUtf8(), but I'm hoping this is used much more for appending from non-memory sources.

swankjesse: Yeah, offering anything that takes a String is probably a disservice; users will call that thinking it’s the same as `value(String)`

NightlyNexus:
added a Javadoc. I'm block this PR on that PR that adds the error message for invalid map keys.

PR #504, opened 2018-04-15
Model target types, parameters, constructors and properties

swankjesse:
This is intended to make it easier to implement support for subtypes.

### DIFF
--- a/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/JsonClassCodeGenProcessor.kt
+++ b/kotlin-codegen/compiler/src/main/java/com/squareup/moshi/JsonClassCodeGenProcessor.kt
@@ -15,46 +15,18 @@
  */
 package com.squareup.moshi
 
-import com.google.auto.common.AnnotationMirrors
 import com.google.auto.service.AutoService
-import com.squareup.kotlinpoet.ClassName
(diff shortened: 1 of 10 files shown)


kotlin-codegen/compiler/src/main/java/com/squareup/moshi/JsonClassCodeGenProcessor.kt
```diff
@@ -112,255 +84,44 @@ class JsonClassCodeGenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
...
+            ERROR, "No property for required constructor parameter $name", parameter.element)
         return null
       }
     }
 
     // Sort properties so that those with constructor parameters come first.
-    val propertyGenerators = propertiesByName.values.toMutableList()
-    propertyGenerators.sortBy {
+    val sortedProperties = properties.values.toMutableList()
+    sortedProperties.sortBy {
```
JakeWharton: I know this is existing code, but you can skip `toMutableList()` by just calling `sortedBy { }` directly on `values`.

swankjesse: Will fix in follow up!

PR #506, opened 2018-04-16
Begin to resolve supertype type parameters

swankjesse:


kotlin-codegen/compiler/src/main/java/com/squareup/moshi/TypeResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+  fun resolve(typeName: TypeName): TypeName {
+    return when (typeName) {
+      is ClassName -> typeName
+
+      is ParameterizedTypeName -> {
+        ParameterizedTypeName.get(
+            typeName.rawType, *(typeName.typeArguments.map { resolve(it) }.toTypedArray()))
+      }
+
+      is WildcardTypeName -> {
```
NightlyNexus: existing behavior, unrelated to the PR:
i wonder if these types are worth supporting. i'm unsure of the use case.

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AppliedType.kt
```diff
@@ -0,0 +1,71 @@
...
+        val argument = mirror.typeArguments[index]
+        return argument.asTypeName()
+      }
+    }
+  }
+
+  override fun toString() = mirror.toString()
+
+  companion object {
+    fun get(typeElement: TypeElement): AppliedType {
```
JakeWharton: extension? `TypeElement.toAppliedType()`?

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/TypeResolver.kt
swankjesse: They come up invisibly in Kotlin’s immutable List type.

kotlin-codegen/compiler/src/main/java/com/squareup/moshi/AppliedType.kt
swankjesse: Good call. Will do in follow-up.

Issue #510, opened 2018-04-24
KotlinJsonAdapter fromJson error messages state the wrong path.


NightlyNexus:
https://github.com/square/moshi/blob/1c68437f3c62a0092fab16afab84c19f5127a35c/kotlin/src/main/java/com/squareup/moshi/kotlin/KotlinJsonAdapter.kt#L90-L108

Three calls to JsonReader.getPath() are incorrect. These are always called after ending the object, not where the error actually is.

PR #511, opened 2018-04-24
Fix error message for assigning to non-null properties.

NightlyNexus:
instead of falling down to "Required property 'a' missing at $"

### DIFF
--- a/kotlin-codegen/integration-test/src/test/kotlin/com/squareup/moshi/GeneratedAdaptersTest.kt
+++ b/kotlin-codegen/integration-test/src/test/kotlin/com/squareup/moshi/GeneratedAdaptersTest.kt
@@ -383,6 +383,20 @@ class GeneratedAdaptersTest {
     }
   }
 
+  @Test fun nonNullConstructorParameterCalledWithNullFromAdapterFailsWithJsonDataException() {
+    val moshi = Moshi.Builder().add(object {
+      @FromJson fun fromJson(string: String): String? = null
+    }).build()
+    val jsonAdapter = moshi.adapter(HasNonNullConstructorParameter::class.java)
+
+    try {
+      jsonAdapter.fromJson("{\"a\":\"hello\"}")
+      fail()
+    } catch (expected: JsonDataException) {
+      assertThat(expected).hasMessage("Non-null value 'a' was null at \$.a")
+    }
+  }
+
   @JsonClass(generateAdapter = true)
   class HasNonNullConstructorParameter(val a: String)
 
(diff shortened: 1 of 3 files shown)


kotlin/src/test/java/com/squareup/moshi/kotlin/KotlinJsonAdapterTest.kt
```diff
@@ -152,6 +152,20 @@ class KotlinJsonAdapterTest {
...
+    val moshi = Moshi.Builder().add(object {
+      @FromJson fun fromJson(string: String): String? = null
+    }).add(KotlinJsonAdapterFactory()).build()
+    val jsonAdapter = moshi.adapter(HasNonNullConstructorParameter::class.java)
+
+    try {
+      jsonAdapter.fromJson("{\"a\":\"hello\"}")
+      fail()
+    } catch (expected: JsonDataException) {
+      assertThat(expected).hasMessage("Non-null value 'a' was null at \$")
```
NightlyNexus: existing behavior with incorrect path (same as nonNullConstructorParameterCalledWithNullFailsWithJsonDataException) #510.

kotlin-codegen/integration-test/src/test/kotlin/com/squareup/moshi/GeneratedAdaptersTest.kt
```diff
@@ -383,6 +383,20 @@ class GeneratedAdaptersTest {
...
+    val moshi = Moshi.Builder().add(object {
+      @FromJson fun fromJson(string: String): String? = null
+    }).build()
+    val jsonAdapter = moshi.adapter(HasNonNullConstructorParameter::class.java)
+
+    try {
+      jsonAdapter.fromJson("{\"a\":\"hello\"}")
+      fail()
+    } catch (expected: JsonDataException) {
+      assertThat(expected).hasMessage("Non-null value 'a' was null at \$.a")
```
NightlyNexus: note the different message from the null literal. `"Unexpected null at \$.a"`

swankjesse:
I like the behavior but I don’t like the impact on the generated code. Is there a way to get good behavior with less generated code?

swankjesse:
In particular, I think it’s worthwhile making different tradeoffs with generated vs. reflective code. We don’t want to generate more code than we have to, even if it makes the errors a little more awkward in corner cases. (And in practice I don’t think many adapters take non-null values and return null.)

NightlyNexus:
Yeah, I'll take another look and close out if there's not a way to do it with less code. Agree it's probably a rare case.

swankjesse:
I’m gonna merge as-is ’cause it‘s a net win. We can figure out whether we want to shrink the generated code later.

NightlyNexus:
cool, i'm pretty sure this can be improved.

PR #516, opened 2018-04-29
Update to kotlin-metadata 1.4 and use shaded compiler

hzsweers:
Per https://github.com/Takhion/kotlin-metadata/releases/tag/v1.4.0

Now the compiler is shaded and not prone to breaking on kotlin updates, making it more robust until Jetbrains releases an official API for reading metadata. The protos won't break since they're, you know, protos, but the APIs could break if the classes were moved around (as happened in kotlin 1.2.4x).


swankjesse:
Thanks!

PR #519, opened 2018-05-01
Fix path for non-null value message.

### LINKED ISSUE #510: KotlinJsonAdapter fromJson error messages state the wrong path.

NightlyNexus:
Closes #510.


NightlyNexus:
totally!

PR #524, opened 2018-05-04
Fix some tests that have the wrong expected exception message

swankjesse:
The message got improved in a conflicting change to these tests being added.


NightlyNexus:
thanks!

PR #525, opened 2018-05-07
Bump Okio to 1.15

swankjesse:


swankjesse:
Ugh, 1.15 isn’t out yet. I was looking at the POM and read 1.15.0-SNAPSHOT and did something dumb.

PR #542, opened 2018-05-15
Update Proguard configuration for codegen

pdegand:
`-keepnames` will prevent Proguard from renaming the class during obfuscation phase but won't protect the class from code shrinking. ([source from Proguard manual](https://www.guardsquare.com/en/proguard/manual/usage#keepnames))
In most cases, the generated adapter classes are never referenced directly in the code as the adapters are loaded dynamically by Moshi.
This means that, for Proguard, the classes are unused and Proguard will remove them during the shrinking phase.

The adapters need to be kept as well as there primary constructor. This is the best configuration I figured out.

Thank you for this great new feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,9 @@ Additional rules are needed if you are using Kotlin:
 ```
 *...If you are using the codegen API (i.e. `JsonClass(generateAdapter = true)`):*
 ```
--keepnames class **JsonAdapter
+-keep class **JsonAdapter {
+    <init>(...);
+}
 -keepnames @com.squareup.moshi.JsonClass class *
 ```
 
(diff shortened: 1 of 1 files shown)


swankjesse:
Awesome. Will this also keep the field names? We need that for JsonQualifiers to work!

pdegand:
They weren't, now they are.

For this simple class
```Kotlin
@Retention(AnnotationRetention.RUNTIME)
@JsonQualifier
annotation class HexColor

data class FooQualifier(
  @HexColor val color: Int
)
```
This is how the generated adapter ends up looking after proguard.
capture d ecran 2018-05-16 a 01 58 47

Does this looks good ? I never used `@JsonQualifier` so I'm not 100% sure this is fine but as the field is kept, `Types.getFieldJsonQualifierAnnotations()` will probably work just fine. :)

swankjesse:
Yeah, that’s perfect.

PR #543, opened 2018-05-16
Add more CheckReturnValues for JsonReader.

NightlyNexus:
This encourages skipName over nextName.


JakeWharton:
Need some test suppressions

PR #549, opened 2018-05-16
Fix companion object names not being resolved

### LINKED ISSUE #546: Kotlin Codegen - Named companion objects fail the build
When I create a `@JsonClass(generateAdapter = true) class Foo` and add the default `Parcelable` implementation quick fix the project will not compile. The same will happen when I simply give my companion object a different name.

> FooJsonAdapter.kt: (11, 9): Unresolved reference: Companion

The extension function is using the wrong name for the companion object.

---

Example with a named companion object:

    class Foo {
        companion object Bar
    }

Line with the error in `FooJsonAdapter`, should be `Foo.Bar` to compile:

    fun Foo.Companion.jsonAdapter(moshi: Moshi): JsonAdapter<Foo> = FooJsonAdapter(moshi)

The extension method should be either removed in this case or it should use the custom name of the companion object.

hzsweers:
This slipped through the cracks before the release

Fixes #546


hzsweers:
Rebased, looks like it works :)

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TargetType.kt
```diff
@@ -129,7 +129,10 @@ internal data class TargetType(
           properties.putIfAbsent(name, property)
         }
       }
-      return TargetType(proto, element, constructor, properties, typeVariables)
+      val companionObjectName = if (proto.hasCompanionObjectName()) {
+        typeMetadata.data.nameResolver.getQualifiedClassName(proto.companionObjectName)
+      } else null
```
swankjesse: the `{` asymmetry on this feels bad to me

PR #551, opened 2018-05-17
Fix generic typealiases

### LINKED ISSUE #547: Kotlin Codegen - Parameterized typealias fails the build
When using a `typealias` with a parameterized type moshi will generate code for the referenced base type only, omitting any type parameters.

    class Foo(
            val someMap: HashMap<String, String>
    )
    
    typealias MapStringString = HashMap<String, String>

    class Bar(
            val someMap: MapStringString
    )

`Foo` and `Bar` effectively are the same, but `Bar` is using a `typealias`. The adapter for `Foo` will be properly generated, but it will fail for `Bar` with:

> BarJsonAdapter.kt: (16, 45): 2 type arguments expected for class HashMap<K : Any!, V : Any!>
> BarJsonAdapter.kt: (21, 22): 2 type arguments expected for class HashMap<K : Any!, V : Any!>

The generated code in question looks as follows, clearly missing the parameters:

    private val hashMapAdapter: JsonAdapter<HashMap> = moshi.adapter(HashMap::class.java).nonNull()

    override fun toString(): String = "GeneratedJsonAdapter(Bar)"

    override fun fromJson(reader: JsonReader): Bar {
        var someMap: HashMap? = null
        reader.beginObject()
        // ...
    }

This will work properly for simple typealiases like `String` and as a workaround the typealias can be replaced with the referenced type itself.

It would be nice if there was full support for typealiases.

hzsweers:
We had a missing `resolveAliases` check when gathering argument lists 🙄 

This is important because the argument is on the backing type, not on the aliased one.

As part of it, I also reworded the API to be a bit more understandable. Now instead of the confusing `resolveAliases` and always being negated when used with `hasAbbreviatedType()`, it's called `useAbbreviatedType` (which matches how it's referred to in the kotlin compiler as well).

Resolves #547

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TargetType.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TargetType.kt
@@ -182,7 +182,7 @@ internal data class TargetType(
       for (property in classProto.propertyList) {
         val name = nameResolver.getString(property.name)
         val type = typeResolver.resolve(property.returnType.asTypeName(
-            nameResolver, classProto::getTypeParameter, true))
+            nameResolver, classProto::getTypeParameter, false))
         result[name] = TargetProperty(name, type, property, constructor.parameters[name],
             annotationHolders[name], fields[name], setters[name], getters[name])
       }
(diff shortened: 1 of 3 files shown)


swankjesse:
Nice refactor.

PR #566, opened 2018-05-29
Update kotlin version

joshfriend:
I've been seeing a warning in Android Studio about having a version on kotlin-stdlib that doesn't match the kotlin plugin version, so I've updated it (and kotlin-metadata) to the latest versions available.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,8 +29,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.version>1.7</java.version>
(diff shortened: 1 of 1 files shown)


JakeWharton:
You should be declaring the Kotlin version and dependency yourself to prevent this warning. We will never make a release just to bump transitive dependencies unless there is a required bug fix in them.

joshfriend:
I did declare the kotlin stdlib dependency explicitly to be the latest version, but still saw a warning in AS.



Could this instead be an AS bug, or am I just tripping? Moshi is the only thing in my dependency tree that requests this version of kotlin-stdlib.

PR #570, opened 2018-06-04
(WIP Prototype) Switch to public kotlinx-metadata API

hzsweers:
This swaps out kotlin-metadata for the new [kotlinx-metadata](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm) API

I _think_ this is actually done, but I'm unable to get the build working in the `kotlin/tests` project due to `MetadataExtensions` services not being visible for some reason. They work in the `codegen/src/test` tests though.

The high level architecture:
  - Switch to new `Data` class representations of specific types, replacing the raw protos they were reading before.
  - The `metadata.kt` class has been switched to provide APIs for resolving `TypeName`s from the library's `KmVisitor` APIs. This is a significant departure from before since it's a visitor pattern, but ended up working nicely once I wrote `TypeNameKmTypeVisitor`.
  - Shade the kotlinx-metadata library since it's not stable yet, and also avoid dependence on the kotlinx bintray repo

The other changes were pretty lightweight after migrating `metadata.kt`. Overall pretty happy with how this turned out, and definitely cleaner than reading the raw protos from before.

### DIFF
--- a/kotlin/codegen/pom.xml
+++ b/kotlin/codegen/pom.xml
@@ -12,6 +12,14 @@
 
   <artifactId>moshi-kotlin-codegen</artifactId>
 
+  <repositories>
+    <repository>
+      <id>bintray-kotlin-kotlinx</id>
+      <name>bintray</name>
+      <url>https://kotlin.bintray.com/kotlinx</url>
+    </repository>
+  </repositories>
+
   <dependencies>
     <dependency>
       <groupId>com.squareup.moshi</groupId>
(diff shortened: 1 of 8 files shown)


kotlin/codegen/pom.xml
```diff
@@ -176,6 +186,11 @@
                   <pattern>com.squareup.kotlinpoet</pattern>
                   <shadedPattern>com.squareup.moshi.kotlinpoet</shadedPattern>
                 </relocation>
+                <!-- Repackage Kotlinx-metadata until its API is stable. -->
+                <relocation>
+                  <pattern>kotlinx.metadata.jvm</pattern>
```
udalov: I suppose you need to relocate `kotlinx.metadata` here, not `kotlinx.metadata.jvm`. I think this may fix the ServiceLoader issue.

hzsweers: Unfortunately this happens even if I disable shading

hzsweers:
So an update - this will either need to wait for 0.0.3 of the library and possibly Kotlin 1.2.60, as the library's compiled against APIs that aren't available yet in kotlin 1.2.50

hzsweers:
I have this moooostly working now in https://github.com/hzsweers/moshi/tree/z/kotlinxmetadata1260

I get strange issues with the runtime reflective lookup of adapters not working though. Wondering if kotlin 1.2.6x regressed/changes generated sources linking in kapt

hzsweers:
Confirmed kapt sources location moved. Nearly there!

PR #580, opened 2018-06-14
Trailing comma

johnjohndoe:
# Description
- I like to add a test for a **trailing comma** since this is a common mistake which is hard to find. I hope that the issue will be easier to discover by others by adding the test and the exception message to the sources.

# Not ready to merge :x: 
- Currently, the `[Utf8]` test succeeds while the `[Value`] test **fails**. I could not yet figure out why. Maybe you can help with that.

# Future improvements
- You might want to discuss whether it would be possible to throw a more precise exception such as `TrailingCommaException: Trailing comma detected at path $.a` instead of `JsonEncodingException: Expected name at path $.`

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
@@ -165,6 +165,20 @@ public final class JsonReaderTest {
     assertThat(reader.peek()).isEqualTo(JsonReader.Token.END_DOCUMENT);
   }
 
+  @Test public void failOnTrailingComma() throws IOException {
+    JsonReader reader = newReader("{\"a\": \"android\", }");
+    reader.beginObject();
+    assertThat(reader.nextName()).isEqualTo("a");
+    assertThat(reader.nextString()).isEqualTo("android");
+    try {
+      //noinspection ResultOfMethodCallIgnored
+      reader.nextName();
+      fail();
+    } catch (JsonEncodingException expected) {
+      assertThat(expected).hasMessage("Expected name at path $.a");
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -165,6 +165,20 @@ JsonReader newReader(String json) throws IOException {
     assertThat(reader.peek()).isEqualTo(JsonReader.Token.END_DOCUMENT);
   }
 
+  @Test public void failOnTrailingComma() throws IOException {
```
swankjesse: we already have this, com.squareup.moshi.JsonUtf8ReaderTest#strictExtraCommasInMaps

johnjohndoe: Ah! Thanks for pointing this out. I did not find it. Here are a few follow-up questions:

1. Should we assert for the error message like I did in my test?
2. Why is your test named `InMaps`?
3. Why is my test failing for `[Value]` or why is your test in `JsonUtf8ReaderTest` instead in `JsonReaderTest`?
4. Do you have a opinion about the `TrailingCommaException` proposal or do you prefer to discuss this in a separate issue?

swankjesse:
Could you clarify your problem? When you have a trailing comma what does Moshi do, and what do you expect?

johnjohndoe:
The initial reason which made me looking into the sources was that I had a trailing comma in one of my files. The resulting error message `JsonEncodingException: Expected name at path $.` was not verbose enough for me to understand the source of the problem quick enough. Therefore, I would like to discuss if this can be improved.
Further, the questions above came up later.

ZacSweers:
Nothing to be done here for now

PR #582, opened 2018-06-16
Fix GeneratedAnnotation member definition

hzsweers:
Before this was just adding both as varargs of "value"


swankjesse:
Ick. We should have tests for this.

PR #584, opened 2018-06-22
Fix hasNext to return false at document end.

NightlyNexus:
I doubt anybody is relying on the previous behavior.


JakeWharton:
I'm good with the behavior change.

swankjesse:
Good call.

PR #592, opened 2018-07-03
Update readme to add information about depending on stdlib.

runningcode:


README.md
```diff
@@ -544,6 +544,9 @@ add the following to your build to enable the annotation processor:
 kapt 'com.squareup.moshi:moshi-kotlin-codegen:1.6.0'
 ```
 
+You must also add the `kotlin-stdlib` as a dependency in order for the metadata annotations
```
hzsweers: Maybe "You must also have the `kotlin-stdlib` dependency on the classpath during compilation in order for the compiled code to have the required metadata annotations that Moshi's processor looks for.

Basically to clarify that it's not like they're there and not visible, but rather that they're not generated at all without it.

runningcode:
Updated with @hzsweers 's wording. Thanks!

swankjesse:
Thanks!

PR #596, opened 2018-07-11
Parsing String to Map -> Map to String fails in Kotlin

andretietz:
This is a small Bug which appears to happen only in kotlin.

### DIFF
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -849,6 +849,14 @@ class KotlinJsonAdapterTest {
     assertThat(adapter.fromJson(json)).isEqualTo(value)
     assertThat(adapter.toJson(value)).isEqualTo(json)
   }
+  
+  @Test fun backAndForthMap() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+    val json = """{"map-key1":"map-value","map-key2":"map-value"}"""
+    val map = moshi.adapter(Map::class.java).fromJson(json)!!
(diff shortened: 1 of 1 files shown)


andretietz:
I've seen this and moshi is trying to downcast the class, but ends up with the error(see logs):

``` java.lang.IllegalArgumentException: No JsonAdapter for java.util.Comparator<? super K> (with no annotations)```

I would've expected that it casts down to Map itself.

Same test using (not only)gson does work, too.

I would like to work parser independent, so I only have an interface passing a string and the expected type, returning an instance of the expected type.

NightlyNexus:
Gson is wrong for trying to serialize types that you don't own.
If you're asking for an adapter for a LinkedHashTreeMap, Moshi won't (or at least shouldn't) give it to you. LinkedHashTreeMap is an implementation detail that could change structure at any time.

As an aside, that's an unfortunate Parser abstraction, since you'll need to look up the JsonAdapter for the type for every call.

I can't exactly offer advice, since I don't see the picture of what we're trying to achieve. Maybe post on Stack Overflow with the Moshi tag and a full code example of what we're trying to get working, or maybe somebody else here sees this better?

andretietz:
Since moshi can parse a json string into `Map<String, Any>`, shouldn't it be possible to do it the other way around, too? At least if there's no adapter set up?

I'll try to setup a stackoverflow question for it.

andretietz:
As closer I was looking to the problem, the more I was convinced that the method I am using on how to parse that was wrong.

What I need is a half-parsed object, in order to parse the elements in the app later. 

The idea was to parse an unknown part into a `Map<String, Any>` (which was working so far) and when needed parse this map back to json(this is where I got into problems with moshi) and then parse only the part of the map into an object with a given runtime class object.

What I did now:
I created an adapter which parses the unknown part directly back into a json string, so that I have a `Map<String, String>`

Whenever I now need a part of it I need the key and an adapter for that json string in order to parse it.  I am still not very convinced about this approach, but it works so far. If you have any suggestions on how to do this better, please let me know.

PR #606, opened 2018-07-27
Add RuntimeJsonAdapterFactory to adapters module.

NightlyNexus:
I made this simple instead of flexible.

This is slightly different from Gson's RuntimeJsonAdapterFactory.
It doesn't remove the type value when deserializing and add it when serializing because I don't know why it should. Let me know if there's a use case or if consistency with Gson is important.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
@@ -71,7 +71,7 @@ public final class RuntimeJsonAdapterFactory<T> implements JsonAdapter.Factory {
       labelToAdapter.put(label, adapter);
       typeToAdapter.put(typeValue, adapter);
     }
-    return new RuntimeJsonAdapter(labelKey, labelToAdapter, typeToAdapter);
+    return new RuntimeJsonAdapter(labelKey, labelToAdapter, typeToAdapter).nullSafe();
   }
 
   static final class RuntimeJsonAdapter extends JsonAdapter<Object> {
--- a/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
@@ -1,0 +1,143 @@
+package com.squareup.moshi.adapters;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonDataException;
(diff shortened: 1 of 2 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -0,0 +1,143 @@
+package com.squareup.moshi.adapters;
```
swankjesse: nit: copyright header. If this is derived from Gson’s, copy over that copyright header?

adapters/src/test/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactoryTest.java
```diff
@@ -0,0 +1,205 @@
+package com.squareup.moshi.adapters;
```
swankjesse: same

adapters/src/test/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactoryTest.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Map;
+import okio.Buffer;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
+
+@SuppressWarnings("CheckReturnValue")
+public final class RuntimeJsonAdapterFactoryTest {
+  @Test public void fromJson() throws IOException {
```
swankjesse: nice

adapters/src/test/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactoryTest.java
```diff
@@ -0,0 +1,205 @@
...
+    try {
+      adapter.fromJson("{\"type\":{},\"value\":\"Okay!\"}");
+      fail();
+    } catch (JsonDataException expected) {
+      assertThat(expected).hasMessage("Label for 'type' must be a string but was {},"
+          + " a class com.squareup.moshi.LinkedHashTreeMap");
+    }
+  }
+
+  @Test public void nonObjectDoesNotConsume() throws IOException {
```
swankjesse: nice test!

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
NightlyNexus: thanks, done!

PR #607, opened 2018-07-27
Add EnumJsonAdapter to adapters module.

NightlyNexus:
for consideration. dunno what to call this type.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
@@ -1,0 +1,88 @@
+package com.squareup.moshi.adapters;
+
+import com.squareup.moshi.Json;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonDataException;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import java.io.IOException;
+import java.util.Arrays;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A JsonAdapter for enums that allows having a fallback enum value when a deserialized string does
+ * not match any enum value.
+ */
+public final class EnumJsonAdapter<T extends Enum<T>> extends JsonAdapter<T> {
+  final Class<T> enumType;
+  final String[] nameStrings;
+  final T[] constants;
+  final JsonReader.Options options;
+  final @Nullable T fallbackValue;
+
+  public static <T extends Enum<T>> EnumJsonAdapter<T> create(Class<T> enumType) {
+    return new EnumJsonAdapter<>(enumType, null);
+  }
+
+  /**
+   * Create a new adapter for this enum with a fallback value to use when the JSON string does not
+   * match any of the enum's constants. Note that this value will not be used when the JSON value is
+   * null, absent, or not a string. Also, the string values are case-sensitive, and this fallback
+   * value will be used even on case mismatches.
+   */
+  public EnumJsonAdapter<T> withUnknownFallback(T fallbackValue) {
+    if (fallbackValue == null) {
+      throw new NullPointerException("fallbackValue == null");
+    }
+    return new EnumJsonAdapter<>(enumType, fallbackValue);
+  }
+
+  EnumJsonAdapter(Class<T> enumType, @Nullable T fallbackValue) {
+    this.enumType = enumType;
+    this.fallbackValue = fallbackValue;
+    try {
+      constants = enumType.getEnumConstants();
+      nameStrings = new String[constants.length];
+      for (int i = 0; i < constants.length; i++) {
+        String constantName = constants[i].name();
+        Json annotation = enumType.getField(constantName).getAnnotation(Json.class);
+        String name = annotation != null ? annotation.name() : constantName;
+        nameStrings[i] = name;
+      }
+      options = JsonReader.Options.of(nameStrings);
+    } catch (NoSuchFieldException e) {
+      throw new AssertionError("Missing field in " + enumType.getName(), e);
+    }
+  }
+
+  @Override public @Nonnull T fromJson(JsonReader reader) throws IOException {
+    int index = reader.selectString(options);
+    if (index != -1) return constants[index];
+
+    String path = reader.getPath();
+    if (fallbackValue == null) {
+      String name = reader.nextString();
+      throw new JsonDataException("Expected one of "
+          + Arrays.asList(nameStrings) + " but was " + name + " at path " + path);
+    }
+    if (reader.peek() != JsonReader.Token.STRING) {
+      throw new JsonDataException(
+          "Expected a string but was " + reader.peek() + " at path " + path);
+    }
+    reader.skipValue();
+    return fallbackValue;
+  }
+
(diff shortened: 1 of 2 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
```diff
@@ -0,0 +1,72 @@
+package com.squareup.moshi.adapters;
```
swankjesse: nit: copyright header

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+import com.squareup.moshi.JsonWriter;
+import java.io.IOException;
+import java.util.Arrays;
+import javax.annotation.Nullable;
+
+/**
+ * A JsonAdapter for enums that allows having a fallback enum value when a deserialized string does
+ * not match any enum value.
+ */
+public final class ConfigurableEnumJsonAdapter<T extends Enum<T>> extends JsonAdapter<T> {
```
swankjesse: I wonder if this should be just called EnumJsonAdapter ? The other one isn’t a public type so I don’t think it matters

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+  final String[] nameStrings;
+  final T[] constants;
+  final JsonReader.Options options;
+  final @Nullable T fallbackValue;
+
+  public static <T extends Enum<T>> ConfigurableEnumJsonAdapter<T> create(Class<T> enumType) {
+    return new ConfigurableEnumJsonAdapter<>(enumType, null);
+  }
+
+  public ConfigurableEnumJsonAdapter<T> withFallbackValue(T fallbackValue) {
```
swankjesse: Should document when we fallback? Like only when an enum is unknown. (But not absent or null!)

Maybe should be `withUnknownValue()` ?

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+
+  ConfigurableEnumJsonAdapter(Class<T> enumType, @Nullable T fallbackValue) {
+    this.enumType = enumType;
+    this.fallbackValue = fallbackValue;
+    try {
+      constants = enumType.getEnumConstants();
+      nameStrings = new String[constants.length];
+      for (int i = 0; i < constants.length; i++) {
+        T constant = constants[i];
+        Json annotation = enumType.getField(constant.name()).getAnnotation(Json.class);
```
swankjesse: this is great

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
swankjesse: Might be extra worth a warning to say that this adapter is case-sensitive and that the fallback will be used even on just a case mismatch

adapters/src/test/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapterTest.java
```diff
@@ -0,0 +1,46 @@
+package com.squareup.moshi.adapters;
```
swankjesse: nit: copyright header

adapters/src/test/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapterTest.java
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      adapter.fromJson("\"SPOCK\"");
+      fail();
+    } catch (JsonDataException expected) {
+      assertThat(expected).hasMessage(
+          "Expected one of [ROCK, PAPER, scr] but was SPOCK at path $");
+    }
+  }
+
+  @Test public void withFallbackValue() throws Exception {
```
swankjesse: nice

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
NightlyNexus: yes, totally. good call.

adapters/src/main/java/com/squareup/moshi/adapters/ConfigurableEnumJsonAdapter.java
NightlyNexus: oops, i missed putting in a doc. will update.

`withUnknownValue` reads strange to me. it sounds like the value passed in is unknown, when it is really the substitute for unknown values.

swankjesse: Maybe withUnknownFallback() ?

NightlyNexus: sounds good.

adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
```diff
@@ -0,0 +1,88 @@
...
+    if (fallbackValue == null) {
+      String name = reader.nextString();
+      throw new JsonDataException("Expected one of "
+          + Arrays.asList(nameStrings) + " but was " + name + " at path " + path);
+    }
+    if (reader.peek() != JsonReader.Token.STRING) {
+      throw new JsonDataException(
+          "Expected a string but was " + reader.peek() + " at path " + path);
+    }
+    reader.skipValue();
```
NightlyNexus: changed to skipValue just now.

edenman:
Is there a reason this doesn't allow `null` as the fallback value?

edenman:
/shrug I find it cleaner to model as `MyEnumType?`...seems weird to prescribe an approach: why not just default to the "return null if not a known value" behavior and instead use a fallback if one was provided?  I guess this might be why moshi didn't ship with one of these in the first place, everybody wants something slightly different :)

hzsweers:
you can exhaustively switch on unknown, I find that to be a nicer pattern

JakeWharton:
Not allowing null is weird. What if you don't control the enum and thus cannot add an `UNKOWN`?

shanyuliang:
Agree with @edenman , in my working project we do have 10+ enum classes. For centain enum class we sure can provide our own fallback value. But would be cool if we have a general "enum-fallback-to-null" solution for all other enums.

NightlyNexus:
The factory that makes enum adapters that always fall back to the universal null value is tempting.

As for this adapter, I'm really not opinionated about adding null as a possible fallback value. PR welcome, or I'll get to it at some point. Just needs a flag to track if a fallback has been set for the adapter.

swankjesse:
Yeah I think a null fallback is handy.

NightlyNexus:
#640

PR #611, opened 2018-08-05
Fix jsonAdapter extension functions on generics, demo KClass alternative

hzsweers:
The existing implementation of `jsonAdapter` extension functions on `companion object`s is currently broken if the target class has generics. This fixes that by using reified generics. Not sure how to best go about this, as the previous implementation with generics didn't work anyway. We could have the `types` parameter instead of reified generics, but that leaves an awkward API hole where a user could define the generic on `jsonAdapter` with different ones than what are passed in the `types` parameter. Note that the IDE marks the generic specification in the function body as unnecessary, but this appears to be a bug since it doesn't compile without it and I plan to file separately.

The alternative implementation proposed along with this is using `KClass` as the receiver for the extension function rather than rely on a `companion object`, similar to how Kotlinx-serialization works for its adapters. The IDE does mark the receiver as unnecessary, but it is necessary for scoping (and works wonderfully in the IDE autocomplete). As this is a very different API, there's a few ways we could go about it if we want to pursue it. Namely, the implementation I show here is a deprecation approach where a full deprecated annotation + autofix. This is nice because users can run a "fix-all" on their code base. Other approaches are to remove the `companion object` support entirely, or not use the `KClass` support at all.

A final option is to just remove this API all together, and leave it to a 3rd party library to do instead (though it would probably never be able to be stable since it would have to "predict" the names of the generated JsonAdapter classes)

Proposing this up for discussion/review/demo/bug report :)

### Example

For this snippet:

```kotlin
internal data class NamedGenericCompanionObjectClass<T, R>(val foo: T, val bar: R) {
  companion object CustomCompanionObject
}
```

#### Before 

(missing type variables and `types` argument)

```kotlin
NamedGenericCompanionObjectClass.CustomCompanionObject.jsonAdapter(moshi: Moshi): JsonAdapter<GeneratedAdaptersTest.NamedGenericCompanionObjectClass> = NamedGenericCompanionObjectClassJsonAdapter(moshi)
```

#### After

```kotlin
internal inline fun <reified T, reified R> KClass<NamedGenericCompanionObjectClass<T, R>>.jsonAdapter(moshi: Moshi): JsonAdapter<NamedGenericCompanionObjectClass<T, R>> = NamedGenericCompanionObjectClassJsonAdapter<T, R>(moshi, arrayOf(T::class.java, R::class.java))

@Deprecated(
        message = "Replaced by {the class}::class.jsonAdapter(...)",
        replaceWith = ReplaceWith(expression = "com.squareup.moshi.kotlin.codgen.GeneratedAdaptersTest.NamedGenericCompanionObjectClass::class.jsonAdapter(moshi)"),
        level = DeprecationLevel.WARNING
)
internal inline fun <reified T, reified R> NamedGenericCompanionObjectClass.CustomCompanionObject.jsonAdapter(moshi: Moshi): JsonAdapter<NamedGenericCompanionObjectClass<T, R>> = NamedGenericCompanionObjectClassJsonAdapter<T, R>(moshi, arrayOf(T::class.java, R::class.java))
```


swankjesse:
Interesting. I don’t think these `jsonAdapter()` methods are very useful. Can we just omit ’em for companion objects? Perhaps we can get rid of ’em altogether?

hzsweers:
That's an option, but technically a breaking change. It'd be nice if `Types` exposed a mechanism for deducing the expected generated JsonAdapter class name of a `@JsonClass`-annotated class so another library could potentially build something on top of this. What do you think?

hzsweers:
nixing in favor of #687

PR #614, opened 2018-08-07
Add new error message for fields missing type adapters.

### LINKED ISSUE #557: Feature Request: include enclosing type name when parameterized platform time is missing adapter
I received the following error message when converting a deeply nested pojo into json :

```
java.lang.IllegalArgumentException: Platform java.util.ArrayList<java.lang.String> annotated [] requires explicit JsonAdapter to be registered
```

It would be beneficial for the message to enclude the enclosing type's name to make it easier to know what field is the culprit

NightlyNexus:
The error message says the field that is missing an adapter, so you don't need to use a debugger.
The previous error message is moved to the exception's cause.

Closes #557 and may help folks on [SO](https://stackoverflow.com/questions/51669838/moshi-for-parsing-youtube-search-api).

If this is a good idea, maybe the Kotlin adapters should do something similar.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -95,7 +95,13 @@ final class ClassJsonAdapter<T> extends JsonAdapter<T> {
         // Look up a type adapter for this type.
         Type fieldType = resolve(type, rawType, field.getGenericType());
         Set<? extends Annotation> annotations = Util.jsonAnnotations(field);
-        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations);
+        JsonAdapter<Object> adapter;
+        try {
+          adapter = moshi.adapter(fieldType, annotations);
+        } catch (IllegalArgumentException e) {
+          throw new IllegalArgumentException("Error creating adapter for field '" + field.getName()
+              + "' in " + type, e);
+        }
 
         // Create the binding between field and JSON.
         field.setAccessible(true);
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -95,7 +95,13 @@ private void createFieldBindings(
         // Look up a type adapter for this type.
         Type fieldType = resolve(type, rawType, field.getGenericType());
         Set<? extends Annotation> annotations = Util.jsonAnnotations(field);
-        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations);
+        JsonAdapter<Object> adapter;
+        try {
+          adapter = moshi.adapter(fieldType, annotations);
+        } catch (IllegalArgumentException e) {
```
JakeWharton: This is a bit specific. Worth doing `RuntimeException` since we're running user code here?

swankjesse:
I like this. Can we promote it to be part of the framework? Perhaps Moshi.java could generate the chain by tracking re-entrant calls? May also be useful to allow Moshi.adapter() to specify a JSON path fragment (like `.name`or whatever)

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
NightlyNexus: was mirroring https://github.com/square/moshi/blob/29bdc0aa45cc80569c2524b2f8e370d0649bdd80/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java#L53

i think you're right.

NightlyNexus:
that's cool! i'll take a stab at that when i get some time to think it through!

NightlyNexus:
closing in favor of #616 when I get back to it!

PR #616, opened 2018-08-10
Track field names in the adapter lookup stack.

NightlyNexus:
This allows for an error message that includes field names to track down the cause of adapter creation failure for deeply nested structures.

This got complicated, so I am not sure I have a good solution yet.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -95,14 +95,15 @@ final class ClassJsonAdapter<T> extends JsonAdapter<T> {
         // Look up a type adapter for this type.
         Type fieldType = resolve(type, rawType, field.getGenericType());
         Set<? extends Annotation> annotations = Util.jsonAnnotations(field);
-        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations);
+        String fieldName = field.getName();
+        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations, fieldName);
 
         // Create the binding between field and JSON.
         field.setAccessible(true);
 
         // Store it using the field's name. If there was already a field with this name, fail!
         Json jsonAnnotation = field.getAnnotation(Json.class);
-        String name = jsonAnnotation != null ? jsonAnnotation.name() : field.getName();
+        String name = jsonAnnotation != null ? jsonAnnotation.name() : fieldName;
         FieldBinding<Object> fieldBinding = new FieldBinding<>(name, field, adapter);
         FieldBinding<?> replaced = fieldBindings.put(name, fieldBinding);
         if (replaced != null) {
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -90,8 +90,14 @@ public final class Moshi {
   }
 
   @CheckReturnValue
-  @SuppressWarnings("unchecked") // Factories are required to return only matching JsonAdapters.
   public <T> JsonAdapter<T> adapter(Type type, Set<? extends Annotation> annotations) {
+    return adapter(type, annotations, null);
+  }
+
+  @CheckReturnValue
+  @SuppressWarnings("unchecked") // Factories are required to return only matching JsonAdapters.
+  <T> JsonAdapter<T> adapter(Type type, Set<? extends Annotation> annotations,
+      @Nullable String fieldName) {
     if (type == null) {
       throw new NullPointerException("type == null");
     }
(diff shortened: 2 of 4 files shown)


moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -133,20 +148,46 @@
...
       }
+    } catch (RuntimeException e) {
+      if (lastIndex == 0) { // Rewrite the exception at the top level.
+        e = errorWithFields(type, fieldNames, e);
+      }
+      throw e;
     } finally {
-      deferredAdapters.remove(deferredAdapters.size() - 1);
+      deferredAdapters.remove(lastIndex);
       if (deferredAdapters.isEmpty()) {
```
swankjesse: We should combine the list with the deferred adapters. Maybe give a deferred adapter a name?

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -133,20 +148,46 @@
...
 
+  static RuntimeException errorWithFields(Type type, List<String> fieldNames, RuntimeException e) {
+    int size = fieldNames.size();
+    if (size == 0) {
+      return e;
+    }
+    StringBuilder stringBuilder = new StringBuilder("Error creating adapter for field ")
+        .append(Types.getRawType(type).getName());
+    for (int i = 0; i < size; i++) {
+      stringBuilder
```
swankjesse: I think there should be a delimiter before the list?

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -95,14 +95,15 @@ private void createFieldBindings(
         // Look up a type adapter for this type.
         Type fieldType = resolve(type, rawType, field.getGenericType());
         Set<? extends Annotation> annotations = Util.jsonAnnotations(field);
-        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations);
+        String fieldName = field.getName();
+        JsonAdapter<Object> adapter = moshi.adapter(fieldType, annotations, fieldName);
```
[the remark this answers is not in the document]
NightlyNexus: I did that deliberately. I assumed users would look for the Java field name instead of the JSON name.

I guess it depends on what the error message should look like? If somebody has a field `@Json("image_urls") ArrayList<HttpUrl> imageUrls`, "Error creating adapter for field User.Profile.imageUrls" seems better than "Error creating adapter for field User.Profile.image_urls"

Or should we treat it like a JSON path?

moshi/src/main/java/com/squareup/moshi/Moshi.java
NightlyNexus: Do you mean make a DeferredAdapter subclass to hold onto the field stack?

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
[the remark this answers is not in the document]
NightlyNexus: i'll have to rethink the error message.
here's a case where the message would be confusing and wrong.
```
final class Foo {
  static final class Bar {
    ArrayList<String> field;
  }

  Map<String, Bar> map;
}
```
this would have some error message about "Foo.map.field"

NightlyNexus:
took a wild stab at this.
at the top of the stack (first entry), a list is kept of types and field names. whenever a reentrant call comes in to this new internal adapter lookup method, a field name is added. whenever a reentrant call comes into to any other adapter lookup method, the type is added. it makes error messages like in the added test case.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -133,11 +156,19 @@
...
         }
       }
+    } catch (RuntimeException e) {
+      if (lastIndex == 0) { // Rewrite the exception at the top level.
+        e = errorWithFields(typesAndFieldNames, e);
+      }
+      throw e;
     } finally {
-      deferredAdapters.remove(deferredAdapters.size() - 1);
+      deferredAdapters.remove(lastIndex);
```
swankjesse: it’s weird that we remove from `deferredAdapters` here, but from `typesAndFieldNames` in the loop. Should be the same place.

NightlyNexus: yeah, can do!

moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -955,6 +956,53 @@
...
+    static final class Wrapper {
+      HasPlatformType hasPlatformType;
+    }
+
+    static final class ListWrapper {
+      List<HasPlatformType> platformTypes;
+    }
+  }
+
+  @Test public void reentrantFieldErrorMessages() {
```
NightlyNexus: gonna split this into 3 tests.

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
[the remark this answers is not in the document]
NightlyNexus: for posterity, the error message would now be about `field 'com.squareup.moshi.Moshi$Foo$Bar.field' in field 'com.squareup.moshi.Moshi$Foo.map'`.

NightlyNexus:
apologies for the back-and-forth. i think we can do this now with one stack, since now adapters are one-to-one with types or field names.

NightlyNexus:
...I think this needs a reasonable public API, actually.

swankjesse:
The current API exposes just a hint on what field name we should report if an adapter lookup is problematic. What else should we add / remove?

NightlyNexus:
yep! I played with this a little more, and I'm more confident this can be public. I'll update to make it public as-is and send followup PRs for the Kotlin adapter and codegen.

moshi/src/test/java/com/squareup/moshi/MoshiTest.java
```diff
@@ -955,6 +956,74 @@
...
+              + "\nfor java.util.ArrayList<java.lang.String> strings"
+              + "\nfor class com.squareup.moshi.MoshiTest$HasPlatformType hasPlatformType"
+              + "\nfor class com.squareup.moshi.MoshiTest$HasPlatformType$Wrapper");
+      assertThat(e).hasCauseExactlyInstanceOf(IllegalArgumentException.class);
+      assertThat(e.getCause()).hasMessage("Platform java.util.ArrayList<java.lang.String> "
+          + "(with no annotations) requires explicit JsonAdapter to be registered");
+    }
+  }
+
+  @Test public void reentrantFieldErrorMessagesListWrapper() {
```
swankjesse: this is great!

NightlyNexus:
thanks for the help and improvements to this error message!

PR #622, opened 2018-08-14
Fix README link to jar.

NightlyNexus:
Sonatype changed its site.

I'm not sure if there's a way to link directly to the latest jar now.


JakeWharton:
1 down, 50 to go!

PR #623, opened 2018-08-15
Generate Kotlin documentation with Dokka

swankjesse:


kotlin/tests/src/assembly/dokka.xml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3 http://maven.apache.org/xsd/assembly-1.1.3.xsd">
+  <id>javadoc</id>
+  <formats>
+    <format>jar</format>
+  </formats>
+  <baseDirectory>/</baseDirectory>
+  <fileSets>
+    <fileSet>
+      <directory>target/dokka/moshi-kotlin-tests</directory>
```
hzsweers: Is this to upload docs for tests?

swankjesse: It's simplest if all of our modules are symmetric. Documenting a test-only module isn’t useful, but making it special means that deploying docs with Osstrich is awkward.

Issue #626, opened 2018-08-17
Moshi with Kotlin codegen ignores custom adapters?


marcosalis:
I'm using Moshi 1.6 with the new codegen processor with Retrofit to parse Json response models.
I have a Kotlin data class annotated with `@JsonClass(generateAdapter = true)`, and a simple custom type adapter to convert some string properties to longs:

```
public class StringLongMoshiAdapter {
    @ToJson @NonNull String toJson(@Nullable Long aLong) {
        return String.valueOf(aLong);
    }
    @FromJson @Nullable Long fromJson(@Nullable String longString) {
        if (longString == null || longString.isEmpty()) {
            return null;
        }
        return Long.parseLong(longString);
    }
}
```

This is how I initialize Moshi:
```
new Moshi.Builder()
                    .add(new StringLongMoshiAdapter())
                    .build();
```

And this is (part) of the data class:
```
@Keep
@JsonClass(generateAdapter = true)
internal data class MyResponse(
(...)
        @Json(name = "parent-id") val parentId: Long?,
```

However, when parsing I get this exception, which means that the custom adapter is not used (this was working before, with `KotlinJsonAdapterFactory` and `moshi-kotlin`):
```
com.squareup.moshi.JsonDataException: Expected a long but was  at path $.categories[0].parent-id
        at com.squareup.moshi.JsonUtf8Reader.nextLong(JsonUtf8Reader.java:788)
```

Is this expected behaviour? Shouldn't the generated type adapter for `MyResponse` use the custom type adapters specified when initialising Moshi?

NightlyNexus:
Thanks for the report.
I ran into something like this a few months ago and forgot to file. IIRC there's a missing .javaObjectType, and it's tangentially related to #586.

marcosalis:
Wow, that was quick! Thanks a lot guys, I appreciate it.
I can confirm that the issue is sorted in `1.7.0-SNAPSHOT`.

PR #627, opened 2018-08-17
Make nullable primitives in codegen use boxed type adapters.

NightlyNexus:
```kotlin
private val nullableBooleanAdapter: JsonAdapter<Boolean?> =
        moshi.adapter(Boolean::class.javaObjectType).nullSafe()
```
instead of
```kotlin
private val nullableBooleanAdapter: JsonAdapter<Boolean> = moshi.adapter(Boolean::class).nullSafe()
```

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
@@ -43,6 +43,16 @@ abstract class TypeRenderer {
 
   fun render(typeName: TypeName): CodeBlock {
     if (typeName.nullable) {
+      if (typeName == BOOLEAN.asNullable()
+          || typeName == BYTE.asNullable()
+          || typeName == CHAR.asNullable()
+          || typeName == DOUBLE.asNullable()
+          || typeName == FLOAT.asNullable()
+          || typeName == INT.asNullable()
+          || typeName == LONG.asNullable()
+          || typeName == SHORT.asNullable()) {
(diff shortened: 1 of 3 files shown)


NightlyNexus:
Closes #626.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
```diff
@@ -43,6 +43,16 @@ abstract class TypeRenderer {
...
   fun render(typeName: TypeName): CodeBlock {
     if (typeName.nullable) {
+      if (typeName == BOOLEAN.asNullable()
+          || typeName == BYTE.asNullable()
+          || typeName == CHAR.asNullable()
+          || typeName == DOUBLE.asNullable()
+          || typeName == FLOAT.asNullable()
+          || typeName == INT.asNullable()
+          || typeName == LONG.asNullable()
+          || typeName == SHORT.asNullable()) {
```
JakeWharton: `typeName.isBoxedPrimitive()` should work here

NightlyNexus: that exists in JavaPoet. is there something for KotlinPoet? i'm a little behind.

JakeWharton: Oh right. I've spent the week with JavaPoet and it's on the mind.

hzsweers:
nice :+1:

PR #628, opened 2018-08-19
Embed ProGuard rules in the jar

gabrielittner:
This should probably not be merged until AGP 3.2.0 is out, because the `-if` rule used for generated adapters is only supported in ProGuard 6.0.0+ (AGP 3.1.x uses ProGuard 5.x).

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,34 +578,14 @@ implementation 'com.squareup.moshi:moshi:1.6.0'
 Snapshots of the development version are available in [Sonatype's `snapshots` repository][snap].
 
 
-ProGuard
+R8 / ProGuard
 --------
 
-If you are using ProGuard you might need to add the following options:
-```
--dontwarn okio.**
--dontwarn javax.annotation.**
--keepclasseswithmembers class * {
-    @com.squareup.moshi.* <methods>;
-}
--keep @com.squareup.moshi.JsonQualifier interface *
-```
-Additional rules are needed if you are using Kotlin:
+If you are using R8 or ProGuard add the options from [this file](https://github.com/square/moshi/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro).
 
-*...If you are using the reflect API (i.e. `KotlinJsonAdapterFactory`):*
-```
--keepclassmembers class kotlin.Metadata {
-    public <methods>;
-}
-```
-*...If you are using the codegen API (i.e. `JsonClass(generateAdapter = true)`):*
-```
--keep class **JsonAdapter {
-    <init>(...);
-    <fields>;
-}
--keepnames @com.squareup.moshi.JsonClass class *
-```
+The `moshi-kotlin` artifact additionally requires the options from [this file](https://github.com/square/moshi/blob/master/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro)
+
+You might also need rules for Okio which is a dependency of this library.
 
 License
 --------
--- a/moshi/src/main/resources/META-INF/proguard/moshi.pro
+++ b/moshi/src/main/resources/META-INF/proguard/moshi.pro
@@ -7,10 +7,10 @@
 
 -keep @com.squareup.moshi.JsonQualifier interface *
 
-# The name of @JsonClass types is used to look up the generated adapter
+# The name of @JsonClass types is used to look up the generated adapter.
 -keepnames @com.squareup.moshi.JsonClass class *
 
-# Retain generated JsonAdapters if annotated type is retained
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -578,34 +578,14 @@ implementation 'com.squareup.moshi:moshi:1.6.0'
...
-```
--keep class **JsonAdapter {
-    <init>(...);
-    <fields>;
-}
--keepnames @com.squareup.moshi.JsonClass class *
-```
+The `moshi-kotlin` artifact additionally requires the options from [this file](https://github.com/square/moshi/blob/master/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro)
+
+You might also need rules for Okio which is a dependency of this library.
```
NightlyNexus: Can we bundle the rules in Okio's jar?

moshi/src/main/resources/META-INF/proguard/moshi.pro
```diff
@@ -0,0 +1,18 @@
...
+}
+
+-keep @com.squareup.moshi.JsonQualifier interface *
+
+# The name of @JsonClass types is used to look up the generated adapter
+-keepnames @com.squareup.moshi.JsonClass class *
+
+# Retain generated JsonAdapters if annotated type is retained
+-if @com.squareup.moshi.JsonClass class *
+-keep class <1>JsonAdapter {
```
NightlyNexus: lazyweb: what does the `<1>` mean?

moshi/src/main/resources/META-INF/proguard/moshi.pro
```diff
@@ -0,0 +1,18 @@
...
+-keepclasseswithmembers class * {
+    @com.squareup.moshi.* <methods>;
+}
+
+-keep @com.squareup.moshi.JsonQualifier interface *
+
+# The name of @JsonClass types is used to look up the generated adapter
+-keepnames @com.squareup.moshi.JsonClass class *
+
+# Retain generated JsonAdapters if annotated type is retained
```
NightlyNexus: nit: Add a period after line 10 and 13? Line 1's comment has a period.

README.md
```diff
@@ -578,34 +578,14 @@ implementation 'com.squareup.moshi:moshi:1.6.0'
...
-```
--dontwarn okio.**
--dontwarn javax.annotation.**
--keepclasseswithmembers class * {
-    @com.squareup.moshi.* <methods>;
-}
--keep @com.squareup.moshi.JsonQualifier interface *
-```
-Additional rules are needed if you are using Kotlin:
+If you are using R8 or ProGuard add the options from [this file](https://github.com/square/moshi/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro).
```
NightlyNexus: if you're using R8, you don't have to do anything, though, right? that's the benefit to bundling it into the jar.

moshi/src/main/resources/META-INF/proguard/moshi.pro
JakeWharton: It matches the first wildcard in the `if`. In this case it's the entire class name.

README.md
JakeWharton: Yes. Once we ship it I plan to change all the READMEs to say ProGuard only and that for R8 you do nothing.

README.md
JakeWharton: I think we did, didn't we? Or did I not get to it yet?

moshi/src/main/resources/META-INF/proguard/moshi.pro
NightlyNexus: Nice. Thanks.

moshi/src/main/resources/META-INF/proguard/moshi.pro
gabrielittner: done 24e0777

moshi/src/main/resources/META-INF/proguard/moshi.pro
```diff
@@ -0,0 +1,18 @@
...
+    @com.squareup.moshi.* <methods>;
+}
+
+-keep @com.squareup.moshi.JsonQualifier interface *
+
+# The name of @JsonClass types is used to look up the generated adapter.
+-keepnames @com.squareup.moshi.JsonClass class *
+
+# Retain generated JsonAdapters if annotated type is retained.
+-if @com.squareup.moshi.JsonClass class *
```
hzsweers: This is really neat

README.md
KonradSzewczuk: @JakeWharton Is it still planned to update the README's for R8 case?

PR #642, opened 2018-09-02
Switch to public kotlinx-metadata API

hzsweers:
> Continuation from #570

This swaps out kotlin-metadata for the new [kotlinx-metadata](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm) API

The high level architecture:
 - Switch to new `Data` class representations of specific types, replacing the raw protos they were reading before.
 - The `metadata.kt` class has been switched to provide APIs for resolving `TypeName`s from the library's `KmVisitor` APIs. This is a significant departure from before since it's a visitor pattern, but ended up working nicely once I wrote `TypeNameKmTypeVisitor`.
 - With the new APIs, we can actually write simple unit tests for various parts of the `metadata.kt` utilities and read class headers directly off the test class. This is great for debugging purposes and the general value of not always having to write integration tests. I started with some basic tests in `MetadataTest`. It's nowhere near as exhaustive as the integration tests, but it's a start and covers most of the common areas.
 - Some other misc little things

The other changes were pretty lightweight after migrating `metadata.kt`. Overall pretty happy with how this turned out, and definitely cleaner than reading the raw protos from before. The changeset is a bit big (particularly the first commit), but I don't think it would have been feasible to do this more incrementally in a way that still made sense.

Open bits:
- This is functional in Kotlin 1.2.60+, but `@Metadata` is not officially public until Kotlin 1.3
- kotlinx-metadata is not considered a public API yet and is also only available in the kotlinx jfrog repo. We could shade, but it uses a service loader mechanism to load its JVM support as well. Might take some work and my maven-fu is not good


hzsweers:
The failure is due to the shading. If we remove those lines, it works fine (and in general `mvn clean test` works fine). I suspect we can make shading work if we properly reshade the services file as well, but not sure if that's possible with the shade plugin

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/MetadataTest.kt
```diff
@@ -0,0 +1,187 @@
...
+
+  @Test
+  fun generics() {
+    val classData = Generics::class.loadClassData()
+
+    assertThat(classData.typeVariables).hasSize(3)
+    val tType = classData.typeVariables[0]
+    assertThat(tType.name).isEqualTo("T")
+    assertThat(tType.reified).isFalse()
+    assertThat(tType.variance).isNull() // we don't redeclare out variance
```
Egorand: Can you explain this? What does "redeclaring" mean in this context?

hzsweers: It's in reference to https://github.com/square/moshi/blob/78821bbc8041bbc1735aa3e1f41629c8d230ff7c/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt#L113

Open to wording suggestions :)

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+    typeAliasName = name
+  }
+
+  override fun visitTypeParameter(id: Int) {
+    typeParameter = getTypeParameter?.invoke(id) ?: throw IllegalStateException(
+        "Visiting TypeParameter when there are no type parameters!")
+  }
+
+  override fun visitEnd() {
+    var finalType = flexibleTypeUpperBound ?: outerType ?: typeParameter
```
swankjesse: I find this surprising . . . I assume it’ll always be exactly one of these and not multiple?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+  }
+}
+
+internal fun KotlinClassMetadata.Class.readClassData(): ClassData {
+  @Suppress("RedundantExplicitType")
+  var classFlags: Flags = 0
+  lateinit var className: String
+  var constructorData: ConstructorData? = null
+  var companionObjectName: String? = null
+  val typeParameters = LinkedHashMap<Int, TypeVariableName>()
```
swankjesse: mapOf ?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+        override fun visitEnd() {
+          val finalVariance = variance.asKModifier().let {
+            if (it == KModifier.OUT) {
+              // We don't redeclare out variance here
+              null
+            } else {
+              it
+            }
+          }
+          val typeVariableName = if (upperBoundList.isEmpty()) {
```
swankjesse: this doesn’t get populated with Any, does it? Would be nice if it did

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+  return ClassData(className.replace("/", "."),
+      classFlags,
+      companionObjectName,
+      constructorData,
+      superTypes,
+      typeParameters.values.toList(),
+      properties)
+}
+
+internal data class ClassData(
```
swankjesse: nit: do we have a more descriptive suffix than Data? Even Metadata seems more appropriate

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+    val name: String,
+    val flags: Flags,
+    val companionObjectName: String?,
+    val constructorData: ConstructorData?,
+    val superTypes: MutableList<TypeName>,
+    val typeVariables: List<TypeVariableName>,
+    val properties: List<PropertyData>
+) {
+  fun getPropertyOrNull(methodElement: ExecutableElement): PropertyData? {
```
swankjesse: is the `OrNull` implied by the return type?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+        ?.substringBefore(kotlinPropertyAnnotationsFunPostfix)
+        ?.let { propertyName -> properties.firstOrNull { propertyName == it.name } }
+  }
+
+  companion object {
+    /**
+     * Postfix of the method name containing the [kotlin.Metadata] annotation for the relative property.
+     * @see [getPropertyOrNull]
+     */
+    const val kotlinPropertyAnnotationsFunPostfix = "\$annotations"
```
swankjesse: CONST_CASE ?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
swankjesse: Maybe this should be called `KmClass` because it appears like that’s what we’re trying to recover from the awful visitor

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TargetType.kt
```diff
@@ -63,47 +48,43 @@ internal data class TargetType(
...
-          return null
-        }
-        proto.visibility == LOCAL -> {
-          messager.printMessage(
-              ERROR, "@JsonClass can't be applied to $element: must not be local", element)
-          return null
-        }
+      val typeMetadata = classMetadata as KotlinClassMetadata.Class
+      val classData = typeMetadata.readClassData()
+      if (Flag.Class.IS_ENUM_CLASS(classData.flags)) {
```
swankjesse: when ?

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/MetadataTest.kt
```diff
@@ -0,0 +1,187 @@
...
+    assertThat(fooParam.isVarArg).isFalse()
+    val barParam = classData.constructorData.parameters[1]
+    assertThat(barParam.name).isEqualTo("bar")
+    assertThat(barParam.type).isEqualTo(IntArray::class.asClassName())
+    assertThat(barParam.isVarArg).isTrue()
+    assertThat(barParam.varargElementType).isEqualTo(INT)
+  }
+
+  class ConstructorClass(val foo: String, vararg bar: Int) {
+    constructor(bar: Int) : this("defaultFoo") // Secondary constructors are ignored
```
swankjesse: nit: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/MetadataTest.kt
```diff
@@ -0,0 +1,187 @@
...
+  @Test
+  fun superType() {
+    val classData = SuperType::class.loadClassData()
+
+    assertThat(classData.superTypes).hasSize(1)
+    assertThat(classData.superTypes[0]).isEqualTo(BaseType::class.asClassName())
+  }
+
+  abstract class BaseType
+  class SuperType : BaseType()
```
swankjesse: nit: supertype is one word

swankjesse:
The visitor pattern makes everything more complicated here. Is there  a value objects API we could use instead? Would make this a lot better.

hzsweers:
I agree it makes things a bit more ceremonious, but I also suspect that's a bit intentional since they don't want this to be too tightly coupled to the implementation of using protos. Most of this work was sort of building a value objects API around it

kotlin/codegen/pom.xml
```diff
@@ -180,6 +190,11 @@
                   <pattern>com.squareup.kotlinpoet</pattern>
                   <shadedPattern>com.squareup.moshi.kotlinpoet</shadedPattern>
                 </relocation>
+                <!-- Repackage Kotlinx-metadata until its API is stable. -->
```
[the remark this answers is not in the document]
hzsweers: We do need to do something to wire resources correctly still since it uses an SPI. Maybe that's just as simple as manually writing one that points to the new one and excluding the original one we get in the dependency

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -49,9 +47,9 @@ internal class AdapterGenerator(
   private val propertyList: List<PropertyGenerator>
 ) {
   private val className = target.name
-  private val isDataClass = target.proto.isDataClass
+  private val isDataClass = Flag.Class.IS_DATA(target.data.flags)
   private val companionObjectName = target.companionObjectName
-  private val visibility = target.proto.visibility!!
+  private val isInternal = Flag.IS_INTERNAL(target.data.flags)
```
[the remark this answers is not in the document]
hzsweers: Agreed. The previous API was actually something @Takhion [built as a bunch of extension functions](https://github.com/Takhion/kotlin-metadata/blob/master/lib/src/main/kotlin/me/eugeniomarletti/kotlin/metadata/Flags.kt) routing to this, we could port that as part of this?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
```diff
@@ -63,6 +69,11 @@ class JsonClassCodegenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
 
   private val annotation = JsonClass::class.java
   private var generatedType: TypeElement? = null
+  private lateinit var messager: Messager
```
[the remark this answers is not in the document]
hzsweers: it is but the method cost is awkward. Also, `JsonClassCodeGenProcessor::messager.isInitialized` is weird

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+import kotlinx.metadata.KmTypeParameterVisitor
+import kotlinx.metadata.KmTypeVisitor
+import kotlinx.metadata.KmValueParameterVisitor
+import kotlinx.metadata.KmVariance
+import kotlinx.metadata.jvm.KotlinClassHeader
+import kotlinx.metadata.jvm.KotlinClassMetadata
+import javax.lang.model.element.Element
+import javax.lang.model.element.ExecutableElement
+
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```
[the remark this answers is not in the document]
hzsweers: Won't be necessary in kotlin 1.3+ fortunately. I learned about them because that's what the kotlinx-metadata sample does :)

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+}
+
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+internal fun Class<*>.readMetadata(): KotlinClassHeader? {
+  return getAnnotation(Metadata::class.java)?.asClassHeader()
+}
+
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+internal fun Metadata.asClassHeader(): KotlinClassHeader {
+  return KotlinClassHeader(k, mv, bv, d1, d2, xs, pn, xi)
```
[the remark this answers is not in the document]
hzsweers: Inclined to agree. @udalov penny for your thoughts on adding (something like) this to the main API?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  override fun visitTypeAlias(name: kotlinx.metadata.ClassName) {
+    if (!useTypeAlias) {
+      return
+    }
+    typeAliasName = name
+  }
+
+  override fun visitTypeParameter(id: Int) {
```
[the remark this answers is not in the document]
hzsweers: Nope, this is a single type and not a class/function definition. In this callback, it's signaling that this type reference should be coerced to a `TypeVariableName`

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+      return null
+    }
+    return TypeNameKmTypeVisitor(flags, getTypeParameter, useTypeAlias) {
+      typeAliasType = it
+    }
+  }
+
+  override fun visitArgument(flags: Flags, variance: KmVariance): KmTypeVisitor? {
+    return TypeNameKmTypeVisitor(flags, getTypeParameter, useTypeAlias) {
+      argumentList.add(
```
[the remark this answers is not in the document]
hzsweers: lgfe

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
```diff
@@ -0,0 +1,350 @@
...
+          when (variance) {
+            KmVariance.IN -> WildcardTypeName.supertypeOf(it)
+            KmVariance.OUT -> {
+              if (it == ANY) {
+                // This becomes a *, which we actually don't want here.
+                // List<Any> works with List<*>, but List<*> doesn't work with List<Any>
```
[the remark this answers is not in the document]
hzsweers: possibly. Writing code this way is something the IDE will yell at you for being redundant, but is technically legal.

From the smoke test type:

```kotlin
val any: List<*>,
val anyTwo: List<Any>,
val anyOut: List<out Any>,
```

in the generated adapter is handled as 

```kotlin
private val listOfAnyAdapter: JsonAdapter<List<*>> =
        moshi.adapter<List<*>>(Types.newParameterizedType(List::class.java, Types.subtypeOf(Any::class.java))).nonNull()

private val listOfAnyAdapter_: JsonAdapter<List<Any>> =
        moshi.adapter<List<Any>>(Types.newParameterizedType(List::class.java, Any::class.java)).nonNull()
```

Where `any` uses `listOfAnyAdapter`, and both `anyTwo` and `anyOut` use `listOfAnyAdapter_`. Kotlin sees `List<out Any>` and `List<Any>` as the same, and different than `List<*>`. The distinction seems to be that `*` to the compiler is actually equivalent to `Types.subtypeOf(Any::class.java)`, but KotlinPoet takes `Any` and converts it to a *.

So... maybe this is something we should fix in KotlinPoet, but it's super special cased. Basically it would say that `WildcardTypeName.subtypeOf(it)` should be `*` IFF `it` isn't `Any` with `out` projection.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
hzsweers: correct. This is probably weird logic to express that, but basically it's "figure out which one of these is present and use that"

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
hzsweers: it technically uses `LinkedHashMap` under the hood, but that's an implementation detail and we care about maintaining order for idempotence. This just makes that explicit. Open to suggestions though if you had something more idiomatic in mind!

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
hzsweers: as in to cover the `NULLABLE_ANY` default case?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/Metadata.kt
hzsweers: `KmClass` works for me!

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/MetadataTest.kt
hzsweers: 3e76596

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/MetadataTest.kt
hzsweers: 5da62eb

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TargetType.kt
hzsweers: f3721cf

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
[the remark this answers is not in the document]
hzsweers: Took a shot at this in 49d16cf

Takhion:
hey @hzsweers @swankjesse, I just saw this :)
I don't know if moving to `kotlinx.metadata` so soon was ideal, since the API is still quite rough around the edges, plus it's so far missing (and probably always will) the utilities to work with annotation processing. In addition, I spoke to @udalov recently and the current visitor pattern is only the lowest layer of the API, which is supposed to be used in environments where allocations must be avoided as much as possible, but will be complemented by a nice sealed-model-based API, or something like that!

swankjesse:
Thanks for the heads up @Takhion. I’d prefer to wait for a more user-friendly API.

@hzsweers ?

udalov:
FYI I've created an issue to track the progress on adding the aforementioned API to kotlinx.metadata: https://youtrack.jetbrains.com/issue/KT-26602. (I have to say though, we're quite busy at the moment with the upcoming release of Kotlin 1.3 and related activities, so it can take quite some time until fixed.)

swankjesse:
Okay, let's not wait for something beyond visitors.

swankjesse:
Hang on, we’re still stuck on these dependencies reaching Maven Central. We publish there and they don’t allow us to have a transitive dependency on other repos.

hzsweers:
If we shade it, does that cover that? I'm in no rush for this to ship before 1.3 (since that's when Metadata is formally public), but not sure if the library itself will be stable anytime soon

swankjesse:
Seems like a net loss from where we are today. ~300 net new lines of code plus a shaded dependency on an unstable API.

hzsweers:
That's fair, though technically the current API we use is also unstable (actually [broke](https://issuetracker.google.com/issues/78328708) in Kotlin 1.2.40 and required a new version to support it again).

Would the 300 lines feel like a mitigated cost if they were somewhere more reusable? I.e. a separate reusable library that other kotlin-aware-annotation-processors could use?

Egorand:
I can totally see this becoming a complementary annotation processor for KotlinPoet, how about we experiment with moving this under the KotlinPoet project and see where that goes?

swankjesse:
... I’d also like to consider promoting our compiler testing code to KotlinPoet.

hzsweers:
Closing this out for now, will open an issue in KotlinPoet to continue discussion later next week!

Issue #615, opened 2018-08-08
Moshi Codegen with Nothing and Retrofit


TimothyEarley:
Not quite sure where this issue belongs (here or retrofit).

Anyway, I ran into an interesting problem while using retrofit (2.4.0) and moshi kotlin codegen (1.6.0). I have a service which takes a parametric class as a `@Body` argument. My use case is [this](http://oneeyedmen.com/nothing-can-save-us.html).

```kotlin
@JsonClass(generateAdapter = true)
data class Bar<T>(val t: T)

interface Service {
    @POST("...") fun send(@Body Bar<String>): Call<Bar<String>>
}
```

The generated JsonAdapter captures this generic parameter no problem and retrofit also uses the correct generic type gathered [here](https://github.com/square/retrofit/blob/7158698314daa138e993fac6a590ed19d78a8599/retrofit/src/main/java/retrofit2/ServiceMethod.java#L160) (and on the master branch [here](https://github.com/square/retrofit/blob/c1b100f3c4e84711fa23bdfed03cb78c4d8a8f19/retrofit/src/main/java/retrofit2/RequestFactory.java#L145)).

The problem arises when I replace `String` with `Nothing`. If my analysis was correct this compiles to a raw type. This means that while the codegen still correctly sees the generic, retrofit fails to see the generic parameter at runtime (using `Method.getGenericParameterTypes()`).

This results in an exception while attempting to call the `send` method from above. The error is 
```
java.lang.IllegalArgumentException: Unable to create @Body converter for class Bar (parameter #1) for method Service.send
                     at retrofit2.ServiceMethod$Builder.methodError(ServiceMethod.java:755)
                      at retrofit2.ServiceMethod$Builder.parameterError(ServiceMethod.java:760)
                      at retrofit2.ServiceMethod$Builder.parseParameterAnnotation(ServiceMethod.java:716)
                      at retrofit2.ServiceMethod$Builder.parseParameter(ServiceMethod.java:339)
                      at retrofit2.ServiceMethod$Builder.build(ServiceMethod.java:207)
                      at retrofit2.Retrofit.loadServiceMethod(Retrofit.java:170)
                      at retrofit2.Retrofit$1.invoke(Retrofit.java:147)
                      at java.lang.reflect.Proxy.invoke(Proxy.java:913)

         [...]

Caused by: java.lang.RuntimeException: Failed to find the generated JsonAdapter constructor for class Bar
[...]

PR #651, opened 2018-09-09
Call Types.newParameterizedTypeWithOwner when necessary.

swankjesse:
Otherwise we crash with an exception attempting to create an adapter
for an enclosed type that has a type parameter.

I ran into this looking for a test case for issue 615.
https://github.com/square/moshi/issues/615

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
@@ -66,12 +66,21 @@ abstract class TypeRenderer {
               Types::class,
               render(typeName.typeArguments[0].objectType()))
         } else {
-          val placeholders = typeName.typeArguments.joinToString(", ") { "%L" }
-          CodeBlock.of(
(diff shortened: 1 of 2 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/TypeRenderer.kt
```diff
@@ -66,12 +66,21 @@ abstract class TypeRenderer {
               Types::class,
               render(typeName.typeArguments[0].objectType()))
         } else {
-          val placeholders = typeName.typeArguments.joinToString(", ") { "%L" }
-          CodeBlock.of(
-              "%T.newParameterizedType(%T::class.java, $placeholders)",
-              Types::class,
-              typeName.rawType.objectType(),
-              *(typeName.typeArguments.map { render(it.objectType()) }.toTypedArray()))
+          val builder = CodeBlock.builder().apply {
```
JakeWharton: We should add buildCodeBlock to KotlinPoet to match buildString and buildSpannedString

swankjesse: https://github.com/square/kotlinpoet/issues/468

PR #654, opened 2018-09-10
Use toString for mixingReflectionAndCodegen.

NightlyNexus:
In the previous test, the class name of the reflective adapter is `com.squareup.moshi.JsonAdapter$2` which seems brittle for the intended comparison.


NightlyNexus:
obsolete.

PR #657, opened 2018-09-11
Fix qualifiedAdaptersAreShared test.

NightlyNexus:
The test needs to inspect the original codegen adapter, not the adapter from the lookup (which is now the null-safe adapter).


NightlyNexus:
Closed by #663

PR #665, opened 2018-09-12
Fix RuntimeJsonAdapterFactory breaking with Object as the base type.

### LINKED ISSUE #664: Object adapter as base for RuntimeJsonAdapterFactory
```java
@Test public void objectAsBaseType() throws IOException {
      .add(RuntimeJsonAdapterFactory.of(Object.class, "type")
          .registerSubtype(Success.class, "success"))
      .build();
  JsonAdapter<Object> adapter = moshi.adapter(Object.class);

  assertThat(adapter.fromJson("{\"type\":\"success\",\"value\":\"Okay!\"}"))
      .isEqualTo(new Success("Okay!"));
  assertThat(adapter.toJson(new Success("Okay!")))
      .isEqualTo("{\"type\":\"success\",\"value\":\"Okay!\"}");
}
```

```
java.lang.IllegalArgumentException: Expected one of [class com.squareup.moshi.adapters.RuntimeJsonAdapterFactoryTest$Success] but found type, a class java.lang.String. Register this subtype.

	at com.squareup.moshi.adapters.RuntimeJsonAdapterFactory$RuntimeJsonAdapter.toJson(RuntimeJsonAdapterFactory.java:151)
	at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
	at com.squareup.moshi.Moshi$DeferredAdapter.toJson(Moshi.java:273)
	at com.squareup.moshi.MapJsonAdapter.toJson(MapJsonAdapter.java:57)
	at com.squareup.moshi.MapJsonAdapter.toJson(MapJsonAdapter.java:30)
	at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
	at com.squareup.moshi.StandardJsonAdapters$ObjectJsonAdapter.toJson(StandardJsonAdapters.java:366)
	at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
	at com.squareup.moshi.adapters.RuntimeJsonAdapterFactory$RuntimeJsonAdapter.toJson(RuntimeJsonAdapterFactory.java:160)
	at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
	at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:52)
	at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:58)
	at com.squareup.moshi.adapters.RuntimeJsonAdapterFactoryTest.objectAsBaseType(RuntimeJsonAdapterFactoryTest.java:183)
```

breaks on master when encoding with the usage of the object adapter. This is because the map (Map<String, Object>) adapter used by the delegate, default Object adapter will try to use this runtime adapter instead of the default Object adapter.

This runtime adapter should detect reentry and delegate to the object adapter.

NightlyNexus:
Resolves #664

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
@@ -21,7 +21,6 @@ import com.squareup.moshi.JsonReader;
 import com.squareup.moshi.JsonWriter;
 import com.squareup.moshi.Moshi;
 import com.squareup.moshi.Types;
-import com.squareup.moshi.internal.Util;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
@@ -41,7 +40,7 @@ public final class RuntimeJsonAdapterFactory<T> implements JsonAdapter.Factory {
   final Map<String, Type> labelToType = new LinkedHashMap<>();
 
   /**
-   * @param baseType The base type for which this factory will create adapters.
+   * @param baseType The base type for which this factory will create adapters. Cannot be Object.
    * @param labelKey The key in the JSON object whose value determines the type to which to map the
    *     JSON object.
    */
@@ -49,6 +48,10 @@ public final class RuntimeJsonAdapterFactory<T> implements JsonAdapter.Factory {
   public static <T> RuntimeJsonAdapterFactory<T> of(Class<T> baseType, String labelKey) {
     if (baseType == null) throw new NullPointerException("baseType == null");
     if (labelKey == null) throw new NullPointerException("labelKey == null");
+    if (baseType == Object.class) {
+      throw new IllegalArgumentException(
+          "The base type must not be Object. Consider using a marker interface.");
+    }
     return new RuntimeJsonAdapterFactory<>(baseType, labelKey);
   }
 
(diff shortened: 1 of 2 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -140,6 +150,25 @@
     }
 
     @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      if (!baseTypeIsObject) {
+        objectJsonAdapter.toJson(writer, toJsonMap(value));
+        return;
+      }
+      if (reentrantFromObjectAdapter) {
```
swankjesse: I don't think this works for recursive structures like trees.

swankjesse: Also this isn't threadsafe.

swankjesse: Curious: how does it fail without this?

NightlyNexus: ahaha, yeah, of course.

>how does it fail without this?

the test in #664 fails on master. is that what you mean?

NightlyNexus:
For encoding, we could not use the object adapter and manually do:
```java
static JsonWriter writeJsonValue(JsonWriter writer, Object jsonValue) throws IOException {
  if (jsonValue == null) {
    return writer.nullValue();
  }
  if (jsonValue instanceof Number) {
    return writer.value((Number) jsonValue);
  }
  if (jsonValue instanceof String) {
    return writer.value((String) jsonValue);
  }
  if (jsonValue instanceof Boolean) {
    return writer.value((Boolean) jsonValue);
  }
  if (jsonValue instanceof Map) {
    writer.beginObject();
    for (Map.Entry<String, Object> entry : ((Map<String, Object>) jsonValue).entrySet()) {
      writer.name(entry.getKey());
      writeJsonValue(writer, entry.getValue());
    }
    return writer.endObject();
  }
  if (jsonValue instanceof List) {
    writer.beginArray();
    List<Object> list = (List<Object>) jsonValue;
    for (int i = 0, size = list.size(); i < size; i++) {
      writeJsonValue(writer, list.get(i));
    }
    return writer.endArray();
  }
  throw new IllegalArgumentException(
      "Expected a JSON value but found " + jsonValue + ", a " + jsonValue.getClass());
}
```

However, there's no clear path forward for decoding for #661 with Object as the base type.

How about not allowing Object to be the base type?

swankjesse:
This is a simple, pragmatic fix. Good call.

PR #668, opened 2018-09-13
Fix JsonQualifiers test

oldergod:
Their seems to have been concurrency between #616 and #659

### DIFF
--- a/moshi/src/test/java/com/squareup/moshi/JsonQualifiersTest.java
+++ b/moshi/src/test/java/com/squareup/moshi/JsonQualifiersTest.java
@@ -339,8 +339,8 @@ public final class JsonQualifiersTest {
       assertThat(expected).hasCauseExactlyInstanceOf(IllegalArgumentException.class);
       assertThat(expected.getCause()).hasMessage("No @FromJson adapter for class java.lang.String "
           + "annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]");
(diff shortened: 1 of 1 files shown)


moshi/src/test/java/com/squareup/moshi/JsonQualifiersTest.java
```diff
@@ -339,9 +339,6 @@
       assertThat(expected).hasCauseExactlyInstanceOf(IllegalArgumentException.class);
       assertThat(expected.getCause()).hasMessage("No @FromJson adapter for class java.lang.String "
           + "annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]");
-      assertThat(expected).hasCauseExactlyInstanceOf(IllegalArgumentException.class);
-      assertThat(expected.getCause()).hasMessage("No next JsonAdapter for class "
-          + "java.lang.String annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]");
```
NightlyNexus: i think we still want the tests from #659

the two statement above should be
```java
assertThat(expected.getCause()).hasCauseExactlyInstanceOf(IllegalArgumentException.class);
assertThat(expected.getCause().getCause()).hasMessage("No next JsonAdapter for class "
    + "java.lang.String annotated [@com.squareup.moshi.JsonQualifiersTest$FooPrefix()]");
```

oldergod: I see, cool. fixed it so.

PR #674, opened 2018-09-19
Use the adapter lookup hint API in Kotlin codegen adapters.

NightlyNexus:


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -133,9 +133,27 @@ internal class AdapterGenerator(
     }
 
     result.addProperty(optionsProperty)
-    for (uniqueAdapter in propertyList.distinctBy { it.delegateKey }) {
+
+    val propertyListSize = propertyList.size
+    val delegateKeyCount = LinkedHashMap<DelegateKey, Int>(propertyListSize)
+    val uniqueAdapters = ArrayList<PropertyGenerator>(propertyListSize)
```
JakeWharton:
```kotlin
val uniqueAdapters = propertyList.distinctBy { it.key }
```

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -133,9 +133,27 @@ internal class AdapterGenerator(
     }
 
     result.addProperty(optionsProperty)
-    for (uniqueAdapter in propertyList.distinctBy { it.delegateKey }) {
+
+    val propertyListSize = propertyList.size
+    val delegateKeyCount = LinkedHashMap<DelegateKey, Int>(propertyListSize)
```
JakeWharton:
```kotlin
val delegateKeyCount = propertyList.groupingBy { it.key }.eachCount()
```

NightlyNexus: ohoho, that's what i was looking for. thank you.

hzsweers:
Agreed with Jake - what's this fixing/improving?

NightlyNexus:
```kotlin
class HasPlatformType(val strings: ArrayList<String>)

val moshi = Moshi.Builder().build()
val adapter = moshi.adapter(HasPlatformType::class.java)
```

before:
>java.lang.IllegalArgumentException: Platform java.util.ArrayList<java.lang.String> (with no annotations) requires explicit JsonAdapter to be registered
</br>Caused by: java.lang.IllegalArgumentException: Platform java.util.ArrayList<java.lang.String> (with no annotations) requires explicit JsonAdapter to be registered

after:
>java.lang.IllegalArgumentException: Platform java.util.ArrayList<java.lang.String> (with no annotations) requires explicit JsonAdapter to be registered
for java.util.ArrayList<java.lang.String> strings
for class com.squareup.moshi.kotlin.codgen.GeneratedAdaptersTest$HasPlatformType
</br>Caused by: java.lang.IllegalArgumentException: Platform java.util.ArrayList<java.lang.String> (with no annotations) requires explicit JsonAdapter to be registered

happy to add the test case if it is helpful.

JakeWharton:
My comment is because I know nothing about this module–not that I have any opinion on whether it was needed.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -133,9 +133,16 @@ internal class AdapterGenerator(
     }
 
     result.addProperty(optionsProperty)
+
+    val delegateKeyCount = propertyList.groupingBy { it.delegateKey }.eachCount()
+
     for (uniqueAdapter in propertyList.distinctBy { it.delegateKey }) {
```
swankjesse: since you’re doing groupingBy and distinctBy in two adjacent calls, could you only do groupingBy ? As-is it’s doing 2x as much work and it’s more awkward to follow

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -133,9 +133,16 @@ internal class AdapterGenerator(
...
+      // If the adapter is shared for multiple properties,
+      // don't provide a the fieldName hint to the adapter lookup.
```
NightlyNexus: looking at this again, i realize this is wrong. it might as well have a hint.
if you have
```
class HasPlatformType(val strings: ArrayList<String>, val list: ArrayList<String>)
```
the error message might as well tell you that you have an error in `strings` instead of not saying anything.

Issue #672, opened 2018-09-18
Impact when using readJsonValue


warabei14:
Hi!
You've implemented `RuntimeJsonAdapterFactory`
I am writing analogue of this for case if polymorphic json looks like that:
```{
"type" : "type1",
"value" : { /*some object details here*/}
}
I have two variants in this case:
1. Implementing `fromJson` like that:
```
val jsonValue = reader.readJsonValue()
            val jsonObject = jsonValue as Map<String, Any>?
            val resultLabel = jsonObject!![labelKey] as String
            val delegate = labelToAdapter[resultLabel]
            val obj = delegate!!.fromJsonValue(jsonObject["value"])
            return obj
```
And in this case simple 10 mb json with array of such objects will be parsed in about 350 ms
And if i am implementing like this:
```
            reader.readJsonValue()
            reader.beginObject()
            reader.nextName() //asserttype
            val resultLabel = reader.nextString()
            reader.nextName() //value - if value
            val delegate = labelToAdapter[resultLabel]
            val obj = delegate!!.fromJson(reader)
            reader.endObject()
            return obj
```
This gives me around 170 ms.

Possibility to read all the json at once especially in _top of json hierarchy_ can lead to decrease of time parse up to 2 times (and increased memory footprint ofc). 
So, if i use `RuntimeJsonAdapterFactory` my parsing speed is slowing down. 
In my case it is pretty simple to do such a thing (i omitted case when fields order can be random btw).

Why i am asking, looks like for `RuntimeJsonAdapterFactory` in most cases there we are *not need to read all the json as object*, and simply go through in "flat" mode (do not go deeply inside a structure, just see the "top level"). Then looks like it would be more efficient to go through reader in "not markable" mode and then use `fromJson` with `reader`(move peek to start of method).
As far as i understand it can be tricky or just not in nature of json parsers, but I hope i showed a good point for optimisation.
Maybe there any workarounds? Maybe okio can help?

ps:
As far as i know the same is the issue for gson (we have 100kb request with nesting polymorphic objects, that can be parsed up to 800ms)

NightlyNexus:
>10 mb json

huge. How many Java objects are getting allocated in `readJsonValue()` then? I'm guessing that's what's taking so much time. Also, all the names in the maps have to be decoded and have strings allocated in the readJsonValue case.
Typically, it's not a big deal to create a bunch of extra intermediate Java objects, but that's a big structure.

warabei14:
[...]

PR #680, opened 2018-09-24
Implement peekJson() for JsonValueReader

swankjesse:
Related to https://github.com/square/moshi/issues/672

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -180,10 +180,10 @@ public abstract class JsonReader implements Closeable {
   // The nesting stack. Using a manual array rather than an ArrayList saves 20%. This stack will
   // grow itself up to 256 levels of nesting including the top-level document. Deeper nesting is
   // prone to trigger StackOverflowErrors.
-  int stackSize = 0;
-  int[] scopes = new int[32];
-  String[] pathNames = new String[32];
-  int[] pathIndices = new int[32];
+  int stackSize;
+  int[] scopes;
+  String[] pathNames;
+  int[] pathIndices;
 
   /** True to accept non-spec compliant JSON. */
   boolean lenient;
@@ -196,8 +196,21 @@ public abstract class JsonReader implements Closeable {
     return new JsonUtf8Reader(source);
   }
 
+  // Package-private to control subclasses.
   JsonReader() {
-    // Package-private to control subclasses.
+    scopes = new int[32];
+    pathNames = new String[32];
+    pathIndices = new int[32];
+  }
+
+  // Package-private to control subclasses.
+  JsonReader(JsonReader copyFrom) {
+    this.stackSize = copyFrom.stackSize;
+    this.scopes = copyFrom.scopes.clone();
+    this.pathNames = copyFrom.pathNames.clone();
+    this.pathIndices = copyFrom.pathIndices.clone();
+    this.lenient = copyFrom.lenient;
+    this.failOnUnknown = copyFrom.failOnUnknown;
   }
 
   final void pushScope(int newTop) {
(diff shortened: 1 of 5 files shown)


moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -49,17 +50,31 @@
...
+    stack = new Object[32];
     stack[stackSize++] = root;
   }
 
+  /** Copy-constructor makes a deep copy for peeking. */
+  JsonValueReader(JsonValueReader copyFrom) {
+    super(copyFrom);
+
+    stack = copyFrom.stack.clone();
+    for (int i = 0; i < stackSize; i++) {
```
NightlyNexus: style nit: use enhanced for loop for the array?

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -374,4 +393,32 @@ private void remove() {
...
+    @Override public Object next() {
+      return list[next++];
+    }
+
+    @Override public void remove() {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override protected JsonIterator clone() {
+      return new JsonIterator(endToken, list, next);
```
NightlyNexus: worth a comment to state that the `list` doesn't need a clone?

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -985,4 +985,98 @@ JsonReader newReader(String json) throws IOException {
...
+    reader.endObject();
+    assertThat(reader.peek()).isEqualTo(JsonReader.Token.END_DOCUMENT);
+  }
+
+  @Test public void peekJsonReader() throws IOException {
+    JsonReader reader = newReader("[12,34,{\"a\":56,\"b\":78},90]");
+    assumeTrue(reader instanceof JsonValueReader); // Not implemented for JsonUtf8Reader yet!
+    for (int i = 0; i < 12; i++) {
+      readPeek12Steps(reader.peekJson(), i, 12);
+      readPeek12Steps(reader, i, i + 1);
```
NightlyNexus: this _looks_ like the from and until indices are unused?
i don't follow. what's this test case behavior?

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -985,4 +985,98 @@ JsonReader newReader(String json) throws IOException {
...
+    }
+  }
+
+  /**
+   * Read a fragment of {@code reader}. This assumes the fixed document defined in {@link
+   * #peekJsonReader} and reads a range of it on each call.
+   */
+  private void readPeek12Steps(JsonReader reader, int from, int until) throws IOException {
```
NightlyNexus: is this simply `if (from == until) return;`?

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -461,6 +470,32 @@ public final void setFailOnUnknown(boolean failOnUnknown) {
     }
   }
 
+  /**
+   * Returns a new {@code JsonReader} that can read data from this {@code JsonReader} without
```
NightlyNexus: nice! will we need a note to say that behavior is undefined if the original gets ahead of the peeked reader?

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -200,6 +200,15 @@
     // Package-private to control subclasses.
```
NightlyNexus: nit: should we move the other assignments into this constructor like JsonValueReader's `stack` did?

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
swankjesse: it’s slightly different semantics ’cause the loop doesn’t span the physical storage!

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
swankjesse: Renamed to `array`!

swankjesse: Fixed.

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
swankjesse: The real reader uses 12 operations to read a JSON document. Before each of the real reader’s operations we create a peeking reader and let it read the rest of the document.

Each `.` in here is a single read operation.
 
```
PEEK ............
REAL .           
PEEK  ...........
REAL  .          
PEEK   ..........
REAL   .         
PEEK    .........
REAL    .        
PEEK     ........
REAL     .       
PEEK      .......
REAL      .      
PEEK       ......
REAL       .     
PEEK        .....
REAL        .    
PEEK         ....
REAL         .   
PEEK          ...
REAL          .  
PEEK           ..
REAL           . 
PEEK            .
REAL            .
```

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
swankjesse: Not exactly.

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
NightlyNexus: that's what i thought it was supposed to be, and i read this at least two dozen times looking for a loop until i finally realized i was glazing over the fall-throughs. switch statements get me every time!

fantastic! 👍

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
NightlyNexus: forgot switch statements today. 👍

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
NightlyNexus: whoops, yes, ofc.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
NightlyNexus: yeah, mirroring with Okio is perfect!

PR #681, opened 2018-09-24
add [openjdk11] to Travis build matrix

sullis:


swankjesse:
Looks like we have a problem already.
```
[INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ moshi ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 27 source files to /home/travis/build/square/moshi/moshi/target/classes
/home/travis/build/square/moshi/moshi/src/main/java/com/squareup/moshi/JsonEncodingException.java:18: error: cannot access IOException
import java.io.IOException;
              ^
  bad class file: /modules/java.base/java/io/IOException.class
    class file has wrong version 55.0, should be 53.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

ZacSweers:
Closing this out for now, I suspect it's a problem with our maven setup and will be easier to fix when we build with gradle

PR #686, opened 2018-09-25
Mention proguard 6 requirement

hzsweers:


gabrielittner:
I would rather mention AGP 3.2.0 as requirement. Most probably don't know which proguard version they use and just have the one AGP depends on.

hzsweers:
Will update!

PR #687, opened 2018-09-25
Completely remove companion object jsonAdapter extension function gen

hzsweers:
It's broken currently (see #611), and after talking with @swankjesse decided it's best to just nix this API


NightlyNexus:
GH: "1 useless comment was hidden"
my review approval lol. ouch, GH.

Issue #689, opened 2018-09-25
Obfuscation and enums


gabrielittner:
The integrated `EnumJsonAdapter` in `StandardJsonAdapters` can't be instantiated because the field name doesn't match the name anymore.

My solution is to annotate enums that are used with Moshi with `@JsonClass(generateAdapter = false)` and then keep the field names for annotated enums:

```
-keepclassmembernames @com.squareup.moshi.JsonClass class * extends java.lang.Enum {
    <fields>;
}
```
This way it's possible to avoid keeping all enums.

Do you think it would make sense to add it to the shipped Proguard config?

swankjesse:
Yeah, I like that.

ezamelczykespeo:
I don't understand why `@Json(name = "name)` isn't sufficient. Enums work with Gsons `@SerializedName`

realdadfish:
@ezamelczykespeo Issue probably resides in this line:

https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java#L240

The Java method `getEnumConstants()` still gives back the literal enum names, not the obfuscated ones, so initializing Moshi with any enum adapter will crash with an `AssertionError` that wraps a `NoSuchFieldException` stemming from the failed reflection call.

VarunBarad:
@gabrielittner you just saved my day. 🙇

PR #693, opened 2018-09-26
Always keep fields of annotated enums

technoir42:
We need to keep enum fields for `EnumJsonAdapter` even if they are not used elsewhere.

```kotlin
@JsonClass(generateAdapter = false)
enum class StatusCode {
    @Json(name = "success")
    Success,
    @Json(name = "error")
    Error
}

if (status == StatusCode.Success) {
    ...
}
```

`-keepclassmembernames` won't prevent `StatusCode.Error` from being removed that's why `-keepclassmembers` should be used instead.
@gabrielittner

### DIFF
--- a/moshi/src/main/resources/META-INF/proguard/moshi.pro
+++ b/moshi/src/main/resources/META-INF/proguard/moshi.pro
@@ -9,7 +9,7 @@
 
 # Enum field names are used by the integrated EnumJsonAdapter.
 # Annotate enums with @JsonClass(generateAdapter = false) to use them with Moshi.
--keepclassmembernames @com.squareup.moshi.JsonClass class * extends java.lang.Enum {
+-keepclassmembers @com.squareup.moshi.JsonClass class * extends java.lang.Enum {
     <fields>;
 }
 


hzsweers:
errrr I don't think you should be annotating enums with `@JsonClass`, you should just keep enums (or just the ones you specifically need) instead. I'm not sure the generated adapter for this would even work, let alone be something that Moshi would look for in its adapter resolution.

technoir42:
@hzsweers ProGuard rule for enums annotated with `@JsonClass` was added a day ago. See #689 and #691 for context.

hzsweers:
I've been on vacation, so I missed that :). Just noticed the `= false` bit as well, so in that case I think this looks good :+1:

gabrielittner:
LGTM.

I didn't run into this but I think that's just because my enums are used in exhaustive when statements so all fields are used.

adams-y-chen:
My enum class name is still obfuscated in the apk after adding the `@JsonClass(generateAdapter = false)`. I'm using moshi 1.12.0.
Do I have to add proguard rule to my app project in addition to annotation my enum class?

```
@JsonClass(generateAdapter = false)
enum class StatusCode {
    @Json(name = "success")
    Success,
    @Json(name = "error")
    Error
}
```

PR #697, opened 2018-09-27
Make RuntimeJsonAdapterFactory public

jocmp:
RuntimeJsonAdapterFactory was public in a previous 1.7-SNAPSHOT build.

After viewing the todo, I changed the visibility modifier ahead of the next 1.8 snapshot release.


swankjesse:
I wanna stall this until the changes we’re planning for it are complete.

JoaquimLey:
Hello moshiers,

Since this being withheld is there a snapshot we can import in order to get access to this class? 
Thanks 🙏

PR #704, opened 2018-10-07
Implement peekJson() for JsonUtf8Reader

swankjesse:
This required an upgrade to Okio 2.1.0.


pom.xml
```diff
@@ -35,7 +35,7 @@
     <maven-assembly.version>3.1.0</maven-assembly.version>
 
     <!-- Dependencies -->
-    <okio.version>1.15.0</okio.version>
+    <okio.version>2.1.0</okio.version>
```
swankjesse: might hold this on 1.16
https://github.com/square/okio/pull/520

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -98,6 +98,27 @@
...
+    this.buffer = sourcePeek.getBuffer();
+    this.peeked = copyFrom.peeked;
+    this.peekedLong = copyFrom.peekedLong;
+    this.peekedNumberLength = copyFrom.peekedNumberLength;
+    this.peekedString = copyFrom.peekedString;
+
+    // Make sure our buffer has as many bytes as the source's buffer. This is necessary because
+    // JsonUtf8Reader assumes any data it has peeked (like the peekedNumberLength) are buffered.
+    try {
+      this.source.require(copyFrom.buffer.size());
```
JakeWharton: `this.source` -> `sourcePeek` since it's already a local

pom.xml
JakeWharton: Meh

JakeWharton: Picasso was going in on Kotlin for peek. Even if backported I'm not sure we'll change there. Moshi also feels like it would be fine going Kotlin. OkHttp and Retrofit should probably wait a year.

swankjesse: In my day job I’m to get a bunch of different dependencies to play nicely together: Guava, gRPC, Vitess, and the Google Cloud Datastore. The experience has left me pretty grumpy!

I’m trying to give our users a little flexibility.

That said, I do really want to do a Moshi 2.0 that’s the same in spirit and execution to Okio 2.0. That should be great!

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
swankjesse: Fixed.

PR #707, opened 2018-10-10
Change RuntimeJsonAdapterFactory to peek for type names.

swankjesse:
This is a bit awkward because JsonReader.Options doesn't tell you
what its values are.

Also awkward because we don't yet have an equivalent to stream
the encode of the value.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
@@ -24,7 +24,10 @@ import com.squareup.moshi.Types;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.annotation.CheckReturnValue;
@@ -39,8 +42,8 @@ import javax.annotation.CheckReturnValue;
  *
  *   Moshi moshi = new Moshi.Builder()
  *       .add(RuntimeJsonAdapterFactory.of(Message.class, "type")
- *           .registerSubtype(Success.class, "success")
- *           .registerSubtype(Error.class, "error"))
+ *           .withSubtype(Success.class, "success")
+ *           .withSubtype(Error.class, "error"))
  *       .build();
  * }</pre>
  */
@@ -48,7 +51,16 @@ import javax.annotation.CheckReturnValue;
 final class RuntimeJsonAdapterFactory<T> implements JsonAdapter.Factory {
   final Class<T> baseType;
   final String labelKey;
-  final Map<String, Type> labelToType = new LinkedHashMap<>();
+  final List<String> labels;
+  final List<Type> subtypes;
+
+  RuntimeJsonAdapterFactory(
+      Class<T> baseType, String labelKey, List<String> labels, List<Type> subtypes) {
+    this.baseType = baseType;
+    this.labelKey = labelKey;
+    this.labels = labels;
+    this.subtypes = subtypes;
+  }
 
   /**
    * @param baseType The base type for which this factory will create adapters. Cannot be Object.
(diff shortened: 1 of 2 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -63,112 +75,117 @@
...
   }
 
   /**
-   * Register the subtype that can be created based on the label. When an unknown type is found
+   * Returns a new factory that decodes instances of {@code subtype}. When an unknown type is found
    * during encoding an {@linkplain IllegalArgumentException} will be thrown. When an unknown label
    * is found during decoding a {@linkplain JsonDataException} will be thrown.
    */
-  public RuntimeJsonAdapterFactory<T> registerSubtype(Class<? extends T> subtype, String label) {
+  public RuntimeJsonAdapterFactory<T> withSubtype(Class<? extends T> subtype, String label) {
```
NightlyNexus: nice.

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -39,16 +42,25 @@
...
  *       .add(RuntimeJsonAdapterFactory.of(Message.class, "type")
- *           .registerSubtype(Success.class, "success")
- *           .registerSubtype(Error.class, "error"))
+ *           .withSubtype(Success.class, "success")
+ *           .withSubtype(Error.class, "error"))
  *       .build();
  * }</pre>
  */
 // TODO(jwilson): make this class public in Moshi 1.8.
 final class RuntimeJsonAdapterFactory<T> implements JsonAdapter.Factory {
```
NightlyNexus: public in a followup?

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -63,112 +75,117 @@
...
+    if (labels.contains(label) || subtypes.contains(subtype)) {
       throw new IllegalArgumentException("Subtypes and labels must be unique.");
     }
-    labelToType.put(label, subtype);
-    return this;
+    List<String> newLabels = new ArrayList<>(labels);
+    newLabels.add(label);
+    List<Type> newSubtypes = new ArrayList<>(subtypes);
+    newSubtypes.add(subtype);
+    return new RuntimeJsonAdapterFactory<>(baseType, labelKey, newLabels, newSubtypes);
```
NightlyNexus: 👍

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -63,112 +75,117 @@
...
-    int size = labelToType.size();
-    Map<String, JsonAdapter<Object>> labelToAdapter = new LinkedHashMap<>(size);
-    Map<Type, String> typeToLabel = new LinkedHashMap<>(size);
-    for (Map.Entry<String, Type> entry : labelToType.entrySet()) {
-      String label = entry.getKey();
-      Type typeValue = entry.getValue();
-      typeToLabel.put(typeValue, label);
-      labelToAdapter.put(label, moshi.adapter(typeValue));
+
+    List<JsonAdapter<Object>> jsonAdapters = new ArrayList<>();
```
NightlyNexus: tiny nit: could be `new ArrayList<>(size)`

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
```diff
@@ -63,112 +75,117 @@
...
+        if (labelIndex == -1) {
+          throw new JsonDataException("Expected one of "
+              + labels
+              + " for key '"
+              + labelKey
+              + "' but found '"
+              + reader.nextString()
+              + "'. Register a subtype for this label.");
+        }
+        reader.close();
```
NightlyNexus: not really necessary, right?

adapters/src/test/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactoryTest.java
```diff
@@ -101,25 +101,24 @@
...
         .build();
     JsonAdapter<Message> adapter = moshi.adapter(Message.class);
 
     try {
       adapter.fromJson("{\"type\":{},\"value\":\"Okay!\"}");
       fail();
     } catch (JsonDataException expected) {
-      assertThat(expected).hasMessage("Label for 'type' must be a string but was {},"
-          + " a class com.squareup.moshi.LinkedHashTreeMap");
+      assertThat(expected).hasMessage("Expected a string but was BEGIN_OBJECT at path $.type");
```
NightlyNexus: yep!

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
swankjesse: Yeah, I still wanna find a way to stream out the type itself.

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
swankjesse: Can fix in follow-up!

adapters/src/main/java/com/squareup/moshi/adapters/RuntimeJsonAdapterFactory.java
swankjesse: I wanna fix PeekingSource so that it clears buffers on close. That could potentially allow us to save some GC load.

NightlyNexus: cool 👍

PR #708, opened 2018-10-14
JSON flattening.

swankjesse:
This is a new API that makes it possible to do more interesting things
with composition. It's currently write-only; doing composition on reads
is much more difficult.


moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -357,6 +379,65 @@ final void promoteValueToName() throws IOException {
...
+   *   }
+   * }</pre>
+   *
+   * <p>This method supports flattening arrays within arrays and objects within objects. Other
+   * combinations are allowed but do not participate in flattening.
+   *
+   * <p>This method returns an opaque token. Callers must match all calls to this method with a call
+   * to {@link #endFlatten} with the matching token.
+   */
+  public final int beginFlatten() {
```
swankjesse: An alternative design might return an implementation of JsonWriter that delegates everything but the nesting calls. I could be convinced that’s the way to go, though it’ll be a bit awkward as the base class currently has a lot of state.

NightlyNexus: good thought. not sure, but I'm leaning towards what's here because it implies advanced usage.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -357,6 +379,65 @@ final void promoteValueToName() throws IOException {
...
+   *     int token = writer.beginFlatten();
+   *     writer.value(1);
+   *     integersAdapter.toJson(writer, Arrays.asList(2, 3, 4));
+   *     writer.value(5);
+   *     writer.endFlatten(token);
+   *     writer.endArray();
+   *   }
+   * }</pre>
+   *
+   * <p>This method supports flattening arrays within arrays and objects within objects. Other
```
JakeWharton: I had to read these sentences a few times. I don't think I like the words "supports" or "allowed" because I have to infer the behavior a bit. I would prefer them to be a lot more definite, especially the second sentence.

swankjesse: Yeah, good feedback.

I wrote more docs with examples.

PR #710, opened 2018-10-14
Retain generated JsonAdapters for nested classes

### LINKED ISSUE #709: R8 removes generated JsonAdapters for nested classes
This rule doesn't work for nested classes since their adapters are named like `OuterClass_NestedClassJsonAdapter` but the rule expects `OuterClass$NestedClassJsonAdapter`.
```
-if @com.squareup.moshi.JsonClass class *
-keep class <1>JsonAdapter {
    <init>(...);
    <fields>;
}
```

technoir42:
Closes #709.


hzsweers:
This is clever, though I wish we could somehow add a test for this. Mind adding a comment above that linking the issue for posterity?

Also - what happens with nesting in more than one level?

technoir42:
This will work only for 1 level of nesting. I guess we'll need to either add multiple rules for several levels of nesting or change the way adapters are generated for nested classes so it'd be easier to write a generic rule.

hzsweers:
Let me ask proguard/R8 contacts in case there's a way, but I suppose worst case we could add something like the first 95overloads or something and say any (pathological) nestings deeper than that should be handled manually by the developers writing them?

hzsweers:
Sorry for the delay on this. Talked to the folks and the best solution is to just add overloads and leave the rest to users. Let's go with say, 5 nesting levels?

swankjesse:
@technoir42 thumbs up to 5 nesting levels.

technoir42:
I've added rules for more nesting levels.

hzsweers:
Thanks!

Issue #679, opened 2018-09-22
IllegalStateException: Type adapter isn't ready


Tolriq:
Got some very very very rare exception (see below).

Any idea of the cause and possible workaround ?
```
Fatal Exception: java.lang.IllegalStateException: Type adapter isn't ready
       at com.squareup.moshi.Moshi$DeferredAdapter.toJson(Moshi.java:272)
       at com.squareup.moshi.MapJsonAdapter.toJson(MapJsonAdapter.java:1057)
       at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
       at com.squareup.moshi.StandardJsonAdapters$ObjectJsonAdapter.toJson(StandardJsonAdapters.java:367)
       at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
       at com.squareup.moshi.ClassJsonAdapter.com.squareup.moshi.ClassJsonAdapter$FieldBinding.write(ClassJsonAdapter.java:1203)
       at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
       at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:52)
       at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:58)
       at com.mediacenters.kodi.api.KodiApi$JsonRpcCall.getBody(KodiApi.kt:71)
       at com.mediacenters.kodi.api.KodiApiConnection$asynchronousCall$1$$special$$inlined$forEach$lambda$1.invoke(KodiApiConnection.kt:1069)
       at kotlin.text.StringsKt__StringBuilderKt.appendElement(StringBuilder.kt:58)
       at kotlin.collections.CollectionsKt___CollectionsKt.sortedWith(_Collections.kt:2195)
       at com.mediacenters.kodi.api.KodiApiConnection$asynchronousCall$1.kotlin.collections.CollectionsKt___CollectionsKt.joinTo$default$1296cf06(KodiApiConnection.kt:5189)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
       at java.lang.Thread.run(Thread.java:764)
```

NightlyNexus:
Post the code that caused the exception?

Tolriq:
Ultra basic stuff.

```
[...]

PR #714, opened 2018-10-22
Fix a race condition on deferred adapters

swankjesse:
This changes how we lookup and cache adapters. Previously we were pretty
optimistic about putting adapters in the cache; these adapters could have
depended upon stubs that were incomplete.

Now we're a lot more careful: we only put adapters in the cache if the
root object that triggered a set of recursive calls was constructed
successfully.

Closes: https://github.com/square/moshi/issues/679

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/Moshi.java
+++ b/moshi/src/main/java/com/squareup/moshi/Moshi.java
@@ -19,9 +19,12 @@ import com.squareup.moshi.internal.Util;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Deque;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -48,7 +51,7 @@ public final class Moshi {
   }
 
   private final List<JsonAdapter.Factory> factories;
-  private final ThreadLocal<List<DeferredAdapter<?>>> reentrantCalls = new ThreadLocal<>();
+  private final ThreadLocal<LookupChain> lookupChainThreadLocal = new ThreadLocal<>();
   private final Map<Object, JsonAdapter<?>> adapterCache = new LinkedHashMap<>();
 
   Moshi(Builder builder) {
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -283,38 +245,128 @@ Builder addAll(List<JsonAdapter.Factory> factories) {
...
-   * <p>Typically this is necessary in self-referential object models, such as an {@code Employee}
-   * class that has a {@code List<Employee>} field for an organization's management hierarchy.
+   * <p>This class defers putting any JSON adapters in the cache until the topmost JSON adapter has
+   * successfully been computed. That way we don't pollute the cache with incomplete stubs, or
+   * adapters that may transitively depend on incomplete stubs.
    */
-  private static final class DeferredAdapter<T> extends JsonAdapter<T> {
+  final class Call {
+    final List<Lookup<?>> callLookups = new ArrayList<>();
+    final List<Lookup<?>> stack = new ArrayList<>();
```
JakeWharton: Seems like this could be a Deque/ArrayDeque since you only do head/tail operations and iteration.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -283,38 +245,128 @@ Builder addAll(List<JsonAdapter.Factory> factories) {
...
+   * List<Employee>} field for an organization's management hierarchy.
    *
-   * <p>Typically this is necessary in self-referential object models, such as an {@code Employee}
-   * class that has a {@code List<Employee>} field for an organization's management hierarchy.
+   * <p>This class defers putting any JSON adapters in the cache until the topmost JSON adapter has
+   * successfully been computed. That way we don't pollute the cache with incomplete stubs, or
+   * adapters that may transitively depend on incomplete stubs.
    */
-  private static final class DeferredAdapter<T> extends JsonAdapter<T> {
+  final class Call {
```
JakeWharton: If only to avoid confusion for having a third Call in our lives, is there anything else we can call this? don't have any good suggestions though...

swankjesse: First name was ReentrantCall. Maybe LookupRoot ?

swankjesse: LookupChain? LookupTree?

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -283,38 +245,128 @@ Builder addAll(List<JsonAdapter.Factory> factories) {
...
+              .append(lookup.fieldName);
+        }
+      }
+
+      return new IllegalArgumentException(errorMessageBuilder.toString(), e);
+    }
+  }
+
+  /** This class implements {@code JsonAdapter} so it can be used as a stub for re-entrant calls. */
+  static final class Lookup<T> extends JsonAdapter<T> {
```
JakeWharton: Does this need a delegating toString()?

moshi/src/main/java/com/squareup/moshi/Moshi.java
JakeWharton: What about reverting `Lookup` to `DeferredAdapter` and naming this as the (stateful) `Lookup`?

Otherwise root and chain both seem fine.

moshi/src/main/java/com/squareup/moshi/Moshi.java
swankjesse: Good idea. Fixed.

moshi/src/main/java/com/squareup/moshi/Moshi.java
swankjesse: Renamed to LookupChain!

moshi/src/main/java/com/squareup/moshi/Moshi.java
swankjesse: Fixed!

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -283,38 +248,132 @@ Builder addAll(List<JsonAdapter.Factory> factories) {
   }
 
   /**
-   * Sometimes a type adapter factory depends on its own product; either directly or indirectly.
-   * To make this work, we offer this type adapter stub while the final adapter is being computed.
-   * When it is ready, we wire this to delegate to that finished adapter.
+   * A possibly-reentrant chain of lookups a JSON adapter.
```
NightlyNexus: lookups *for* a JSON adapter?

moshi/src/test/java/com/squareup/moshi/DeferredAdapterTest.java
```diff
@@ -0,0 +1,117 @@
...
+    Moshi moshi = new Moshi.Builder()
+        .add(factory)
+        .build();
+
+    JsonAdapter<BlueNode> jsonAdapter = moshi.adapter(BlueNode.class);
+    assertThat(jsonAdapter.toJson(new BlueNode(new GreenNode(new BlueNode(null, null)), null)))
+        .isEqualTo("{\"green\":{\"blue\":{}}}");
+
+    assertThat(failures).isEmpty();
+  }
```
NightlyNexus: liked reading this test!

moshi/src/test/java/com/squareup/moshi/DeferredAdapterTest.java
```diff
@@ -0,0 +1,117 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public final class DeferredAdapterTest {
+  /**
+   * When a type's JsonAdapter is circularly-dependent, Moshi creates a 'deferred adapter' to make
+   * the cycle work. It's important that any adapters that depend on this deferred adapter don't
+   * leak out until it's ready.
+   *
+   * <p>This test sets up a circular dependency BlueNode to GreenNode to BlueNode and then tries to
+   * use it before the BlueNode JSON adapter is built, but after the GreenNode adapter is built. It
```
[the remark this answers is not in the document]
swankjesse: Follow up coming that improves these docs.

moshi/src/main/java/com/squareup/moshi/Moshi.java
swankjesse: Fixed!

swankjesse:
https://github.com/square/moshi/pull/729

PR #726, opened 2018-11-03
Update to KotlinPoet 1.0.0-RC2.

NightlyNexus:
Also, use KotlinPoet's escaping for $ characters in formatted strings.
This is partly a revert of 01f600c (https://github.com/square/moshi/pull/604/).


Egorand:
Nice, thanks @NightlyNexus! I like that it helped remove a few hacks.

PR #731, opened 2018-11-04
Use javaObjectType instead of Java primitives while creating parametrized type

### LINKED ISSUE #725: Kotlin codegen produces code with warning for a collection of primitives
```
class Data(val listOfInts: List<Int>)
```
Compilation of generated adapter for this class produces warning `This class shouldn't be used in Kotlin. Use kotlin.Int instead`. And you can't suppress a warning inside generated code, so you have to remove `-Werror` from your compiler args :(

I am using Moshi 1.7.0 and Kotlin 1.2.71.

daugeldauge:
Fixes #725


hzsweers:
D'oh just saw #730 :)

swankjesse:
Nice test!

PR #733, opened 2018-11-04
Fix variance issues

daugeldauge:
There are several issues with variance:
```
data class Data(
    /** Doesn't compile: `String?::class` in generated adapter */
    val nullableWildcardOut: MutableList<out String?>,
    /** Doesn't compile: Can't assign MutableList<out Any> to MutableList<Any> */
    val anyOut: MutableList<out Any>,
    /**
     * Doesn't compile: `Any?::class` in generated adapter
     * This will become a star projection after kotlinpoet will be fixed https://github.com/square/kotlinpoet/issues/520
     */
    val nullableAnyOut: MutableList<out Any?>
)
```


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -108,15 +108,7 @@ internal fun Type.asTypeName(
               nullableProjection?.let { projection ->
                 when (projection) {
                   Type.Argument.Projection.IN -> WildcardTypeName.supertypeOf(argumentTypeName)
-                  Type.Argument.Projection.OUT -> {
-                    if (argumentTypeName == ANY) {
-                      // This becomes a *, which we actually don't want here.
-                      // List<Any> works with List<*>, but List<*> doesn't work with List<Any>
```
daugeldauge: I've removed this because it is wrong in general. The logic works for `List` only because it has declaration-site `out` variance. And `out Any` should never be printed as `*`, see https://github.com/square/kotlinpoet/issues/520

PR #744, opened 2018-11-17
Allow Object base type for PolymorphicJsonAdapterFactory.

NightlyNexus:
This works now.
Using general types like Object, Map, or List for the base type is error-prone, but the checks for these cases are not worth the code cost.


NightlyNexus:
the test in 2b6e6547142d6e1875bb36923cfbbc656dcf29db shows it works.

PR #745, opened 2018-11-17
Fix infinite loop when calling skipValue at the end an object or array.

NightlyNexus:
Fail with a clear message. This is useful when iteratoring through an array, making assumptions about the size without using hasNext.

Also, disallow calling skipValue at the end of a document. This is a behavior change.


moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -988,6 +988,44 @@ JsonReader newReader(String json) throws IOException {
     assertThat(reader.hasNext()).isFalse();
   }
 
+  @Test public void skipValueAtEndOfObjectFails() throws IOException {
```
swankjesse: awesome

PR #746, opened 2018-11-17
WIP: Add JsonReader.readJsonString.

### LINKED ISSUE #675: JsonReader.readJsonString
The opposite of `JsonWriter.value(BufferedSource)` which allows you to read a value as raw JSON. It can return either a String or ByteString depending on the ergonomics you want.

It can be implemented by [`peek()`-ing][1] the source, calling `skipValue()`, and then taking the offset you stopped at and calling `readUtf8`/`readByteString` with that offset (as a length) on the original source.

 [1]: https://github.com/square/okio/issues/506

NightlyNexus:
Returns the next JSON name or value.
Closes #675.

I fixed up some old code I had and added a sample. I need to find a way to reduce code duplication here. Possibly by peeking, skipping, and then reading as Jake suggested.

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
+++ b/examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
@@ -1,0 +1,64 @@
+package com.squareup.moshi.recipes;
+
+import com.squareup.moshi.FromJson;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonQualifier;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.ToJson;
+import java.io.IOException;
+import java.lang.annotation.Retention;
+import okio.Buffer;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class ReadJsonString {
+  @Retention(RUNTIME)
+  @JsonQualifier
+  public @interface JsonString {
+    Object ADAPTER = new Object() {
+      @FromJson @JsonString String fromJson(JsonReader reader) throws IOException {
+        return reader.readJsonString().utf8();
+      }
+
+      @ToJson void toJson(JsonWriter writer, @JsonString String value) throws IOException {
+        writer.value(new Buffer().writeUtf8(value));
+      }
+    };
+  }
+
+  static final class Location {
+    final String name;
+    final @JsonString String metadata;
+
+    Location(String name, String metadata) {
+      this.name = name;
+      this.metadata = metadata;
+    }
+  }
+
+  void run() throws IOException {
+    Moshi moshi = new Moshi.Builder()
+        .add(JsonString.ADAPTER)
+        .build();
+    JsonAdapter<Location> locationAdapter = moshi.adapter(Location.class);
+    String json = "{\n"
+        + "  \"name\": \"Niagara Falls\",\n"
+        + "  \"metadata\": {\n"
+        + "    \"latitude\": 43.0962,\n"
+        + "    \"longitude\": -79.0377\n"
+        + "  }\n"
+        + "}";
+    Location location = locationAdapter.fromJson(json);
+    System.out.println(location.metadata);
+    System.out.println(locationAdapter.toJson(new Location("Niagara Falls", "{\n"
+        + "    \"latitude\": 43.0962,\n"
+        + "    \"longitude\": -79.0377\n"
+        + "  }")));
+  }
+
+  public static void main(String[] args) throws IOException {
+    new ReadJsonString().run();
+  }
+}
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -424,6 +424,12 @@ public abstract class JsonReader implements Closeable {
   public abstract void skipValue() throws IOException;
 
   /**
+   * Reads the next JSON name or value without decoding it. White space and comments will be
+   * removed.
+   */
+  public abstract ByteString readJsonString() throws IOException;
+
+  /**
    * Returns the value of the next token, consuming it. The result may be a string, number, boolean,
    * null, map, or list, according to the JSON structure.
    *
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -20,8 +20,10 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import javax.annotation.Nullable;
 import okio.Buffer;
+import okio.BufferedSink;
 import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
 
 final class JsonUtf8Reader extends JsonReader {
   private static final long MIN_INCOMPLETE_INTEGER = Long.MIN_VALUE / 10;
@@ -64,6 +66,8 @@ final class JsonUtf8Reader extends JsonReader {
   private static final int NUMBER_CHAR_EXP_E = 5;
   private static final int NUMBER_CHAR_EXP_SIGN = 6;
   private static final int NUMBER_CHAR_EXP_DIGIT = 7;
+
+  static final BufferedSink BLACKHOLE = Okio.buffer(Okio.blackhole());
 
   /** The input JSON. */
   private final BufferedSource source;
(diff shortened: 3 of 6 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -978,6 +994,62 @@ private void skipUnquotedValue() throws IOException {
...
+      p = doPeek();
+    }
+    if (p == PEEKED_BEGIN_ARRAY) {
+      sink.writeUtf8("[");
+    } else if (p == PEEKED_BEGIN_OBJECT) {
+      sink.writeUtf8("{");
+    } else if (p == PEEKED_DOUBLE_QUOTED || p == PEEKED_DOUBLE_QUOTED_NAME) {
+      sink.writeUtf8("\"");
+    } else if (p == PEEKED_SINGLE_QUOTED || p == PEEKED_SINGLE_QUOTED_NAME) {
+      sink.writeUtf8("'");
```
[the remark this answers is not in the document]
NightlyNexus: 👍

Dkhusainov:
I've tried it and it doesn't work with null values.

Test to reporduce:

`@Test public void readJsonStringWithNulls() throws IOException {
    JsonReader reader = newReader("{\n" +
            "  \"a\": {\n" +
            "    \"b\": null,\n" +
            "    \"c\": \"\\t\\nqweert\",\n" +
            "    \"d\": null\n" +
            "  }\n" +
            "}");
    reader.setLenient(true);

    reader.beginObject();
    reader.skipName();
    String supposedToBeValidJson = reader.readJsonString().utf8();

    System.out.println(supposedToBeValidJson);

    newReader(supposedToBeValidJson).readJsonValue();
  }`

NightlyNexus:
Yep, this revision doesn't work with literals like null and true and false.
I'm coming back to this change when I have time.

examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
```diff
@@ -0,0 +1,64 @@
+package com.squareup.moshi.recipes;
```
swankjesse: nit: copyright header

examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.IOException;
+import java.lang.annotation.Retention;
+import okio.Buffer;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+final class ReadJsonString {
+  @Retention(RUNTIME)
+  @JsonQualifier
+  public @interface JsonString {
```
swankjesse: nit: this is a weird example because it looks like it should be part of the framework

swankjesse: if it was a name like @LocationMetadata that would be less of-the-framework

examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
```diff
@@ -0,0 +1,64 @@
...
+
+      @ToJson void toJson(JsonWriter writer, @JsonString String value) throws IOException {
+        writer.value(new Buffer().writeUtf8(value));
+      }
+    };
+  }
+
+  static final class Location {
+    final String name;
+    final @JsonString String metadata;
```
swankjesse: honest question: should we encourage users to store these as ByteStrings? If it’s an opaque blob it might as well be?

examples/src/main/java/com/squareup/moshi/recipes/ReadJsonString.java
```diff
@@ -0,0 +1,64 @@
...
+    String json = "{\n"
+        + "  \"name\": \"Niagara Falls\",\n"
+        + "  \"metadata\": {\n"
+        + "    \"latitude\": 43.0962,\n"
+        + "    \"longitude\": -79.0377\n"
+        + "  }\n"
+        + "}";
+    Location location = locationAdapter.fromJson(json);
+    System.out.println(location.metadata);
+    System.out.println(locationAdapter.toJson(new Location("Niagara Falls", "{\n"
```
swankjesse: one interesting feature here is that the `equals()` method totally won’t work unless the user canonicalizes the JSON

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -423,6 +423,12 @@ public final void setFailOnUnknown(boolean failOnUnknown) {
    */
   public abstract void skipValue() throws IOException;
 
+  /**
+   * Reads the next JSON name or value without decoding it. White space and comments will be
+   * removed.
```
swankjesse: Should we be more formal here? Like should we say anything about which characters are encoded? Should we make promises around how numbers are formatted?

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -228,13 +232,17 @@
...
+
+  int doPeek(BufferedSink sink) throws IOException {
     int peekStack = scopes[stackSize - 1];
     if (peekStack == JsonScope.EMPTY_ARRAY) {
       scopes[stackSize - 1] = JsonScope.NONEMPTY_ARRAY;
     } else if (peekStack == JsonScope.NONEMPTY_ARRAY) {
       // Look for a comma before the next element.
       int c = nextNonWhitespace(true);
-      buffer.readByte(); // consume ']' or ','.
+      sink.write(buffer, 1); // consume ']' or ','.
```
swankjesse: neat!

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -310,7 +313,7 @@
 
     if (skipped instanceof Map.Entry) {
       // We're skipping a name. Promote the map entry's value.
-      Map.Entry<?, ?> entry = (Map.Entry<?, ?>) stack[stackSize - 1];
+      Map.Entry<String, ?> entry = (Map.Entry<String, ?>) stack[stackSize - 1];
```
swankjesse: FWIW, this change makes it go from a safe cast to an unsafe cast

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -1147,4 +1147,54 @@ private void readValue(JsonReader reader, boolean peekJsonFirst) throws IOExcept
...
+  @Test public void readJsonStringArray() throws IOException {
+    JsonReader reader = newReader("{\"pizzas\": [\"cheese\", \"pepperoni\"]}");
+    reader.beginObject();
+    reader.skipName();
+    assertEquals("[\"cheese\",\"pepperoni\"]", reader.readJsonString().utf8());
+    reader.endObject();
+    assertThat(reader.peek()).isEqualTo(JsonReader.Token.END_DOCUMENT);
+  }
+
+  @Test public void readJsonStringName() throws IOException {
```
swankjesse: should this be allowed?

swankjesse:
I played with making a method to return a `Source`. It’s doable, but we’d need to build a skipping variant of `doPeek()` which gets pretty ugly pretty fast. Not sure where to go next. I’d like to go back and study what the use cases want.

Dkhusainov:
Our use case is that we are saving json stream to db, where nested objects/arrays saved to db as json strings to later use SQLite `json1` extension to query it. Best case for us is to have `String JsonReader.readValueAsString(): String ` with as minimum allocations as possible.

NightlyNexus:
@Dkhusainov if we make the method return a Source instead of a string, you can skip the buffered memory, and stream directly to your disk database.

Dkhusainov:
I would't do that becasue:
1. Too low level and hard
2. Doubt it's even possible with SQLite without doing some native stuff(see p1).

Returning a `String`(or `Buffer`) would be sufficent(and would cover more use cases in general I think). Right now it's just deserializing and then serializing json back, which is far from optimal

angusholder:
Returning `Source` provides a superset of the functionality of returning `String`; if you have a use-case where you want to avoid buffering the JSON string into memory, `Source` allows you to do that, and if not, you can just call `readJsonSource().buffer().readUtf8()`. Then maybe just provide a convenience method `readJsonString()` that is equivalent to the above.

The streaming to database is very much possible in Java, for example there's the `java.sql.Blob` API where you can call `setBytes` to do partial writes to an already-allocated blob in the database, so could use that to stream say 8KB of JSON at a time.

Dkhusainov:
I don't mind `Source`.

About database streaming: there is no currently no way to stream bytes into `SQLite` on Android without using JNI or something like that. I don't even think `SQLite` api itself supports it.

macdanny:
I am also very interested in this feature for a similar use case. I have an object hierarchy that at some depth has a JSON array. I want to pull the original JSON for each element in the array and save them into a database for later parsing. In my case the database is JetBrains Xodus and could make use of a stream of bytes if I had one to hand, but a String would also be usable as each record would be relatively small.

Dkhusainov:
Any ETA on when this will be implemented? 

Are there blockers in `moshi` that need to be done before `readJsonString` is possible?

Dkhusainov:
😴

ZacSweers:
Sorry for the delay on this. We're going to go with a better/more flexible solution described in https://github.com/square/moshi/issues/675#issuecomment-536130040 (PRs welcome if anyone wants to contribute it sooner!)

PR #749, opened 2018-11-19
Refuse kotlin classes in ClassJsonAdapter

### LINKED ISSUE #748: Require reflection or codegen for Kotlin types
When you don’t have either installed things sort of work, but there are traps. It would be better to fail early with an obvious explanation of what’s wrong.

https://github.com/square/moshi/issues/743

hzsweers:
Resolves #748

There are two existing tests that fail with this now, let me know what you all want to do with them (migrate/adjust/delete/etc)

`KotlinJsonAdapterTest#delegatesToInstalledAdaptersBeforeNullChecking`

`KotlinJsonAdapterTest#nullablePrimitivesUseBoxedPrimitiveAdapters`

### DIFF
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -957,7 +957,7 @@ class KotlinJsonAdapterTest {
     val moshi = Moshi.Builder().build()
     try {
       moshi.adapter<PlainKotlinClass>(PlainKotlinClass::class.java)
-      throw AssertionError("Should not pass here")
+      fail("Should not pass here")
     } catch (e: IllegalArgumentException) {
       assertThat(e).hasMessageContaining("Reflective serialization of Kotlin classes")
     }
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -952,4 +952,16 @@ class KotlinJsonAdapterTest {
     assertThat(adapter.fromJson("null")).isNull()
     assertThat(adapter.toJson(null)).isEqualTo("null")
   }
+
+  @Test fun kotlinClassesWithoutAdapterAreRefused() {
+    val moshi = Moshi.Builder().build()
+    try {
+      moshi.adapter<PlainKotlinClass>(PlainKotlinClass::class.java)
+      throw AssertionError("Should not pass here")
+    } catch (e: IllegalArgumentException) {
+      assertThat(e).hasMessageContaining("Reflective serialization of Kotlin classes")
+    }
+  }
+
+  class PlainKotlinClass
 }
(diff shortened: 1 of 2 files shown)


NightlyNexus:
Simply my mistakes in those tests. will need a rebase after #750

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -950,4 +950,16 @@ class KotlinJsonAdapterTest {
     assertThat(adapter.fromJson("null")).isNull()
     assertThat(adapter.toJson(null)).isEqualTo("null")
   }
+
+  @Test fun kotlinClassesWithoutAdapterAreRefused() {
+    val moshi = Moshi.Builder().build()
+    try {
+      moshi.adapter<PlainKotlinClass>(PlainKotlinClass::class.java)
+      throw AssertionError("Should not pass here")
```
NightlyNexus: nit: use `fail()`? for consistency with all the other test cases.

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -75,6 +75,14 @@
       if (Modifier.isAbstract(rawType.getModifiers())) {
         throw new IllegalArgumentException("Cannot serialize abstract class " + rawType.getName());
       }
+      for (Annotation annotation : rawType.getDeclaredAnnotations()) {
+        if ("kotlin.Metadata".equals(annotation.annotationType().getName())) {
+          throw new IllegalArgumentException("Cannot serialize Kotlin type " + rawType.getName()
+              + ". Reflective serialization of Kotlin classes without using kotlin-reflect can "
+              + "have undefined and unexpected behavior. Please use KotlinJsonAdapter from the "
```
NightlyNexus: nit: instead of "can have undefined and unexpected behavior," how about "has undefined and unexpected behavior?"

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
hzsweers: derp, good call 7058846

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
hzsweers: Done ca79cce

hzsweers:
Rebased!

moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -75,6 +75,14 @@
       if (Modifier.isAbstract(rawType.getModifiers())) {
         throw new IllegalArgumentException("Cannot serialize abstract class " + rawType.getName());
       }
+      for (Annotation annotation : rawType.getDeclaredAnnotations()) {
```
swankjesse: This is a very expensive reflection call. Calling isAnnotationPresent is much faster. Can you change to that?

swankjesse: You might need to load the metadata class using Class.forName() and that can fail. If it does, I don't _think_ it's possible for Kotlin to be present.

hzsweers: > Calling isAnnotationPresent is much faster. Can you change to that?

Can do, completely forgot about that. #754

> I don't think it's possible for Kotlin to be present

Kotlin technically could be present, but only if the stdlib isn't used. We wouldn't be able to tell in this case :/

PR #765, opened 2018-12-02
Add reified adapter extension for Kotlin

rom4ek:
I thought it would be nice to have this extension similar to how it's done in [retrofit](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/KotlinExtensions.kt)

```Kotlin
val adapter = moshi.adapter<BlackjackHand>()
```

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/KotlinExtensions.kt
+++ b/moshi/src/main/java/com/squareup/moshi/KotlinExtensions.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Hide the class from Java consumers.
+@file:JvmName("-KotlinExtensions")
+
+package com.squareup.moshi
+
+inline fun <reified T> Moshi.adapter(): JsonAdapter<T> = adapter(T::class.java)
--- a/moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
+++ b/moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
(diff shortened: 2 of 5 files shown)


moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import org.assertj.core.api.Assertions.assertThatCode
+import org.junit.Test
+
+class KotlinExtensionsTest {
+
+  @Test fun reifiedAdapter() {
+    val moshi = Moshi.Builder().build()
+
+    assertThatCode { moshi.adapter<TestUtil.Empty>() }.doesNotThrowAnyException()
```
hzsweers: nit: can we smoke test a simple happy path here? (as opposed to just making sure it doesn't crash, we want to also make sure it still works)

moshi/src/main/java/com/squareup/moshi/KotlinExtensions.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Hide the class from Java consumers.
+@file:JvmName("-KotlinExtensions")
+
+package com.squareup.moshi
+
+inline fun <reified T> Moshi.adapter(): JsonAdapter<T> = adapter(T::class.java)
```
hzsweers: What happens if I do 

```kotlin
moshi.adapter<List<String>>()
```

?

My sense is it doesn't work, but may not be avoidable either

rom4ek: Mhm, it returns a `CollectionJsonAdapter`, which I assume is a correct one?

hzsweers: Yes, but I'm willing to bet it's a raw type and treats it as a `List<Object>`

rom4ek:
I actually overlooked method overloads, extending the extension with default arguments right now. Or your concern is about that we can't support `Type` with this approach?

hzsweers:
Yes, and in particular that it might be a footgun to people that think `adapter<List<String>>` works when it maybe fails at runtime :/

moshi/src/main/java/com/squareup/moshi/KotlinExtensions.kt
rom4ek: Yep, you're right. But looks like we could overcome this with super type tokens. Not sure how idiomatic it is though. But it's also done like this in [Jackson extensions for Kotlin](https://github.com/FasterXML/jackson-module-kotlin/blob/8ab36da8ba1115d0ef1c2ff20a2f6b1a221fdf73/src/main/kotlin/com/fasterxml/jackson/module/kotlin/Extensions.kt)

swankjesse:
Yeah, without being able to do `List<String>` I think this approach creates more problems than it solves.

rom4ek:
Ok, I see the point guys. Closing this then. Still, thanks for your reviews.

PR #769, opened 2018-12-11
KotlinPoet 1.0.0

Egorand:


swankjesse:
Yeah, though that doesn’t block this. Lets do both!
https://github.com/square/moshi/issues/770

PR #773, opened 2018-12-14
Create Benchmarks Project

AndrewReitz:
Added benchmarks for to and from json using moshi-kotlin-generator, jackson, and gson with
JMH.


benchmarks/src/main/java/com/squareup/benchmarks/FromJsonBenchmarkBase.kt
```diff
@@ -0,0 +1,37 @@
+package com.squareup.benchmarks
```
swankjesse: nit: copyright header

benchmarks/src/main/java/com/squareup/benchmarks/gson/FromJsonBenchmark.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import com.squareup.benchmarks.FromJsonBenchmarkBase
+import com.squareup.benchmarks.gson.model.Response
+import org.openjdk.jmh.annotations.Benchmark
+import org.openjdk.jmh.annotations.Level
+import org.openjdk.jmh.annotations.Setup
+import java.io.InputStreamReader
+
+open class FromJsonBenchmark : FromJsonBenchmarkBase() {
```
swankjesse: flatten this and its base? Typically with JMH you combine benchmarks with parameters, not subtyping

benchmarks/src/main/java/com/squareup/benchmarks/gson/ToJsonBenchmark.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.squareup.benchmarks.ToJsonBenchmarkBase
+import com.squareup.benchmarks.gson.model.Response
+import com.squareup.benchmarks.json
+import org.openjdk.jmh.annotations.Benchmark
+import org.openjdk.jmh.annotations.Level
+import org.openjdk.jmh.annotations.Setup
+import java.io.ByteArrayOutputStream
+import java.io.OutputStreamWriter
+
+open class ToJsonBenchmark: ToJsonBenchmarkBase() {
```
swankjesse: Similarly: flatten this and its base?

benchmarks/src/main/java/com/squareup/benchmarks/gson/FromJsonBenchmark.kt
swankjesse: ... you’d have `@Benchmark` methods like fromJsonWithGson, fromJsonWithMoshi, etc.

benchmarks/src/main/java/com/squareup/benchmarks/gson/model/Friend.java
```diff
@@ -0,0 +1,43 @@
+package com.squareup.benchmarks.gson.model;
+
+import java.util.Objects;
+
+public final class Friend {
```
swankjesse: Single model for all libraries?

benchmarks/src/main/java/com/squareup/benchmarks/io.kt
```diff
@@ -0,0 +1,14 @@
...
+package com.squareup.benchmarks
+
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.Okio
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.file.Path
+
+inline fun InputStream.toBufferSource(): BufferedSource = Okio.buffer(Okio.source(this))
```
swankjesse: nit: toBufferedSource ?

benchmarks/src/main/java/com/squareup/benchmarks/moshi.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import java.util.*
+
+val moshi: Moshi = Moshi.Builder()
+        .add(Date::class.java, Rfc3339DateJsonAdapter())
+        .add(KotlinJsonAdapterFactory())
+        .build()
+
+/**
+ * A little magic for kotlin to only have to provide the type we want instead of the class file.
```
swankjesse: Remove this? It doesn’t work in cases where `T` is not a simple type (such as `List<Int>`) but the compiler won’t help you to discover that

benchmarks/src/main/java/com/squareup/benchmarks/moshi.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param T the type the adapter is for.
+ */
+inline fun <reified T> Moshi.adapter(): JsonAdapter<T> = adapter<T>(T::class.java)
+
+/**
+ * Like [adapter] but for a List type.
```
swankjesse: Remove this? Doing it longhand is more code, but less magic

Issue #783, opened 2019-01-06
Confirm codegen can handle package names that look like class names


swankjesse:
```
package com.example.Project
```

hzsweers:
FWIW the IDE does gripe about trying to do this

![image](https://user-images.githubusercontent.com/1361086/51081347-6a925a00-16a2-11e9-9721-981addb64206.png)

Adding a test case for this in an upcoming PR but seems to work fine

NightlyNexus:
i think there's still a problem via a usage of bestGuess for a type name.

try with a property from a package that looks like a class name:

```kotlin
package com.squareup.moshi.kotlin.codgen.LooksLikeAClass

import com.squareup.moshi.JsonClass

/**
 * https://github.com/square/moshi/issues/783
 */
@JsonClass(generateAdapter = true)
data class ClassInPackageThatLooksLikeAClass(val foo: String)

@JsonClass(generateAdapter = true)
data class HasPropertyWithClassInPackageThatLooksLikeAClass(
  val foo: ClassInPackageThatLooksLikeAClass
)
```

<img width="981" alt="screen shot 2019-01-15 at 2 50 35 pm" src="https://user-images.githubusercontent.com/4032667/51215191-04414d80-18d5-11e9-8b26-1a518149908e.png">

ZacSweers:
Would that be more of an issue for Kotlinpoet?

swankjesse:
bestGuess is bad and careful code shouldn’t use it.

PR #786, opened 2019-01-13
Some miscellaneous kotlinpoet updates

hzsweers:
- Update KotlinPoet to 1.0.1
- Drop `asNullableIf` in favor of the plain `copy()` functions
- Add a test for #783


swankjesse:
👌🏻

PR #791, opened 2019-01-15
Support multiple transient properties in KotlinJsonAdapter.

### LINKED ISSUE #775: Kotlin multiple @Transient fields
Having more than one @ Transient field is causing following exception.

```
Exception in thread "main" java.lang.IllegalArgumentException: duplicate option: [text=\\u0000"]
	at okio.Options.of(Options.java:66)
	at com.squareup.moshi.JsonReader$Options.of(JsonReader.java:538)
	at com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory.create(KotlinJsonAdapter.kt:260)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:137)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:97)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:71)

```

Code to reproduce

```
data class TransientTest(
        val a: String,
        @Transient
        val b: String? = "b",
        @Transient
        val c: String? = "c"
)

fun main(args: Array<String>) {

    val adapter = Moshi.Builder()
            .build()
            .adapter(TransientTest::class.java)

    val test = TransientTest("a")

    adapter.toJson(test)
}
```

NightlyNexus:
Fixes #775

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -54,35 +54,36 @@ private val ABSENT_VALUE = Any()
  * constructor, and then by setting any additional properties that exist, if any.
  */
 internal class KotlinJsonAdapter<T>(
-  private val constructor: KFunction<T>,
-  private val bindings: List<Binding<T, Any?>?>,
-  private val options: JsonReader.Options
+  val constructor: KFunction<T>,
+  val allBindings: List<Binding<T, Any?>?>,
+  val nonTransientBindings: List<Binding<T, Any?>>,
+  val options: JsonReader.Options
 ) : JsonAdapter<T>() {
 
   override fun fromJson(reader: JsonReader): T {
     val constructorSize = constructor.parameters.size
 
     // Read each value into its slot in the array.
-    val values = Array<Any?>(bindings.size) { ABSENT_VALUE }
+    val values = Array<Any?>(allBindings.size) { ABSENT_VALUE }
     reader.beginObject()
     while (reader.hasNext()) {
       val index = reader.selectName(options)
-      val binding = if (index != -1) bindings[index] else null
-
-      if (binding == null) {
+      if (index == -1) {
         reader.skipName()
         reader.skipValue()
         continue
       }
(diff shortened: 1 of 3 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -55,34 +54,35 @@ private val ABSENT_VALUE = Any()
...
         continue
       }
+      val binding = nonTransientBindings[index]
 
-      if (values[index] !== ABSENT_VALUE) {
+      val propertyIndex = binding.propertyIndex
+      if (values[propertyIndex] !== ABSENT_VALUE) {
         throw JsonDataException(
-            "Multiple values for '${constructor.parameters[index].name}' at ${reader.path}")
+            "Multiple values for '${binding.property.name}' at ${reader.path}")
```
swankjesse: nice

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -55,34 +54,35 @@ private val ABSENT_VALUE = Any()
  * constructor, and then by setting any additional properties that exist, if any.
  */
 internal class KotlinJsonAdapter<T>(
-    val constructor: KFunction<T>,
-    val bindings: List<Binding<T, Any?>?>,
-    val options: JsonReader.Options) : JsonAdapter<T>() {
+  val constructor: KFunction<T>,
+  val allBindings: List<Binding<T, Any?>?>,
+  val nonTransientBindings: List<Binding<T, Any?>>,
```
swankjesse: I find this more difficult to follow now. Can we accomplish the same work with a single list still?

ZacSweers: I've been fiddling with this locally, and I don't think there's a way to get a single list here without making the JsonAdapterFactory logic more convoluted than just having this list is. I'd like to suggest we rebase this PR and merge with the current logic to get the issue resolved, and can revisit the readability separately. WDYT?

PR #792, opened 2019-01-15
Support fail on unknown in PolymorphicJsonAdapterFactory.

NightlyNexus:
Before, this would fail when skipping to find the index of the label.
Note that this still requires the type to have a label field.


NightlyNexus:
#793 for leaving off the `type` field. might not be worth the complexity.

PR #798, opened 2019-01-28
Retain annotated Enum field names with proguard

SimonMarquis:
#691 fixed a related issue but only when using the `@JsonClass` annotation on the Enum itself.
Unfortunately `JsonClass` has some restrictions (and can't be used in java):
https://github.com/square/moshi/blob/ded3bccc6089c8a39ca1ebf7227fcee122572512/moshi/src/main/java/com/squareup/moshi/JsonClass.java#L34-L38

### DIFF
--- a/moshi/src/main/resources/META-INF/proguard/moshi.pro
+++ b/moshi/src/main/resources/META-INF/proguard/moshi.pro
@@ -7,6 +7,11 @@
 
 -keep @com.squareup.moshi.JsonQualifier interface *
 
+# Retain annotated Enum field names.
+-keepclassmembers class * extends java.lang.Enum {
+    @com.squareup.moshi.* <fields>;
+}
+
 # Enum field names are used by the integrated EnumJsonAdapter.
 # Annotate enums with @JsonClass(generateAdapter = false) to use them with Moshi.
 -keepclassmembers @com.squareup.moshi.JsonClass class * extends java.lang.Enum {


NightlyNexus:
oh, i thought this worked as-is if you annotate your Java enum with `@JsonClass(generateAdapter = false)`. what happens when you do that?

SimonMarquis:
Indeed the Enums are retained when adding `@JsonClass(generateAdapter = false)`.

But from a user's perspective, it's not very clear why you should use this annotation in the first place.
The documentation introduces `JsonClass` as a CodeGen related feature.

For a simple java example like this, I think retaining the enum without adding the `@JsonClass` annotation would be a perfectly valid default behavior.

```java
enum State {
    @Json(name = "state_1") ONE,
    @Json(name = "state_2") TWO,
    @Json(name = "state_3") THREE
}
```

SimonMarquis:
Do you think there are cases where this proguard rule would retain too much data and prevent potential obfuscation?

swankjesse:
ProGuard is useful for code-shrinking and optimization.

I do not recommend mixing JSON with obfuscated code. It's unlikely to be effective, especially as it's usually quite easy for an attacker to intercept her own traffic.

If you need to slow down reverse engineering your code, prefer a binary protocol like protobuf.

SimonMarquis:
> If you need to slow down reverse engineering your code, prefer a binary protocol like protobuf.

That wasn't my intention here but rather a side effect of running Proguard on an _internal_ package of a library.

swankjesse:
I don’t like this because it isn’t semantic.

In particular, I don’t like that somebody could remove an unnecessary annotation and break the code:

```
enum Roshambo {
  ROCK,
  @Json(name = "SCISSORS") SCISSORS,
  PAPER

}
```

Issue #800, opened 2019-01-29
Codegen uses wrong name in generated exception message


dvdmunckhof:
Test data class:
```kotlin
package com.company.package

import com.squareup.moshi.Json
import com.squareup.moshi.JsonClass

@JsonClass(generateAdapter = true)
data class Test(
        @Json(name = "json_property_name")
        val kotlinPropertyName: String
)
```

Resulting in the generated adapter:
```kotlin
// Code generated by moshi-kotlin-codegen. Do not edit.
package com.company.package

import com.squareup.moshi.JsonAdapter
import com.squareup.moshi.JsonDataException
import com.squareup.moshi.JsonReader
import com.squareup.moshi.JsonWriter
import com.squareup.moshi.Moshi
import java.lang.NullPointerException
import kotlin.String

class TestJsonAdapter(moshi: Moshi) : JsonAdapter<Test>() {
    private val options: JsonReader.Options = JsonReader.Options.of("json_property_name")

    private val stringAdapter: JsonAdapter<String> =
            moshi.adapter<String>(String::class.java, kotlin.collections.emptySet(), "kotlinPropertyName")

    override fun toString(): String = "GeneratedJsonAdapter(Test)"

    override fun fromJson(reader: JsonReader): Test {
        var kotlinPropertyName: String? = null
        reader.beginObject()
        while (reader.hasNext()) {
            when (reader.selectName(options)) {
                0 -> kotlinPropertyName = stringAdapter.fromJson(reader) ?: throw JsonDataException("Non-null value 'kotlinPropertyName' was null at ${reader.path}")
                -1 -> {
                    // Unknown name, skip it.
                    reader.skipName()
                    reader.skipValue()
                }
            }
        }
        reader.endObject()
        var result = Test(
                kotlinPropertyName = kotlinPropertyName ?: throw JsonDataException("Required property 'kotlinPropertyName' missing at ${reader.path}"))
        return result
    }

    override fun toJson(writer: JsonWriter, value: Test?) {
        if (value == null) {
            throw NullPointerException("value was null! Wrap in .nullSafe() to write nullable values.")
        }
        writer.beginObject()
        writer.name("json_property_name")
        stringAdapter.toJson(writer, value.kotlinPropertyName)
        writer.endObject()
    }
}
```

Wrong part:
```kotlin
throw JsonDataException("Required property 'kotlinPropertyName' missing at ${reader.path}")
```

[...]

PR #801, opened 2019-01-29
Fix codegen generating wrong name in exception message (#800)

dvdmunckhof:
Fixes https://github.com/square/moshi/issues/800


dvdmunckhof:
Nevermind, this is pull request is incomplete and doesn't replace all instances. I'll give it another try tomorrow.

PR #815, opened 2019-02-18
Update moshi.pro for Proguard 6.0.x compatibility

### LINKED ISSUE #738: Build exception when using new Proguard rules
After updating to Moshi 1.7.0 and updating the Proguard-rules I've encountered a Gradle build exception: 

Exception while processing task java.io.IOException: proguard.ParseException: Use of generics not allowed for java type at '<1>_<2>_<3>JsonAdapter' in line 31 of file '/Library/TeamCity/buildAgent/work/3ce9226b513c2493/dummy/proguard/proguard-moshi.pro'

I'm now using my old proguard rules, and the build works fine, but I would like to know how to use the new Proguard rules. Here's my current rules which seems to work fine:

-dontwarn okio.**
-dontwarn javax.annotation.**
-keepclassmembers class * {
    @com.squareup.moshi.FromJson <methods>;
    @com.squareup.moshi.ToJson <methods>;
}
-keep @com.squareup.moshi.JsonQualifier interface *
    public <methods>;
}
-dontwarn org.jetbrains.annotations.**
-keep class kotlin.Metadata { *; }

-dontwarn kotlin.reflect.jvm.internal.**
-keep class kotlin.reflect.jvm.internal.** { *; }

venator85:
fixes #738 
fix by @pawelkw

### DIFF
--- a/moshi/src/main/resources/META-INF/proguard/moshi.pro
+++ b/moshi/src/main/resources/META-INF/proguard/moshi.pro
@@ -17,33 +17,7 @@
 -keepnames @com.squareup.moshi.JsonClass class *
 
 # Retain generated JsonAdapters if annotated type is retained.
--if @com.squareup.moshi.JsonClass class *
--keep class <1>JsonAdapter {
-    <init>(...);
-    <fields>;
-}
--if @com.squareup.moshi.JsonClass class **$*
--keep class <1>_<2>JsonAdapter {
-    <init>(...);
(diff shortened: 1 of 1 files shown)


ZacSweers:
This isn't a bug that needs fixing, we intentionally have these rules for proguard 6.1.x+ and R8. You should add these rules manually to your own configuration and exclude moshi's if you want proguard 6.0.x and below

PR #816, opened 2019-02-25
Decouple code gen API from apt and kotlin-metadata

ZacSweers:
This is a proposal implementation to decouple the code gen API from both annotation processing and kotlin-metadata APIs. 

The idea here is that this separation allows for the implementation of data for the code gen to be an implementation detail, and thus allows the code gen to be reusable for different front-ends. Right now it's heavily coupled to annotation processing + kotlin-metadata, but we could conceivably easily implement new support for things like 
* kotlinx-metadata (or its value-based alternative)
* Compiler plugins
* Non-apt environments

etc. with KotlinPoet being the common foundation in the API for all of them.

At the time of writing, this PR is mostly about separating out API pieces, but not much has been put into the actual design of an API that anything other than APT can use. ~~This also leans on a proposal PR in KotlinPoet to function: https://github.com/square/kotlinpoet/pull/633. As such, CI will fail here, but all the tests pass locally for me with no changes currently.~~


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -123,3 +188,345 @@ internal fun Type.asTypeName(
...
+  for (property in classProto.propertyList) {
+    val name = nameResolver.getString(property.name)
+    val type = typeResolver.resolve(property.returnType.asTypeName(
+        nameResolver, classProto::getTypeParameter, false))
+    val parameter = constructor.parameters[name]
+    val fieldElement = fields[name]
+    val annotationHolder = annotationHolders[name]
+    // Used for setter/getter/is lookups. Guaranteed to be safe because kotlin doesn't allow you to
+    // have both "AAA" and "aAA".
+    val decapitalizedName = name.decapitalizeAsciiOnly()
```
ZacSweers: I actually don't know how master passes without this currently

image

ZacSweers:
I've updated this with KotlinPoet 1.2.0 and rebased, would be curious for thoughts on pursuing this. It works now, next (possibly separate PR) would be trying to tease out a designed API for this

ZacSweers:
Another use case for this: https://github.com/Foso/MpApt

ZacSweers:
Going to move ahead with this! API is definitely not ready for public use, but it's at least decoupled now and will allow us to easily move to kotlinpoet-metadata next

PR #821, opened 2019-03-08
Support covariant Map values, which are used by Kotlin

Syeberman:


Syeberman:
This breaks tests added in https://github.com/square/moshi/pull/406, so this PR may be a bad idea.

moshi/src/test/java/com/squareup/moshi/MapJsonAdapterTest.java
```diff
@@ -86,6 +86,27 @@
     assertThat(jsonAdapter.fromJson(jsonReader)).isEqualTo(null);
   }
 
+  @Test public void covariantValue() throws Exception {
+    // Important for Kotlin maps, which are all Map<K, ? extends T>.
```
swankjesse: nice

swankjesse:
Need to figure out the tests, but I think it’s totally fine for us to treat `? extends Object` as a special case. It’s different from any other wildcard.

moshi/src/test/java/com/squareup/moshi/MapJsonAdapterTest.java
```diff
@@ -86,6 +86,27 @@
...
+    map.put("boolean", true);
+    map.put("float", 42.0);
+    map.put("String", "value");
+
+    String asJson = "{\"boolean\":true,\"float\":42.0,\"String\":\"value\"}";
+
+    Buffer buffer = new Buffer();
+    JsonWriter jsonWriter = JsonWriter.of(buffer);
+    jsonAdapter.toJson(jsonWriter, map);
+    assertThat(buffer.readUtf8()).isEqualTo(asJson);
```
NightlyNexus: i'd remove these 4 lines, since they're not related to this test.

Syeberman: I wanted to have this part of the test to ensure that serialization and deserialization worked with the adapter as-configured.

Syeberman:
@NightlyNexus As implemented, this should handle every `extends` wildcard.

NightlyNexus:
> @NightlyNexus As implemented, this should handle every `extends` wildcard.

Try getting an adapter for `Foo`.
```java
class Foo {
  List<? extends Bar> bars;
}

class Bar<T> {
}
```

Syeberman:
@NightlyNexus As I understand Moshi, that example will fail unless `Bar` has registered an adapter, which is I feel desired behaviour.

What I'm fixing in this PR is how Kotlin's `Map<K, V>` maps to Java's `Map<K, ? extends V>`. I'm also fixing all such wildcard uses involving the Java standard types.

NightlyNexus:
it's the same for maps.
i'm not sure i follow. why change the behavior for types like Boolean, String, Object, etc. only? won't this Kotlin behavior happen for all types?

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -982,4 +984,144 @@ class KotlinJsonAdapterTest {
...
+        MapOfStringToClassCodegen::class.java,
+        """{"map":{"key":{"a":19,"b":42}}}""",
+        MapOfStringToClassCodegen(mapOf("key" to ConstructorParameters(19, 42))))
+  }
+
+  @JsonClass(generateAdapter = true)
+  @JvmSuppressWildcards(suppress = false)
+  data class MapOfStringToClassCodegen(val map: Map<String, ConstructorParameters> = mapOf())
+
+  fun <T> mapWildcardsParametrizedTest(type: Class<T>, json: String, value: T) {
```
swankjesse: tiny nit: Parameterized

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
```diff
@@ -36,27 +36,28 @@ private StandardJsonAdapters() {
...
-      if (type == Character.class) return CHARACTER_JSON_ADAPTER.nullSafe();
-      if (type == Double.class) return DOUBLE_JSON_ADAPTER.nullSafe();
-      if (type == Float.class) return FLOAT_JSON_ADAPTER.nullSafe();
-      if (type == Integer.class) return INTEGER_JSON_ADAPTER.nullSafe();
-      if (type == Long.class) return LONG_JSON_ADAPTER.nullSafe();
-      if (type == Short.class) return SHORT_JSON_ADAPTER.nullSafe();
-      if (type == String.class) return STRING_JSON_ADAPTER.nullSafe();
-      if (type == Object.class) return new ObjectJsonAdapter(moshi).nullSafe();
 
       Class<?> rawType = Types.getRawType(type);
```
swankjesse: just clarifying – this part of the PR doesn’t impact behavior, does it?

moshi/src/main/java/com/squareup/moshi/ArrayJsonAdapter.java
```diff
@@ -32,7 +32,7 @@
   public static final Factory FACTORY = new Factory() {
     @Override public @Nullable JsonAdapter<?> create(
         Type type, Set<? extends Annotation> annotations, Moshi moshi) {
-      Type elementType = Types.arrayComponentType(type);
+      Type elementType = Types.arrayComponentType(Types.getRawType(type));
```
swankjesse: This doesn’t seem like it’s in the right place. I think it breaks say, `List<Number>[]`

moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.java
Syeberman: This is the core fix of the PR. It ensures `Map<String, Int>` (or whatever) in Kotlin works without having to specify `@JvmSuppressWildcards`.

moshi/src/main/java/com/squareup/moshi/ArrayJsonAdapter.java
Syeberman: All other adapters, with the exception of `ArrayJsonAdapter` and `StandardJsonAdapters`, work with `Types.getRawType(type)` rather than `type`.

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -130,9 +130,15 @@ public static Type canonicalize(Type type) {
       return new GenericArrayTypeImpl(g.getGenericComponentType());
 
     } else if (type instanceof WildcardType) {
-      if (type instanceof WildcardTypeImpl) return type;
-      WildcardType w = (WildcardType) type;
-      return new WildcardTypeImpl(w.getUpperBounds(), w.getLowerBounds());
+      Type[] upperBounds = ((WildcardType) type).getUpperBounds();
```
swankjesse: Can you leave this method as-is and create a new method to strip wildcards?

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -158,7 +159,7 @@
       Set<? extends Annotation> annotations) {
     if (annotations == null) throw new NullPointerException("annotations == null");
 
-    type = canonicalize(type);
+    type = removeSubtypeWildcard(canonicalize(type));
```
swankjesse: okay, next gotcha. This doesn’t canonicalize anymore. We don’t have a test to confirm that canonicalization is happening, and we need that too.

swankjesse: Such a test should confirm that a `GenericArrayType` of `String` shares an adapter with `String[]`.

PR #824, opened 2019-03-26
Support gradle incremental processing in code gen

### LINKED ISSUE #589: Do incremental annotation processing if possible
Should make things faster.
https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing

ZacSweers:
This adds support for incremental compilation in gradle via incap helper and marking the code gen as `ISOLATING`.

Resolves #589

### DIFF
--- a/kotlin/codegen/pom.xml
+++ b/kotlin/codegen/pom.xml
@@ -25,7 +25,15 @@
     <dependency>
       <groupId>com.squareup</groupId>
       <artifactId>kotlinpoet</artifactId>
-      <version>1.1.0</version>
+      <!-- TODO Pending a KotlinPoet release with https://github.com/square/kotlinpoet/pull/647 -->
+      <version>1.2.0-LOCAL</version>
+    </dependency>
+    <dependency>
+      <groupId>net.ltgt.gradle.incap</groupId>
+      <artifactId>incap</artifactId>
+      <version>${incap.version}</version>
+      <scope>provided</scope>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>com.google.auto</groupId>
@@ -35,7 +43,7 @@
     <dependency>
       <groupId>com.google.auto.service</groupId>
       <artifactId>auto-service</artifactId>
-      <version>1.0-rc4</version>
+      <version>${autoservice.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
@@ -101,7 +109,12 @@
                 <annotationProcessorPath>
                   <groupId>com.google.auto.service</groupId>
                   <artifactId>auto-service</artifactId>
-                  <version>1.0-rc4</version>
+                  <version>${autoservice.version}</version>
+                </annotationProcessorPath>
+                <annotationProcessorPath>
+                  <groupId>net.ltgt.gradle.incap</groupId>
+                  <artifactId>incap-processor</artifactId>
+                  <version>${incap.version}</version>
                 </annotationProcessorPath>
               </annotationProcessorPaths>
             </configuration>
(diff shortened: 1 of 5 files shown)


ZacSweers:
@shaishavgandhi05 pointed out to me that kotlin 1.3.40-dev-795 has the new incremental processing support. I'll give it a shot in a local project and try to manually verify this works with it

JakeWharton:
It's in 1.3.30 builds as well. Not the latest EAP, but newer builds.

matejdro:
Experimental processing is now included in latest stable kotlin `1.3.30`

ZacSweers:
Going to try this with Kotlin 1.3.30 today to confirm

ZacSweers:
So I've been testing this out, but after trying both 1.8.0 and a local version with this change, I think kapt actually just doesn't print out if/which processors are incremental. In light of that, I think we should move ahead with this anyway. Thoughts?

technoir42:
@ZacSweers you need `kapt.verbose=true`

I just tried this locally and it appears that `moshi-kotlin-codegen` pulls another annotation processor which is not incremental:

```
i: [kapt] Initial analysis took 36 ms
i: [kapt] Need to discovery annotation processors in the AP classpath
i: [kapt] Annotation processors: com.squareup.moshi.kotlin.codegen.JsonClassCodegenProcessor, javaslang.match.PatternsProcessor
i: [kapt] Incremental KAPT support is disabled. Processors that are not incremental: javaslang.match.PatternsProcessor.
```

Any idea where `javaslang.match.PatternsProcessor` might be coming from?

ZacSweers:
@technoir42 thanks! Are you testing that on a plain kotlin project or an android project? I was testing on an android project and still don't have any luck, but wondering if its gradle plugin is interfering with incremental apt somehow based on what I'm seeing with logs.

Regarding that processor, it appears it's part of `org.jetbrains.kotlin:kotlin-compiler-embeddable`. @swankjesse you added that in setting up the compiler testing, are it and `kotlin-annotation-processing-embeddable` just missing a `<scope>test</scope>`? If I test locally with that scope, things still pass

technoir42:
@ZacSweers I changed the scope of `kotlin-compiler-embeddable` to `test` and now incremental apt seems to work as expected. I'm testing this on android project too. Are you sure you've enabled incremental apt via `kapt.incremental.apt=true`?

ZacSweers:
Yep I've got that enabled, unsure why I'm not seeing it. Using https://github.com/ZacSweers/CatchUp as my test project. What flags are you passing to your gradle build?

JayNewstrom:
All annotation processors have to be incremental in order to proc incremental kapt. If a single processor isn't (in your case, crumb isn't, and possibly dagger incremental annotation processing isn't enabled, and others) then kapt incremental annotation processing is disabled.

ZacSweers:
Right, but I can't get _any_ logs to print out in my case :/. Also - I'm testing this on the `app` module in that project that doesn't use the crumb processor

JayNewstrom:
I added `kapt.verbose=true` to the projects `gradle.properties`, and I'm seeing logs for incremental kapt.

Logs are prefixed with `[incremental apt] `

ZacSweers:
Ahhh I've finally gotten it. It's not showing up for `[incremental apt]` nor `i: [kapt]`

I see it as 

```
[INFO] Incremental KAPT support is disabled. Processors that are not incremental: androidx.lifecycle.LifecycleProcessor, androidx.room.RoomProcessor.
```

Confirmed that Moshi does show up with current 1.8.0

```
[INFO] Incremental KAPT support is disabled. Processors that are not incremental: androidx.lifecycle.LifecycleProcessor, androidx.room.RoomProcessor, com.squareup.moshi.kotlin.codegen.JsonClassCodegenProcessor, javaslang.match.PatternsProcessor.
```

Going to go forward with the test scope change in light of that, can revisit if @swankjesse wants those compiled!

PR #827, opened 2019-03-27
Use StringBuilder to deduplicate String constants

oliverbestmann:
Inspired by https://jakewharton.com/the-economics-of-generated-code/

Not really sure if it is worth it though. My apk file size went down from 3590911 byte to 3589999 byte, the uncompressed dex file from 5276820 to 5272776 bytes.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -135,6 +135,7 @@ internal class AdapterGenerator(
     result.addFunction(generateToStringFun())
     result.addFunction(generateFromJsonFun())
     result.addFunction(generateToJsonFun())
+    result.addFunction(generateJsonDataExceptionFun())
 
     return result.build()
   }
@@ -156,6 +157,19 @@ internal class AdapterGenerator(
         .returns(String::class)
         .addStatement("return %S",
             "GeneratedJsonAdapter(${originalTypeName.rawType().simpleNames.joinToString(".")})")
+        .build()
+  }
+
+  private fun generateJsonDataExceptionFun(): FunSpec {
+    return FunSpec.builder("jsonDataException")
+        .addModifiers(KModifier.PRIVATE)
+        .returns(JsonDataException::class)
+        .addParameter("s1", String::class)
+        .addParameter("ident", String::class)
+        .addParameter("s2", String::class)
+        .addParameter("reader", JsonReader::class)
+        .addStatement("return %T(%T(s1).append(ident).append(s2).append(reader.path).toString())",
+            JsonDataException::class, StringBuilder::class)
         .build()
   }
 
@@ -185,10 +199,9 @@ internal class AdapterGenerator(
           result.addStatement("%N = %N.fromJson(%N)",
               property.localName, nameAllocator[property.delegateKey], readerParam)
         } else {
-          result.addStatement("%N = %N.fromJson(%N) ?: throw·%T(%P)",
+          result.addStatement("%N = %N.fromJson(%N) ?: throw·jsonDataException(%S, %S, %S, %N)",
               property.localName, nameAllocator[property.delegateKey], readerParam,
-              JsonDataException::class,
-              "Non-null value '${property.localName}' was null at \${${readerParam.name}.path}")
+              "Non-null value '", property.localName, "' was null at ", readerParam)
         }
         result.addStatement("%N = true", property.localIsPresentName)
         result.endControlFlow()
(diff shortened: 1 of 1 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -159,6 +160,19 @@ internal class AdapterGenerator(
         .build()
   }
 
+  private fun generateJsonDataExceptionFun(): FunSpec {
```
ZacSweers: Can we do this without the extra generated function? I'm open to the idea of this but adding a copy of this method in every class probably adds more bloat than it saves. Either extracting this to something like Util, or I bet we could get by with kotlin's standard string APIs directly

JakeWharton: The method is likely cheaper size-wise than duplicating. For performance we don't care. For method count it's unfortunate. Not sure if R8 will outline to a single method.

ZacSweers: what if we make it `inline`?

JakeWharton: Inline is only useful for library functions. We already have inline available in the code generator by either generating it twice or once in a method. If you generate it twice, R8 will probably outline it as I believe it special-cases 1-6 argument `StringBuilder` usage.

JakeWharton: https://r8.googlesource.com/r8/+/master/src/main/java/com/android/tools/r8/utils/InternalOptions.java#553

Code must be between 3 and 99 bytes and occur 20 times. Probably best to just generate it multiple times then to trigger outlining sooner.

oliverbestmann: I think it would be possible to move it to the Util class in Moshi. I am not sure how outlining works, is it only outlining within one class or would it also deduplicate this method in multiple adapter classes? Maybe if it were static?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -185,10 +199,9 @@ internal class AdapterGenerator(
           result.addStatement("%N = %N.fromJson(%N)",
               property.localName, nameAllocator[property.delegateKey], readerParam)
         } else {
-          result.addStatement("%N = %N.fromJson(%N) ?: throw·%T(%P)",
+          result.addStatement("%N = %N.fromJson(%N) ?: throw·jsonDataException(%S, %S, %S, %N)",
```
ZacSweers: What if we extract a helper method in code gen instead to generate these?

```kotlin
fun jsonDataException(description: String, identifier: String, condition: String, readerParam: ParameterSpec): CodeBlock {
  val message = CodeBlock.builder()
      .beginControlFlow("buildString")
      .addStatement("append(%S)", description)
      .addStatement("append(%S)", identifier)
      .addStatement("append(%S)", condition)
      .addStatement("append(%N.path)", readerParam)
      .endControlFlow()
      .build()
  return CodeBlock.of("%T(%L)", JsonDataException::class, message)
}
```

Or with your logic

```kotlin
fun jsonDataException(description: String, identifier: String, condition: String, readerParam: ParameterSpec): CodeBlock {
  return CodeBlock.of("%T(%T(%S).append(%S).append(%S).append(%N.path).toString())", JsonDataException::class, StringBuilder::class, description, identifier, condition, jsonReader)
}
```

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
JakeWharton: Outlining works on code patterns, not methods, so it would find the StringBuilder and 6 append statements duplicated 2*adapter count times, pull those bytecodes out into a single static method, and rewrite all adapters to use it.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
oliverbestmann: Yes, i like this idea. Tried it and seems to be working. R8 also outlines most of the code (it outlines the first part of the string concatenation `StringBuilder(a).append(b).append(c)` into one method, and it outlines the second part `sb.append(reader.getPath()).toString()` into another method. But all in all it still deduplicates the data, and the resulting dex file is even a little smaller then with the helper method.

JakeWharton:
Fixed some formatting, squashed, and merged to master as 3994723c3ce594cf0305f5b2749b02de18bac317. Thanks!

PR #837, opened 2019-04-18
Make Iso8601Utils public

karlkar:
Iso8601 is a useful util. Making it public would be a huge benefit for my project.


swankjesse:
I'd prefer you copy-paste it in; this API is not designed for direct use.

Issue #836, opened 2019-04-17
Accumulate error strategy 


stojan26:
The `KotlinJsonAdapter` fails fast when it encounters an error. This mean if there is a problem with the JSON we only get the first error. Of course the benefit of this is we don't parse additional data when we know the whole process is gonna fail.

In some cases it might be useful to accumulate the errors and fail at the end of the whole process. This means more information about everything that is wrong with the data.

Since this is a tradeoff I suggest a new adapter that accumulates errors or a setting in the existing one.

Would you interested in a PR that adds that functionality?

swankjesse:
I don‘t think we can do this well without adding significant complexity or performance costs. We might be able to tell you locally about all of the fields in a single class, but even that is awkward.

stojan26:
Good point. When I wrote this I was thinking about locally - in a single class. The way [auto-value](https://github.com/google/auto/blob/master/value/userguide/generated-builder-example.md) does it.

```
String missing = "";
      if (this.name == null) {
        missing += " name";
      }
      if (this.numberOfLegs == null) {
        missing += " numberOfLegs";
      }
      if (!missing.isEmpty()) {
        throw new IllegalStateException("Missing required properties:" + missing);
      }
```

swankjesse:
Hmm... that isn't so bad actually. Maybe for single classes the cost is low enough?

stojan26:
It should not be that hard/costly for a single class. And it could be quite useful to have all errors in one go. 

I'll try and do a prototype next week and we can go from there.

swankjesse:
Cool! Kotlin reflect, then codegen?

swankjesse:
Lets do it like this:
```
  fun fromJson(reader: JsonReader): GetterAndSetter {
    var a: Int? = null
    var b: Int? = null
    var total: Int? = null
    accumulateErrors {
      reader.beginObject()
      while (reader.hasNext()) {
        when (reader.selectName(options)) {
          0 -> a = intAdapter.fromJson(reader) ?: wasNull("a", reader)
          1 -> b = intAdapter.fromJson(reader) ?: wasNull("b", reader)
          2 -> total = intAdapter.fromJson(reader) ?: wasNull("total", reader)
          -1 -> reader.skipNameAndValue()
        }
      }
      reader.endObject()
      require("a", a)
      require("b", b)
    }
    var result = GetterAndSetter(
        a!!,
        b!!
    )
    result.total = total ?: result.total
    return result
  }
```

with this support:

```
[...]

PR #840, opened 2019-04-25
[WIP] Add AccumulateErrors failure strategy.

LordRaydenMK:
As discussed in #836 this PR adds the opt-in AccumulateErrors strategy.

Let me know if I am going in the right direction. 

Thanks for considering it.

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -56,11 +56,14 @@ private val ABSENT_VALUE = Any()
 internal class KotlinJsonAdapter<T>(
     val constructor: KFunction<T>,
     val bindings: List<Binding<T, Any?>?>,
-    val options: JsonReader.Options) : JsonAdapter<T>() {
+    val options: JsonReader.Options,
+    val failureStrategy: FailureStrategy) : JsonAdapter<T>() {
 
   override fun fromJson(reader: JsonReader): T {
     val constructorSize = constructor.parameters.size
 
+    // accumulator for Non-null values that are null or required fields that are missing
+    var failures: MutableList<String>? = null
     // Read each value into its slot in the array.
     val values = Array<Any?>(bindings.size) { ABSENT_VALUE }
     reader.beginObject()
@@ -82,21 +85,37 @@ internal class KotlinJsonAdapter<T>(
       values[index] = binding.adapter.fromJson(reader)
 
       if (values[index] == null && !binding.property.returnType.isMarkedNullable) {
-        throw JsonDataException(
-            "Non-null value '${binding.property.name}' was null at ${reader.path}")
+        when(failureStrategy) {
+            FailureStrategy.FAIL_FAST -> throw JsonDataException(
+                    "Non-null value '${binding.property.name}' was null at ${reader.path}")
+            FailureStrategy.ACCUMULATE_ERRORS -> {
+              if (failures == null) failures = mutableListOf()
+              failures.add("'${binding.property.name}' was null at ${reader.path}")
+            }
+        }
       }
     }
     reader.endObject()
+
 
     // Confirm all parameters are present, optional, or nullable.
     for (i in 0 until constructorSize) {
       if (values[i] === ABSENT_VALUE && !constructor.parameters[i].isOptional) {
         if (!constructor.parameters[i].type.isMarkedNullable) {
-          throw JsonDataException(
-              "Required value '${constructor.parameters[i].name}' missing at ${reader.path}")
+          when(failureStrategy){
+            FailureStrategy.FAIL_FAST -> throw JsonDataException(
+                    "Required value '${constructor.parameters[i].name}' missing at ${reader.path}")
+            FailureStrategy.ACCUMULATE_ERRORS -> {
+              if (failures == null) failures = mutableListOf()
+              failures.add("'${constructor.parameters[i].name}' missing at ${reader.path}")
+            }
+          }
         }
         values[i] = null // Replace absent with null.
       }
+    }
+    if (failureStrategy == FailureStrategy.ACCUMULATE_ERRORS && !failures.isNullOrEmpty()) {
+      throw JsonDataException("Required value(s) ${failures.joinToString()}")
     }
 
     // Call the constructor using a Map so that absent optionals get defaults.
@@ -164,7 +183,25 @@ internal class KotlinJsonAdapter<T>(
   }
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+/**
+ * Failure strategy when a required value is missing or non-null value is null
+ */
+enum class FailureStrategy {
+
+    /**
+     * Throw a [JsonDataException] on the first violation
+     */
+    FAIL_FAST,
+
+    /**
+     * Accumulate all the violations and throw a [JsonDataException] at the end
+     */
+    ACCUMULATE_ERRORS
+}
+
+class KotlinJsonAdapterFactory @JvmOverloads constructor(
+    val failureStrategy: FailureStrategy = FailureStrategy.FAIL_FAST) : JsonAdapter.Factory {
+
   override fun create(type: Type, annotations: MutableSet<out Annotation>, moshi: Moshi)
       : JsonAdapter<*>? {
     if (!annotations.isEmpty()) return null
@@ -237,8 +274,14 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
       val name = jsonAnnotation?.name ?: property.name
       val resolvedPropertyType = resolve(type, rawType, property.returnType.javaType)
-      val adapter = moshi.adapter<Any>(
-          resolvedPropertyType, Util.jsonAnnotations(allAnnotations.toTypedArray()), property.name)
+      val adapter = if(failureStrategy == FailureStrategy.ACCUMULATE_ERRORS &&
+              !property.returnType.isMarkedNullable) {
+        // Use nullSafe() so it accumulates null violations
+        moshi.adapter<Any>(resolvedPropertyType, Util.jsonAnnotations(allAnnotations.toTypedArray()), property.name)
+                .nullSafe()
+      } else {
+        moshi.adapter<Any>(resolvedPropertyType, Util.jsonAnnotations(allAnnotations.toTypedArray()), property.name)
+      }
 
       bindingsByName[property.name] = KotlinJsonAdapter.Binding(name, adapter,
           property as KProperty1<Any, Any?>, parameter)
@@ -257,6 +300,6 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
     bindings += bindingsByName.values
 
     val options = JsonReader.Options.of(*bindings.map { it?.name ?: "\u0000" }.toTypedArray())
-    return KotlinJsonAdapter(constructor, bindings, options).nullSafe()
+    return KotlinJsonAdapter(constructor, bindings, options, failureStrategy).nullSafe()
   }
 }
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -145,6 +145,20 @@ class KotlinJsonAdapterTest {
 
   class RequiredValueAbsent(var a: Int = 3, var b: Int)
 
+  @Test fun multipleRequiredValuesAreAbsentWhenAccumulateErrors() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory(FailureStrategy.ACCUMULATE_ERRORS)).build()
+    val jsonAdapter = moshi.adapter(MultipleRequiredValuesAbsent::class.java)
+
+    try {
+      jsonAdapter.fromJson("""{"c":5}""")
+      fail()
+    } catch (expected: JsonDataException) {
+      assertThat(expected).hasMessage("Required value(s) 'a' missing at $, 'd' missing at $")
+    }
+  }
+
+  class MultipleRequiredValuesAbsent(var a: Int, var b: Int = 42, var c: Int, var d: Int, var e: Int?)
+
   @Test fun nonNullConstructorParameterCalledWithNullFailsWithJsonDataException() {
     val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
     val jsonAdapter = moshi.adapter(HasNonNullConstructorParameter::class.java)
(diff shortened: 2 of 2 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -88,16 +89,25 @@ internal class KotlinJsonAdapter<T>(
     }
     reader.endObject()
 
+    val missing = mutableListOf<String>()
```
JakeWharton: This should be lazily allocated

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +174,12 @@ internal class KotlinJsonAdapter<T>(
   }
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+enum class FailureStrategy {
+  FAIL_FAST, ACCUMULATE_ERRORS
+}
+
+class KotlinJsonAdapterFactory(val failureStrategy: FailureStrategy = FailureStrategy.FAIL_FAST) : JsonAdapter.Factory {
```
JakeWharton: Need `@JvmOverloads`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +174,12 @@ internal class KotlinJsonAdapter<T>(
   }
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+enum class FailureStrategy {
+  FAIL_FAST, ACCUMULATE_ERRORS
```
JakeWharton: While the names are fairly clear, I would still document the behavior on these values.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -88,16 +89,25 @@ internal class KotlinJsonAdapter<T>(
...
+            FailureStrategy.FAIL_FAST -> throw JsonDataException(
+                    "Required value '${constructor.parameters[i].name}' missing at ${reader.path}")
+            FailureStrategy.ACCUMULATE_ERRORS -> missing += "'${constructor.parameters[i].name}'"
+          }
         }
         values[i] = null // Replace absent with null.
       }
     }
+    if(failureStrategy == FailureStrategy.ACCUMULATE_ERRORS) {
+      if (missing.isNotEmpty()) {
```
JakeWharton: Merge nested `if`s

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -79,25 +82,38 @@ internal class KotlinJsonAdapter<T>(
             "Multiple values for '${binding.property.name}' at ${reader.path}")
       }
 
-      values[index] = binding.adapter.fromJson(reader)
+      values[index] = when (failureStrategy) {
+        FailureStrategy.FAIL_FAST -> binding.adapter.fromJson(reader)
+        FailureStrategy.ACCUMULATE_ERRORS -> binding.adapter.nullSafe().fromJson(reader)
```
stojan26: `nullSafe()` because otherwise it throws `JsonDataException` on the first null

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -79,25 +82,38 @@ internal class KotlinJsonAdapter<T>(
...
+            FailureStrategy.FAIL_FAST -> throw JsonDataException(
+                    "Required value '${constructor.parameters[i].name}' missing at ${reader.path}")
+            FailureStrategy.ACCUMULATE_ERRORS -> failures!!.add("'${constructor.parameters[i].name}'")
+          }
         }
         values[i] = null // Replace absent with null.
       }
     }
+    if (failureStrategy == FailureStrategy.ACCUMULATE_ERRORS && failures!!.isNotEmpty()) {
+      throw JsonDataException("Required value(s) ${failures.joinToString(separator = " ")} missing or null at ${reader.path}")
```
stojan26: Non-nullable null values and missing fields in the same exception. Not sure if it's worth it to separate them?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -56,11 +56,14 @@ private val ABSENT_VALUE = Any()
...
     val bindings: List<Binding<T, Any?>?>,
-    val options: JsonReader.Options) : JsonAdapter<T>() {
+    val options: JsonReader.Options,
+    val failureStrategy: FailureStrategy) : JsonAdapter<T>() {
 
   override fun fromJson(reader: JsonReader): T {
     val constructorSize = constructor.parameters.size
 
+    // accumulator for Non-null values that are null or required fields that are missing
+    val failures = if(failureStrategy == FailureStrategy.ACCUMULATE_ERRORS) mutableListOf<String>() else null
```
JakeWharton: This should _always_ be initialized to `null` and lazily instantiated only if an error occurs.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
JakeWharton: This is going to allocate every time it's called, which is going to be a lot. This needs to be done at binding creation.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
JakeWharton: Put the cause of the failure into the message list and remove "missing or null" from here

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
LordRaydenMK: is `nullSafe()` significantly worse than the other version? This was always created here, I just added the `when`.

JakeWharton: What other version? The original version just looks up the adapter instance on a field.

LordRaydenMK: Old code:
`values[index] = binding.adapter.fromJson(reader)`

new code:

```
values[index] = when (failureStrategy) {
        FailureStrategy.FAIL_FAST -> binding.adapter.fromJson(reader)
        FailureStrategy.ACCUMULATE_ERRORS -> binding.adapter.nullSafe().fromJson(reader)
}
```

is `binding.adapter.nullSafe().fromJson(reader)` significantly worse than `binding.adapter.fromJson(reader)` ?

> This is going to allocate every time it's called, which is going to be a lot.

is that the allocation you are referring to?

JakeWharton: Yes

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -237,8 +274,14 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
       val name = jsonAnnotation?.name ?: property.name
       val resolvedPropertyType = resolve(type, rawType, property.returnType.javaType)
-      val adapter = moshi.adapter<Any>(
-          resolvedPropertyType, Util.jsonAnnotations(allAnnotations.toTypedArray()), property.name)
+      val adapter = if(failureStrategy == FailureStrategy.ACCUMULATE_ERRORS &&
+              !property.returnType.isMarkedNullable) {
```
LordRaydenMK: when `returnType.isMarkedNullable` is true the adapter is already `nullSafe()`

if there is a better way to do this let me know and I will update the PR.

LordRaydenMK:
Thanks for the review Jake. I implemented the changes you requested as best as I could. If there is anything else I can do to improve the PR let me know and I am happy to continue improving it. 

If the code here looks OK I would like to continue and implement the same functionality for codegen.

Is there anything else I should do here or should I start working on codegen?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +183,25 @@ internal class KotlinJsonAdapter<T>(
   }
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+/**
+ * Failure strategy when a required value is missing or non-null value is null
+ */
+enum class FailureStrategy {
```
ZacSweers: We may want to extract this to the main moshi artifact for shared use in code gen

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -82,22 +85,38 @@ internal class KotlinJsonAdapter<T>(
       values[index] = binding.adapter.fromJson(reader)
 
       if (values[index] == null && !binding.property.returnType.isMarkedNullable) {
-        throw JsonDataException(
-            "Non-null value '${binding.property.name}' was null at ${reader.path}")
+        when(failureStrategy) {
```
ZacSweers: nit: space after `when`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -82,22 +85,38 @@ internal class KotlinJsonAdapter<T>(
...
+            FailureStrategy.FAIL_FAST -> throw JsonDataException(
+                    "Non-null value '${binding.property.name}' was null at ${reader.path}")
+            FailureStrategy.ACCUMULATE_ERRORS -> {
+              if (failures == null) failures = mutableListOf()
```
ZacSweers: Let's make this a local reusable inline function

```kotlin
inline fun addFailure(name: String) {
  if (failures == null) failures = mutableListOf()
  failures.add("'$name' was null at ${reader.path}")
}

// later
addFailure(binding.property.name)
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -82,22 +85,38 @@ internal class KotlinJsonAdapter<T>(
...
     reader.endObject()
 
+
     // Confirm all parameters are present, optional, or nullable.
     for (i in 0 until constructorSize) {
       if (values[i] === ABSENT_VALUE && !constructor.parameters[i].isOptional) {
         if (!constructor.parameters[i].type.isMarkedNullable) {
-          throw JsonDataException(
-              "Required value '${constructor.parameters[i].name}' missing at ${reader.path}")
+          when(failureStrategy){
```
ZacSweers: `when (failureStrategy) {`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +183,25 @@ internal class KotlinJsonAdapter<T>(
...
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+/**
+ * Failure strategy when a required value is missing or non-null value is null
+ */
+enum class FailureStrategy {
+
+    /**
+     * Throw a [JsonDataException] on the first violation
```
ZacSweers: end sentences in periods. Here and the other docs

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +183,25 @@ internal class KotlinJsonAdapter<T>(
...
+     */
+    FAIL_FAST,
+
+    /**
+     * Accumulate all the violations and throw a [JsonDataException] at the end
+     */
+    ACCUMULATE_ERRORS
+}
+
+class KotlinJsonAdapterFactory @JvmOverloads constructor(
```
ZacSweers: applying this to all classes might be hard to replicate in code gen. I wonder if it'd be worth putting this in `@JsonClass` as a property instead and make them all opt-in.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -237,8 +274,14 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
       val name = jsonAnnotation?.name ?: property.name
       val resolvedPropertyType = resolve(type, rawType, property.returnType.javaType)
-      val adapter = moshi.adapter<Any>(
-          resolvedPropertyType, Util.jsonAnnotations(allAnnotations.toTypedArray()), property.name)
+      val adapter = if(failureStrategy == FailureStrategy.ACCUMULATE_ERRORS &&
```
ZacSweers: `if (failureStrategy)`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -164,7 +183,25 @@ internal class KotlinJsonAdapter<T>(
...
+    FAIL_FAST,
+
+    /**
+     * Accumulate all the violations and throw a [JsonDataException] at the end
+     */
+    ACCUMULATE_ERRORS
+}
+
+class KotlinJsonAdapterFactory @JvmOverloads constructor(
+    val failureStrategy: FailureStrategy = FailureStrategy.FAIL_FAST) : JsonAdapter.Factory {
```
ZacSweers: nit: put the closing brace + superclass on next line

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: I'm not quite sure I understand (but I'm late to this PR so I might be missing prior context). Can you elaborate a bit more?

swankjesse:
I’d like to close this PR and find another strategy to address its goals.

Error messages should cost nearly nothing because they are rare. Like, you’ll do 10 million successful parses for every field error, and 100 million successful parses for every occurrence of multiple field errors. Adding code to detect and handle multiple errors is a bad trade-off, because that code has a cost to be loaded into the CPU etc.

LordRaydenMK:
Thank you all for taking the time to review and considering this PR. I'll close and let's see if we can find another way to achieve accumulating errors.

PR #847, opened 2019-05-09
Allow custom generators

ZacSweers:
This implements support for custom generators via `@JsonClass(generator = "")`. The goal is to allow custom generators to plug in to the same adapter lookup that Moshi supports for its own code gen.

Example:

```kotlin
package example

@JsonClass(generateAdapter = true, generator = "custom")
data class Taco(val seasoning: String)
```

In this event, Moshi codegen will skip generation of a `example.Taco` adapter, but at runtime will expect to find one at `example.TacoJsonAdapter` (via extracted `Types#generatedJsonAdapterName()` API)

This opens up a few opportunities:
- Custom generators. 3rd party JsonAdapter-generating libraries like auto-value-moshi, kotshi, etc could hook into this to allow for zero-configuration usage.
- Custom aggregators could generate reflection-less adapters via their own processors, similar to how auto-value-moshi's factory processor works.
- More lenient constructor lookup by not requiring a Moshi parameter, possibly allowing for future leaner code gen for native JsonReader types (i.e. could use APIs like `nextString()` and possibly avoid needing the Moshi parameter entirely). This also allows custom generators to do the same.

Also opportunistically extracted `NullSafeJsonAdapter` to `.internal` for easier delegate testing

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
@@ -84,7 +84,7 @@ class JsonClassCodegenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
   override fun process(annotations: Set<TypeElement>, roundEnv: RoundEnvironment): Boolean {
     for (type in roundEnv.getElementsAnnotatedWith(annotation)) {
       val jsonClass = type.getAnnotation(annotation)
-      if (jsonClass.generateAdapter) {
+      if (jsonClass.generateAdapter && jsonClass.generator.isEmpty()) {
         val generator = adapterGenerator(type) ?: continue
         generator.generateAndWrite(generatedType)
       }
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest.kt
@@ -26,6 +26,7 @@ import com.squareup.moshi.JsonWriter
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.ToJson
 import com.squareup.moshi.Types
+import com.squareup.moshi.internal.NullSafeJsonAdapter
 import org.assertj.core.api.Assertions.assertThat
 import org.intellij.lang.annotations.Language
 import org.junit.Assert.assertNull
@@ -1125,6 +1126,29 @@ class GeneratedAdaptersTest {
 
     val decoded = adapter.fromJson("""{"listOfInts":[4,-5,6]}""")!!
     assertThat(decoded).isEqualTo(HasCollectionOfPrimitives(listOf(4, -5, 6)))
+  }
+
+  @JsonClass(generateAdapter = true, generator = "custom")
+  data class CustomGeneratedClass(val foo: String)
+
+  @Test fun customGenerator_withClassPresent() {
+    val moshi = Moshi.Builder().build()
+    val adapter = moshi.adapter(CustomGeneratedClass::class.java)
+    val unwrapped = (adapter as NullSafeJsonAdapter<CustomGeneratedClass>).delegate()
+    assertThat(unwrapped).isInstanceOf(GeneratedAdaptersTest_CustomGeneratedClassJsonAdapter::class.java)
+  }
+
+  @JsonClass(generateAdapter = true, generator = "custom")
+  data class CustomGeneratedClassMissing(val foo: String)
+
+  @Test fun customGenerator_withClassMissing() {
+    val moshi = Moshi.Builder().build()
+    try {
+      moshi.adapter(CustomGeneratedClassMissing::class.java)
+      fail()
+    } catch (e: RuntimeException) {
+      assertThat(e).hasMessageContaining("Failed to find the generated JsonAdapter class")
+    }
   }
 }
 
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest.kt
@@ -1190,37 +1190,36 @@ data class NullableTypeParams<T>(
     val nonNullT: T
 )
 
-  /**
-   * This is here mostly just to ensure it still compiles. Covers variance, @Json, default values,
-   * nullability, primitive arrays, and some wacky generics.
-   */
-  @JsonClass(generateAdapter = true)
-  data class SmokeTestType(
-    @Json(name = "first_name") val firstName: String,
-    @Json(name = "last_name") val lastName: String,
-    val age: Int,
-    val nationalities: List<String> = emptyList(),
-    val weight: Float,
-    val tattoos: Boolean = false,
-    val race: String?,
-    val hasChildren: Boolean = false,
-    val favoriteFood: String? = null,
-    val favoriteDrink: String? = "Water",
-    val wildcardOut: MutableList<out String> = mutableListOf(),
-    val nullableWildcardOut: MutableList<out String?> = mutableListOf(),
-    val wildcardIn: Array<in String>,
-    val any: List<*>,
-    val anyTwo: List<Any>,
-    val anyOut: MutableList<out Any>,
-    val nullableAnyOut: MutableList<out Any?>,
-    val favoriteThreeNumbers: IntArray,
-    val favoriteArrayValues: Array<String>,
-    val favoriteNullableArrayValues: Array<String?>,
-    val nullableSetListMapArrayNullableIntWithDefault: Set<List<Map<String, Array<IntArray?>>>>? = null,
-    val aliasedName: TypeAliasName = "Woah",
-    val genericAlias: GenericTypeAlias = listOf("Woah")
-  )
-}
+/**
+ * This is here mostly just to ensure it still compiles. Covers variance, @Json, default values,
+ * nullability, primitive arrays, and some wacky generics.
+ */
+@JsonClass(generateAdapter = true)
+data class SmokeTestType(
+  @Json(name = "first_name") val firstName: String,
+  @Json(name = "last_name") val lastName: String,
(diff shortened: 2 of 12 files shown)


ZacSweers:
I don't understand the merge conflict, my branch is up-to-date 🤔 

image

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest.kt
```diff
@@ -1155,6 +1156,40 @@ class GeneratedAdaptersTest {
...
+      moshi.adapter(CustomGeneratedClassMissing::class.java)
+      fail()
+    } catch (e: RuntimeException) {
+      assertThat(e).hasMessageContaining("Failed to find the generated JsonAdapter class")
+    }
+  }
+}
+
+// Has to be outside to avoid Types seeing an owning class
+@JsonClass(generateAdapter = true)
```
ZacSweers: I don't know why this shows up as changed, I didn't change it in the branch but it shows up here after rebasing in the github GUI since I didn't have any conflicts locally

moshi/src/main/java/com/squareup/moshi/JsonClass.java
```diff
@@ -29,13 +30,44 @@
...
+   *   </li>
+   * </ul>
+   *
+   * <p>Example for a class "CustomType":<pre>{@code
+   *   class CustomTypeJsonAdapter(moshi: Moshi, types: Array<Type>) : JsonAdapter<CustomType>() {
+   *     // ...
+   *   }
+   * }</pre>
+   */
+  String generator() default "";
```
swankjesse: One thing I’m sort of annoyed with myself on is the word _generator_ here and above.

We conflated two concepts:
 - whether to look up an adapter by naming convention
 - whether to generate an adapter

That said, I think we might just document that “generator = manual” is completely legit.

moshi/src/main/java/com/squareup/moshi/Types.java
```diff
@@ -49,6 +49,35 @@
...
   }
 
+  /**
+   * Resolves the generated {@link JsonAdapter} fully qualified class name for a given
+   * {@link JsonClass JsonClass-annotated} {@code clazz}. This is the same lookup logic used by
+   * both the Moshi code generation as well as lookup for any JsonClass-annotated classes. This can
+   * be useful if generating your own JsonAdapters without using Moshi's first party code gen.
+   *
+   * @param clazz the class to calculate a generated JsonAdapter name for.
+   * @return the resolved fully qualified class name to the expected generated JsonAdapter class.
```
swankjesse: Can you add a note that the returned class name will always be a top-level class, not a nested class? That way the interpretation of `.` is unambiguous.

moshi/src/main/java/com/squareup/moshi/internal/NullSafeJsonAdapter.java
```diff
@@ -0,0 +1,40 @@
+package com.squareup.moshi.internal;
```
swankjesse: nit: copyright

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -460,23 +460,35 @@ public static String typeAnnotatedWithAnnotations(Type type,
...
-            = adapterClass.getDeclaredConstructor(Moshi.class, Type[].class);
-        constructor.setAccessible(true);
-        return constructor.newInstance(moshi, ((ParameterizedType) type).getActualTypeArguments())
-            .nullSafe();
+        Type[] typeArgs = ((ParameterizedType) type).getActualTypeArguments();
+        try {
+          // Common case first
+          constructor = adapterClass.getDeclaredConstructor(Moshi.class, Type[].class);
+          args = new Object[] { moshi, typeArgs };
+        } catch (NoSuchMethodException e) {
```
swankjesse: do we really need to fall back here? Requiring a Moshi parameter is good?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
```diff
@@ -85,7 +85,7 @@ class JsonClassCodegenProcessor : KotlinAbstractProcessor(), KotlinMetadataUtils
   override fun process(annotations: Set<TypeElement>, roundEnv: RoundEnvironment): Boolean {
     for (type in roundEnv.getElementsAnnotatedWith(annotation)) {
       val jsonClass = type.getAnnotation(annotation)
-      if (jsonClass.generateAdapter) {
+      if (jsonClass.generateAdapter && jsonClass.generator.isEmpty()) {
```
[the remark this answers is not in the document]
ZacSweers: That's a good question. I _think_ it's technically a nonbreaking change since it has a default value, but new props without a default would break

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: Not every adapter necessarily needs it, such as ones that work entirely with builtins APIs like `nextString()` or `value(String)`. Then it ends up requiring consumers to accept a parameter they don't use

moshi/src/main/java/com/squareup/moshi/JsonClass.java
```diff
@@ -29,13 +30,44 @@
...
+   * other custom code generation tools to run and still allow Moshi to read their generated
+   * JsonAdapter outputs.
+   *
+   * <p>Requirements for :
+   * <ul>
+   *   <li>
+   *     The generated adapter must subclass {@link JsonAdapter} and be parameterized by this type.
+   *   </li>
+   *   <li>
+   *     {@link Types#generatedJsonAdapterName} should be used for the fully qualified class name in
```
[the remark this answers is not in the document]
ZacSweers: What kind of detail would you want to add here?

moshi/src/main/java/com/squareup/moshi/JsonClass.java
ZacSweers: Yeah I wasn't sure what to call this. `generatorTag` felt too long and `tag` felt too vague.

What would that documentation look like?

moshi/src/main/java/com/squareup/moshi/internal/NullSafeJsonAdapter.java
ZacSweers: part of b586ac0

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest_CustomGeneratedClassJsonAdapter.kt
```diff
@@ -0,0 +1,17 @@
+package com.squareup.moshi.kotlin.codgen
```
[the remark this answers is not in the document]
ZacSweers: part of b586ac0

moshi/src/main/java/com/squareup/moshi/Types.java
ZacSweers: f9fd709

moshi/src/main/java/com/squareup/moshi/JsonClass.java
```diff
@@ -29,13 +30,44 @@
...
+   *     order for Moshi to correctly resolve and load the generated JsonAdapter.
+   *   </li>
+   *   <li>The first parameter must be a {@link Moshi} instance.</li>
+   *   <li>
+   *     If generic, a second {@link Type[]} parameter should be declared to accept type arguments.
```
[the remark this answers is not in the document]
ZacSweers: 9ebd467

moshi/src/main/java/com/squareup/moshi/JsonClass.java
```diff
@@ -29,13 +30,44 @@
...
   boolean generateAdapter();
+
+  /**
+   * An optional custom generator tag used to indicate which generator should be used. If empty,
+   * Moshi's annotation processor will generate an adapter for the annotated type. If not empty,
+   * Moshi's processor will skip it and defer to a custom generator. This can be used to allow
+   * other custom code generation tools to run and still allow Moshi to read their generated
+   * JsonAdapter outputs.
+   *
+   * <p>Requirements for :
```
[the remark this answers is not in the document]
ZacSweers: I just like to keep the reader in suspense with cliffhangers

... bcc9774

moshi/src/main/java/com/squareup/moshi/JsonClass.java
```diff
@@ -29,13 +30,44 @@
...
-   * There are currently some restrictions on which types that can be used with generated adapters:
-   *
-   *  * The class must be implemented in Kotlin.
-   *  * The class may not be an abstract class, an inner class, or a local class.
-   *  * All superclasses must be implemented in Kotlin.
-   *  * All properties must be public, protected, or internal.
-   *  * All properties must be either non-transient or have a default value.
+   * <p>There are currently some restrictions on which types that can be used with generated adapters:
+   * <ul>
+   *  <li>The class must be implemented in Kotlin.</li>
```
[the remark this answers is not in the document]
ZacSweers: e107a73

moshi/src/main/java/com/squareup/moshi/JsonClass.java
swankjesse: Just something in the `@JsonAdapter` KDoc that says “your generator doesn’t have to be a program.”

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codgen/GeneratedAdaptersTest_CustomGeneratedClassJsonAdapter.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package com.squareup.moshi.kotlin.codgen
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.kotlin.codgen.GeneratedAdaptersTest.CustomGeneratedClass
+
+// This also tests custom generated types with no moshi constructor
```
swankjesse: Nice.

PR #856, opened 2019-05-28
removed limitation where subtypes should be unique

manthesepeasrock:
There can be use cases where different type labels should match with the same subtype


JakeWharton:
Please add a test for this behavior.

timic:
+1

ZacSweers:
Thanks!

Issue #866, opened 2019-06-18
Add adapter for Void


onacit:
I have a case for handling `Void`.
```java
class Parent<IdType> {
    private IdType id;
}
class Child extends Parent<Void> {
}
```
And I have to add my custom `JsonAapter` for that.
```java
static class MoshiVoidAdapter extends JsonAdapter<Void> {

    @Override
    public Void fromJson(final JsonReader reader) throws IOException {
        return reader.nextNull();
    }

    @Override
    public void toJson(final JsonWriter writer, final Void value) throws IOException {
        writer.nullValue();
    }
}
```
It'll be helpful if a predefined adapter provided.

FYI. no luck with `com.squareup.moshi:moshi-adapters`.

swankjesse:
Your adapter looks correct!

swankjesse:
I’m sorry, I didn’t intend to say “send us a pull request”. I think the adapter above is fine, and I think anyone else who encounters a Void should write an adapter.

Moshi’s goal isn’t to provide all the adapters in the box, instead it’s to make writing adapters really easy.

onacit:
@swankjesse Thank you the comment.

Can I raise another issue for following problem?

```
java.lang.IllegalArgumentException: 
Platform class java.lang.Void (with no annotations) requires explicit JsonAdapter to be registered
for class java.lang.Void id
for class ....HandleMessageNotification
        at ....HandleMessageNotificationTest.handleMessage_02_notification(HandleMessageNotificationTest.java:62)
Caused by: java.lang.IllegalArgumentException: Platform class java.lang.Void (with no annotations) requires explicit JsonAdapter to be registered
        at ....HandleMessageNotificationTest.handleMessage_02_notification(HandleMessageNotificationTest.java:62)
```

It will be much more helpful if Moshi, internally, treats `Void` more handy and automatically.

onacit:
For anyone reached here.

```java
    static class VoidAdapter extends JsonAdapter<Void> {

        @Override
        public Void fromJson(final JsonReader reader) throws IOException {
            reader.skipValue();
            return null;
        }

        @Override
        public void toJson(final JsonWriter writer, final Void value) throws IOException {
            writer.nullValue();
        }
    }
```

PR #870, opened 2019-06-19
[#866] Add VoidAdapter

onacit:
#866 
The second PR after CheckStyle related problems fixed.


swankjesse:
I don’t think this is a good fit. Reading and writing `Void` as data is really weird.

PR #875, opened 2019-07-02
Kotlin-reflect QoL APIs

ZacSweers:
This implements two helper APIs that leverage reified types + `typeOf()` to simplify adapter lookup and builder construction.

Note that this requires a manual implementation of `KType.toType()` as the stdlib's `javaType` reflect property isn't actually implemented.

One test fails because of a mismatched error message, which actually revealed a bug that I'll PR separately to fix.

An interesting point of discussion is how to handle the `@ExperimentalStdLibApi` annotation. I think it's fair to expose it with that annotation transitively applied, and basically not support it until it's graduated from experimental

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/-MoshiKotlinExtensions.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/-MoshiKotlinExtensions.kt
@@ -19,3 +19,6 @@ inline fun <reified T> Moshi.adapter(): JsonAdapter<T> {
     adapter.nonNull()
   }
 }
+
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.Builder.add(adapter: JsonAdapter<T>) = add(typeOf<T>().javaType, adapter)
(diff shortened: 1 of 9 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/-MoshiKotlinExtensions.kt
```diff
@@ -0,0 +1,89 @@
...
+ * @return a [JsonAdapter] for [T], creating it if necessary. Note that while nullability of [T]
+ *         itself is handled, nested types (such as in generics) are not resolved.
+ */
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.adapter(): JsonAdapter<T> {
+  return adapter(typeOf<T>())
+}
+
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.Builder.addAdapter(adapter: JsonAdapter<T>) = add(typeOf<T>().toType(), adapter)
```
ZacSweers: @rharter @swankjesse Thought experiment - if `T` is nullable, should `.nullSafe()` be called on the adapter? and/or vice-versa with `.nonNull()`?

rharter: I'm curious about the use case there.  I don't use the nullSafe variants much since I lean on the properties to handle that.  Would the user have to do this if the class can be used in different contexts with different nullability?

```
Moshi.Builder()
  .addAdapter(JsonAdapter<Foo?>())
  .addAdapter(JsonAdapter<Foo>())
```

ZacSweers: hmm, yeah. And moshi wouldn't be able to tell the difference between those two because they're both just assigned to `Foo`, so I think we have to just leave it at that for now

PR #876, opened 2019-07-02
Pass raw type rather than possibly-generic type in platform type error

ZacSweers:
We only check the raw type here, so the message should match

Example case - in #875 a `Triple<*, *, *>` would yield an error message that `kotlin.Triple<java.lang.Object, java.lang.Object, java.lang.Object>` required an adapter, which doesn't tell you immediately which type it was confused on

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -57,7 +57,7 @@ final class ClassJsonAdapter<T> extends JsonAdapter<T> {
       if (!annotations.isEmpty()) return null;
       if (Util.isPlatformType(rawType)) {
         throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -56,8 +56,12 @@
       if (rawType.isInterface() || rawType.isEnum()) return null;
       if (!annotations.isEmpty()) return null;
       if (Util.isPlatformType(rawType)) {
+        String messagePrefix = "Platform " + rawType;
+        if (type instanceof ParameterizedType) {
+          messagePrefix += " in " + type;
```
JakeWharton: Fun fact: this will produce three StringBuilders. No wait that's an unfun fact. The actual fun fact is that R8 1.6 (in AGP 3.6) will merge these into a single StringBuilder!

(Not that we care about three StringBuilders in an error branch anyway...)

Issue #734, opened 2018-11-05
@Nullable annotation and Proguard


AndroidBunny:
We implemented custom adapters to avoid null values as following:

```
private class DefaultStringAdapter {
    @FromJson
    fun fromJson(@Nullable jsonString: String?) = jsonString ?: ""
    @ToJson
    fun toJson(jsonString: String?) = jsonString
} 
```

Everything works like expected, but unfortunately it looks like when proguarding the annotation gets removed and therefore the adapter isn't used, what leads to an 

com.d.a.j: Non-null value 'msStation' was null at $.list[0].header.ms_station

Proguard rules defined as suggested in the README.

NightlyNexus:
what is the Nullable annotation? does it have runtime retention?

AndroidBunny:
It's javax.annotation.Nullable - RetentionPolicy.RUNTIME

AndroidBunny:
Any ideas on how to make this work?

NightlyNexus:
Can you make a reproducible sample?

AndroidBunny:
Yes, I uploaded a very similar project here: https://github.com/AndroidBunny/MoshiAnnotation

You can see the difference by building DevDebug and DevProDebug. The first one runs how it should, the second one throws "com.c.a.j: Non-null value 'locationInfo' was null at $.list[1].header.location_info"

The Annotation is used in the MoshiModule.

AndroidBunny:
Any updates on this one?

ZacSweers:
Sorry for the delay! You need to add a proguard rule to keep that javax annotation, as proguard will remove it if it deems it as unused even if its retention policy is `RUNTIME`.

```proguard
-keep @interface javax.annotation.Nullable
```

mdeandrea-mrmilu:
This still a issue? If not, please close

Issue #454, opened 2018-02-26
Moshi-kotlin doesn't respect @field:Json annotations in JUnit tests


premnirmal:
Using moshi and moshi-kotlin 1.5.

When using moshi kotlin, the kotlin data class constructor variables are to be marked with `@field:Json` for them to be recognized correctly by the reflection library on obfuscated builds. I noticed that in my unit tests, this doesn't work. In unit tests, the only way to get moshi to recognize the field and parse it is to annotate it as `@Json` (notice that the `@field` prefix isn't added).

This means that my models must have both the annotations `@field:Json` and `@Json`.

Example:

Moshi instance:
``` java
return new Moshi.Builder()
                .add(new KotlinJsonAdapterFactory())
                .build();
```

User class:
``` kotlin
data class User(
        @field:Json(name = "user")
        var inner: UserInner? = null)
```

Parsed in app code (obfuscated):

``` java
User user = moshi.adapter(User.class).fromJson(jsonString);
Log.d("User", user.getInner().toString());
// user.getInner() is not null
```

Fails when parsed in unit tests:

``` java
@RunWith(JUnit4.class)
class UserTest extends junit.framework.TestCase {
    @Test
    public void testUser() {
          User user = moshi.adapter(User.class).fromJson(jsonString);
          assertNotNull(user.getInner()); // fails!
     }
}
```

When I update the user class to **include both annotations:**
``` kotlin
data class User(
        @Json(name = "user") @field:Json(name = "user")
        var inner: UserInner? = null)
```
**Now the tests pass.**

My unit tests are standard JUnit4 unit tests written in java:

premnirmal:
Edit:

The json adapter doesn't find the annotation because the jsonAnnotation is `null` when accessed here in this line:
https://github.com/square/moshi/blob/master/kotlin/src/main/java/com/squareup/moshi/kotlin/KotlinJsonAdapter.kt#L223

However, when running on jdk 1.7 on an android device, the annotation is present.
Only in unit tests (jdk 1.8) the annotation is `null` if denoted with `@field:Json`.

NightlyNexus:
Just use `@Json`? There should be no need to use `@field:Json` when using the KotlinJsonAdapter.
I'm surprised this works on JDK 1.7, though...

premnirmal:
@NightlyNexus `@Json` doesn't work. See https://github.com/square/moshi/issues/315

NightlyNexus:
Have a failing test case?
From your code above, this passes.
[...]

PR #883, opened 2019-08-03
Update moshi-kotlin.pro

### LINKED ISSUE #880: Issue with R8 Rules in Moshi
I believe the rules that are included with Moshi should be updated to include

```
-keep class kotlin.Metadata {
    *;
}
```

I ran into an issue where my model classes were not being created correctly after R8 was applied because the constructor could not be found by Kotlin reflect (returns null for `KClass.primaryConstructor`, and my values were all set to null.

Here is an example project: https://github.com/AndrewReitz/moshi-r8-issue

After adding a rule to keep `kotlin.Metadata` and a rule to keep my class; the entire class, just keeping the constructor wasn't enough, the issue goes away and my constructor is used to create the class.

Diolor:
Fixes #880, #734 & #454

### DIFF
--- a/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro
+++ b/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro
@@ -1,5 +1,7 @@
--keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoaderImpl
+-keep class kotlin.Metadata
 
--keepclassmembers class kotlin.Metadata {
-    public <methods>;
-}
+-keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoader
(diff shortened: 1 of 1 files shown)


Diolor:
I don't think it makes sense. This is a moshi-kotlin-reflect proguard/R8 configuration of kotlin-reflect.

ZacSweers:
I think what @swankjesse meant is that these warnings appear to be issues with the kotlin-reflect.jar itself, and not strictly related to moshi

Diolor:
I see. I don't have that knowledge if it's 100% issues with kotlin-reflect. 
If you know that's the case, I can close this. 

Have any actions been taken to inform the maintainers in kotlin-reflect about this? Is there any chance they will acknowledge and fix "it" on their side?

ZacSweers:
It would be best to search or file on their issue tracker to get an answer for that

PR #885, opened 2019-08-08
Switch to openjdk8 for travis builds

ZacSweers:
oraclejdk8 isn't available unless we ust trusty dist, which is also an option


ZacSweers:
d'oh, missed that one 5116f34

PR #896, opened 2019-08-29
Support dynamic defaults constructor invocation in kotlin code gen without kotlin-reflect

ZacSweers:
This switches code gen to be able to dynamically invoke constructors and indicate which parameters are not set. No more need for the `copy()` trick! Now full dynamic default values are supported as well.

Caveat: This uses reflection to look up the synthetic defaults constructor that kotlinc generates for these types. Its signature is the original constructor + two extra parameters: one `mask` int parameter that's used to indicate which parameter values are set, and one throwaway `DefaultConstructorMarker` just used to differentiate it. This latter type is also what we use to find this constructor. The reflection is cached, and should be simple to set up with proguard (have pinged proguard folks separately to see if we can just match on that `DefaultConstructorMarker` parameter).

Implementation: There's a new `Util#invokeDefaultConstructor` helper method that accepts a target class, argument values (what the adapter has parsed), and a boolean array of which ones are set. The mask just sets bits that correspond to parameter indices. The boolean array is used to create a mask int, and then the args + mask and `null` throwaway for `DefaultConstructorMarker` are all invoked on the looked up defaults constructor. After eyeballing the [kotlin-reflect implementation](https://github.com/JetBrains/kotlin/blob/deb416484c5128a6f4bc76c39a3d9878b38cec8c/core/reflection.jvm/src/kotlin/reflect/jvm/internal/KCallableImpl.kt#L114-L168), this is fairly similar. It has more mechanics for general use (give it a map of parameter names to values) and figures out the set ones for the mask under the hood, but we know exactly which ones are already set from our parsing already and can just skip that step 👍.

To give a better sense - for a given `TestClass`:

```kotlin
data class TestClass(
    val required: String,
    val optional: String = "optional",
    val optional2: Int = 2,
    val dynamicSelfReferenceOptional: String = required,
    val dynamicOptional: Int = createInt(),
    val dynamicInlineOptional: Int = createInlineInt()
)

fun createInt(): Int {
  return 3
}

inline fun createInlineInt(): Int {
  return 3
}
```

The final bytecode for the constructor we're targeting is roughly this:

```kotlin
// $FF: synthetic method
   public TestClass(String var1, String var2, int var3, String var4, int var5, int var6, int var7, DefaultConstructorMarker var8) {
      if ((var7 & 2) != 0) {
         var2 = "";
      }

      if ((var7 & 4) != 0) {
         var3 = 2;
      }

      if ((var7 & 8) != 0) {
         var4 = var1;
      }

      if ((var7 & 16) != 0) {
         var5 = TestClassKt.createInt();
      }

      if ((var7 & 32) != 0) {
         int $i$f$createInlineInt = false;
         var6 = Random.Default.nextInt();
      }

      this(var1, var2, var3, var4, var5, var6);
   }
```

Pros:
- Full dynamic default values support, no more functional caveats
- No more double instantiation hacks needed
- No kotlin-reflect needed

Cons:
- Reflection cost, but amortized by the caching and (ideally) whatever I/O process this is running in partnership with.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -15,6 +15,7 @@
  */
 package com.squareup.moshi.kotlin.codegen
 
+import com.squareup.kotlinpoet.ANY
 import com.squareup.kotlinpoet.ARRAY
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.CodeBlock
@@ -29,12 +30,13 @@ import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.TypeVariableName
 import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
+import com.squareup.kotlinpoet.joinToCode
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.JsonDataException
 import com.squareup.moshi.JsonReader
 import com.squareup.moshi.JsonWriter
 import com.squareup.moshi.Moshi
-import me.eugeniomarletti.kotlin.metadata.isDataClass
+import com.squareup.moshi.internal.Util
 import me.eugeniomarletti.kotlin.metadata.shadow.metadata.ProtoBuf.Visibility
 import me.eugeniomarletti.kotlin.metadata.visibility
 import java.lang.reflect.Type
@@ -47,7 +49,6 @@ internal class AdapterGenerator(
 ) {
   private val nonTransientProperties = propertyList.filterNot { it.isTransient }
   private val className = target.name
-  private val isDataClass = target.proto.isDataClass
   private val visibility = target.proto.visibility!!
   private val typeVariables = target.typeVariables
 
@@ -182,7 +183,7 @@ internal class AdapterGenerator(
 
     for (property in nonTransientProperties) {
       result.addCode("%L", property.generateLocalProperty())
-      if (property.differentiateAbsentFromNull) {
+      if (property.hasLocalIsPresentName) {
         result.addCode("%L", property.generateLocalIsPresentProperty())
       }
     }
@@ -191,8 +192,8 @@ internal class AdapterGenerator(
     result.beginControlFlow("while (%N.hasNext())", readerParam)
     result.beginControlFlow("when (%N.selectName(%N))", readerParam, optionsProperty)
 
-    propertyList.forEachIndexed { index, property ->
-      if (property.differentiateAbsentFromNull) {
+    nonTransientProperties.forEachIndexed { index, property ->
+      if (property.hasLocalIsPresentName) {
         result.beginControlFlow("%L -> ", index)
         if (property.delegateKey.nullable) {
           result.addStatement("%N = %N.fromJson(%N)",
@@ -229,61 +230,60 @@ internal class AdapterGenerator(
     result.endControlFlow() // while
     result.addStatement("%N.endObject()", readerParam)
 
-    // Call the constructor providing only required parameters.
-    var hasOptionalParameters = false
-    result.addCode("«var %N = %T(", resultName, originalTypeName)
     var separator = "\n"
-    for (property in propertyList) {
-      if (!property.hasConstructorParameter) {
-        continue
-      }
-      if (property.hasDefault) {
-        hasOptionalParameters = true
-        continue
-      }
+    var useDefaultsConstructor = false
+    val parameterProperties = propertyList.asSequence()
+        .filter { it.hasConstructorParameter }
+        .onEach {
+          useDefaultsConstructor = useDefaultsConstructor || it.hasDefault
+        }
+        .toList()
+
+    if (useDefaultsConstructor) {
+      // Dynamic default constructor call
+      val booleanArrayBlock = parameterProperties.map { param ->
+        when {
+          param.isTransient -> CodeBlock.of("false")
+          param.hasLocalIsPresentName -> CodeBlock.of(param.localIsPresentName)
+          else -> CodeBlock.of("true")
+        }
+      }.joinToCode(", ")
+      result.addStatement("val argPresentValues = booleanArrayOf(%L)", booleanArrayBlock)
+      result.addCode("«val args: %T = arrayOf(", ARRAY.parameterizedBy(ANY.copy(nullable = true)))
+    } else {
+      // Standard constructor call
+      result.addCode("«val %N = %T(", resultName, originalTypeName)
+    }
+
+    for (property in parameterProperties) {
       result.addCode(separator)
-      result.addCode("%N = %N", property.name, property.localName)
-      if (property.isRequired) {
+      if (useDefaultsConstructor) {
+        if (property.isTransient) {
+          // We have to use the default primitive for the available type in order for
+          // invokeDefaultConstructor to properly invoke it. Just using "null" isn't safe because
+          // the transient type may be a primitive type.
+          result.addCode(property.target.type.defaultPrimitiveValue())
+        } else {
+          result.addCode("%N", property.localName)
+        }
+      } else {
+        result.addCode("%N = %N", property.name, property.localName)
+      }
+      if (!property.isTransient && property.isRequired) {
         result.addCode(" ?: throw·%L", jsonDataException(
             "Required property '", property.localName, "' missing at ", readerParam))
       }
       separator = ",\n"
     }
-    result.addCode(")»\n", originalTypeName)
 
-    // Call either the constructor again, or the copy() method, this time providing any optional
-    // parameters that we have.
-    if (hasOptionalParameters) {
-      if (isDataClass) {
-        result.addCode("«%1N = %1N.copy(", resultName)
-      } else {
-        result.addCode("«%1N = %2T(", resultName, originalTypeName)
-      }
-      separator = "\n"
-      for (property in propertyList) {
-        if (!property.hasConstructorParameter) {
-          continue // No constructor parameter for this property.
-        }
-        if (isDataClass && !property.hasDefault) {
-          continue // Property already assigned.
-        }
-
-        result.addCode(separator)
-        when {
-          property.differentiateAbsentFromNull -> {
-            result.addCode("%2N = if (%3N) %4N else %1N.%2N",
-                resultName, property.name, property.localIsPresentName, property.localName)
-          }
-          property.isRequired -> {
-            result.addCode("%1N = %2N", property.name, property.localName)
-          }
-          else -> {
-            result.addCode("%2N = %3N ?: %1N.%2N", resultName, property.name, property.localName)
-          }
-        }
-        separator = ",\n"
-      }
-      result.addCode("»)\n")
+    result.addCode(")»\n")
+    if (useDefaultsConstructor) {
+      result.addStatement(
+          "val %N = %T.invokeDefaultConstructor(%T::class.java, args, argPresentValues)",
+          resultName,
+          Util::class.asClassName(),
+          originalTypeName
+      )
     }
 
     // Assign properties not present in the constructor.
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -271,7 +271,7 @@ internal class AdapterGenerator(
         }
       }.joinToCode(", ")
       result.addStatement(
-          "val %1L = this.%2L ?: %3T.lookupDefaultsConstructor(%4T::class.java).also { this.%2L = it  }",
+          "val·%1L·= this.%2N ?: %3T.lookupDefaultsConstructor(%4T::class.java).also·{·this.%2N·=·it·}",
           localConstructorName,
           constructorProperty,
           MOSHI_UTIL,
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -15,7 +15,6 @@
  */
 package com.squareup.moshi.kotlin.codegen
 
-import com.squareup.kotlinpoet.ANY
 import com.squareup.kotlinpoet.ARRAY
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.CodeBlock
@@ -260,7 +259,6 @@ internal class AdapterGenerator(
       CodeBlock.of("return·")
     }
     val maskName = nameAllocator.newName("mask")
-    val argsName = nameAllocator.newName("args")
     val localConstructorName = nameAllocator.newName("localConstructor")
     if (useDefaultsConstructor) {
       // Dynamic default constructor call
@@ -280,8 +278,14 @@ internal class AdapterGenerator(
       )
       result.addStatement("val %L = %T.createDefaultValuesParametersMask(%L)",
           maskName, MOSHI_UTIL, booleanArrayBlock)
-      result.addCode("«val %L: %T = arrayOf(", argsName,
-          ARRAY.parameterizedBy(ANY.copy(nullable = true)))
+      result.addCode(
+          "«%L%T.invokeDefaultConstructor(%T::class.java, %L, %L, ",
+          returnOrResultAssignment,
+          MOSHI_UTIL,
+          originalTypeName,
+          localConstructorName,
+          maskName
+      )
     } else {
       // Standard constructor call
       result.addCode("«%L%T(", returnOrResultAssignment, originalTypeName)
@@ -308,18 +312,7 @@ internal class AdapterGenerator(
       separator = ",\n"
     }
 
-    result.addCode(")»\n")
-    if (useDefaultsConstructor) {
-      result.addStatement(
-          "%L%T.invokeDefaultConstructor(%T::class.java, %L, %L, %L)",
-          returnOrResultAssignment,
-          MOSHI_UTIL,
-          originalTypeName,
-          localConstructorName,
-          argsName,
-          maskName
-      )
-    }
+    result.addCode("\n»)\n")
 
     // Assign properties not present in the constructor.
     for (property in nonTransientProperties) {
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -176,8 +176,6 @@ internal class AdapterGenerator(
   }
 
   private fun generateFromJsonFun(): FunSpec {
-    val resultName = nameAllocator.newName("result")
-
     val result = FunSpec.builder("fromJson")
         .addModifiers(KModifier.OVERRIDE)
         .addParameter(readerParam)
@@ -241,6 +239,14 @@ internal class AdapterGenerator(
         }
         .toList()
 
+    val resultName = nameAllocator.newName("result")
+    val hasNonConstructorProperties = nonTransientProperties.any { !it.hasConstructorParameter }
+    val returnOrResultAssignment = if (hasNonConstructorProperties) {
+      // Save the result var for reuse
+      CodeBlock.of("val %N = ", resultName)
+    } else {
+      CodeBlock.of("return·")
+    }
     val maskName = nameAllocator.newName("make")
     val argsName = nameAllocator.newName("ars")
     if (useDefaultsConstructor) {
@@ -257,7 +263,7 @@ internal class AdapterGenerator(
       result.addCode("«val %L: %T = arrayOf(", argsName, ARRAY.parameterizedBy(ANY.copy(nullable = true)))
     } else {
       // Standard constructor call
-      result.addCode("«val %N = %T(", resultName, originalTypeName)
+      result.addCode("«%L%T(", returnOrResultAssignment, originalTypeName)
     }
 
     for (property in parameterProperties) {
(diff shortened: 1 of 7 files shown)


moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -520,4 +534,69 @@ public static String typeAnnotatedWithAnnotations(Type type,
...
+    }
+
+    throw new IllegalStateException("No defaults constructor found for " + targetClass);
+  }
+
+  private static int createMask(boolean[] argPresentValues) {
+    int mask = 0;
+    for (int i = 0; i < argPresentValues.length; ++i) {
+      if (!argPresentValues[i]) {
+        mask = mask | (1 << (i % Integer.SIZE));
```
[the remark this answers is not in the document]
ZacSweers: this was straight from the kotlin-reflect impl, will see if I can find out why it's done this way

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -520,4 +534,69 @@ public static String typeAnnotatedWithAnnotations(Type type,
...
+    }
+    Constructor<?> defaultConstructor = DEFAULT_CONSTRUCTOR_CACHE.get(targetClass);
+    if (defaultConstructor == null) {
+      defaultConstructor = findConstructor(targetClass);
+      defaultConstructor.setAccessible(true);
+      DEFAULT_CONSTRUCTOR_CACHE.put(targetClass, defaultConstructor);
+    }
+    int mask = createMask(argPresentValues);
+    Object[] finalArgs = new Object[args.length + 2];
+    System.arraycopy(args, 0, finalArgs, 0, args.length);
```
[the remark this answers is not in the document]
ZacSweers: 💡 cc4c710

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -520,4 +534,69 @@ public static String typeAnnotatedWithAnnotations(Type type,
...
+    finalArgs[finalArgs.length - 2] = mask;
+    finalArgs[finalArgs.length - 1] = null; // DefaultConstructorMarker param
+    try {
+      //noinspection unchecked
+      return (T) defaultConstructor.newInstance(finalArgs);
+    } catch (InstantiationException e) {
+      throw new IllegalStateException("Could not instantiate instance of " + targetClass);
+    } catch (IllegalAccessException e) {
+      throw new IllegalStateException("Could not access defaults constructor of " + targetClass);
+    } catch (InvocationTargetException e) {
```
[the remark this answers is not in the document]
ZacSweers: baaa0c5

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -228,65 +230,64 @@ internal class AdapterGenerator(
...
+    if (useDefaultsConstructor) {
+      // Dynamic default constructor call
+      val booleanArrayBlock = parameterProperties.map { param ->
+        when {
+          param.isTransient -> CodeBlock.of("false")
+          param.hasLocalIsPresentName -> CodeBlock.of(param.localIsPresentName)
+          else -> CodeBlock.of("true")
+        }
+      }.joinToCode(", ")
+      result.addStatement("val argPresentValues = booleanArrayOf(%L)", booleanArrayBlock)
```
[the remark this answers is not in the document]
ZacSweers: nameallocator - 48d1ef9
inlined masking - a571a4b

moshi/src/main/java/com/squareup/moshi/internal/Util.java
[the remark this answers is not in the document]
ZacSweers: Removed it and it works fine 23a8c9b

ZacSweers:
> Tempted to just generate sibling classes in bytecode which expose trampolines directly to the synthetic constructor rather than relying on (more) reflection.

That would be neat! I fiddled a bit with generating class files into kapt output via bytebuddy, but I'm not able to actually get existing sources to actually link referencing the class. Something for another time maybe, unless you know a better tool? Maybe it works even without the IDE linking, but didn't get that far in my testing yet

JakeWharton:
That sounds like a(nother on the long list and ever-growing pile of) kapt bug 😞

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -228,65 +230,77 @@ internal class AdapterGenerator(
...
+    val resultName = nameAllocator.newName("result")
+    val hasNonConstructorProperties = nonTransientProperties.any { !it.hasConstructorParameter }
+    val returnOrResultAssignment = if (hasNonConstructorProperties) {
+      // Save the result var for reuse
+      CodeBlock.of("val %N = ", resultName)
+    } else {
+      CodeBlock.of("return·")
+    }
+    val maskName = nameAllocator.newName("mask")
+    val argsName = nameAllocator.newName("ars")
```
rharter: nit: Maybe call this `args` for readability of generated code.

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -44,6 +45,19 @@
 public final class Util {
   public static final Set<Annotation> NO_ANNOTATIONS = Collections.emptySet();
   public static final Type[] EMPTY_TYPE_ARRAY = new Type[] {};
+  @Nullable private static final Class<?> DEFAULT_CONSTRUCTOR_MARKER;
+  private static final ConcurrentHashMap<Class<?>, Constructor<?>> DEFAULT_CONSTRUCTOR_CACHE
```
rharter: What are the implications of storing the default constructor cache in a static field?  Would it be better to cache the default constructor in the generated adapters? That would allow the usage of the cache to be scoped to the usage of the adapters.

If an app does a one time read of a large dataset, to populate a DB or something, then these would be retained for the lifetime of the app, as opposed to being able to be collected once parsing was complete (assuming the adapters were only retained within a specific scope).

ZacSweers: That's not a bad idea, but storing mutable state in the adapters makes me worried. What would the implementation of that look like? My first thought is basically a volatile field and race construction. The other side in caching it per-adapter is that adapters are cached in moshi, so this would only really work if you're newBuilder-ing moshi instances around and dropping them when they go out of scope

rharter: That was done in AV: GSON to lazily load child type adapters as needed.  In that case it was due to bad API design, but here it would be a single field. https://github.com/rharter/auto-value-gson/pull/165

This might be a small, unnecessary optimization, but in my cases I use different Moshi instances in different contexts.  

When the app launches I might download new data from the server to update the database depending on the TTL, and those entities aren't read again, so I use a Moshi that just knows about those.

Other places, when I'm writing project files, for instance, my `ProjectRepository` has its' own Moshi instance that only knows about the project related JSON adapters.

ZacSweers: Yeah I see your point. I think there's a balance that could be had somewhere, but for now I think this approach is fine as a starting point. I modeled it after Butterknife, which does [the same](https://github.com/JakeWharton/butterknife/blob/master/butterknife/src/main/java/butterknife/ButterKnife.java#L83). I figure if there hasn't been memory issues there across its long history, this is probably fine. This implementation defers the reflection/lookup until deserialization-time - so it's fairly lazy, won't occur for serialization, and is likely on a background thread amortized by other I/O going on.

The AV-Gson issue (at least what Adam was trying to solve at uber when he opened the PR) was more about avoiding transitive classloading on initialization.

WDYT?

rharter: I think the Butterknife case is quite a bit different since that's primarily a terminal dependency, not a transient one.  Since any of your dependencies can use Moshi under the hood, this means you're storing the ref to the constructor of any type that's been deserialized and meets the criteria, no matter where it comes from, and that's unrecoverable memory. 

I'm sure in most cases this isn't a huge deal, but that effectively means that field could theoretically be storing constructor refs for models from all 6 analytics SDKs in your app, plus all 4 ad SDKs, plus that remote logging SDK, remote config SDK, etc. I can't imagine each individual reference is large, but that can add up.

Normally in these situations we can control our memory usage by deciding what we keep references to, but a static field takes away that option.

I'm not looking to die on this hill, and if you don't think it's an issue, and no one else does either, then that's fine with me.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
ZacSweers: derp, typo. Will fix

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: Updated in 2ae453d after talking with folks offline

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
ZacSweers: updated as part of 2ae453d

ZacSweers:
@Egorand do you understand the kotlinpoet issue? It's specifically an issue with the `addStatement` on `AdapterGenerator` at line 273. Right now it yields this error, but the error goes away (but obviously becomes a later compilation error) if I remove it

```
[ERROR] [kapt] An exception occurred: java.lang.IllegalStateException: statement enter « followed by statement enter «
[ERROR]         at com.squareup.kotlinpoet.CodeWriter.emitCode(CodeWriter.kt:297)
[ERROR]         at com.squareup.kotlinpoet.PropertySpec.emit$kotlinpoet(PropertySpec.kt:88)
[ERROR]         at com.squareup.kotlinpoet.PropertySpec.emit$kotlinpoet$default(PropertySpec.kt:58)
[ERROR]         at com.squareup.kotlinpoet.CodeWriter.emitLiteral(CodeWriter.kt:355)
[ERROR]         at com.squareup.kotlinpoet.CodeWriter.emitCode(CodeWriter.kt:223)
[ERROR]         at com.squareup.kotlinpoet.CodeWriter.emitCode$default(CodeWriter.kt:216)
[ERROR]         at com.squareup.kotlinpoet.FunSpec.emit$kotlinpoet(FunSpec.kt:107)
[ERROR]         at com.squareup.kotlinpoet.TypeSpec.emit$kotlinpoet(TypeSpec.kt:267)
[ERROR]         at com.squareup.kotlinpoet.TypeSpec.emit$kotlinpoet$default(TypeSpec.kt:88)
[ERROR]         at com.squareup.kotlinpoet.FileSpec.emit(FileSpec.kt:160)
[ERROR]         at com.squareup.kotlinpoet.FileSpec.writeTo(FileSpec.kt:64)
[ERROR]         at com.squareup.kotlinpoet.FileSpec.writeTo(FileSpec.kt:110)
[ERROR]         at com.squareup.moshi.kotlin.codegen.JsonClassCodegenProcessor.process(JsonClassCodegenProcessor.kt:91)

```

ZacSweers:
Fixed the code gen issue and added a proguard/R8 rule. I think this is ready for merge, @rharter @JakeWharton want to take another pass at the constructor lookup change and proguard rule? Will test the proguard rule separately in a sample app as well. Example generated adapter here: https://gist.github.com/ZacSweers/39771a511a687268f5f2cc947b89f89a

**Bytecode generation**
I've experimented a bunch with generating raw bytecode directly (via ASM) that bridges to the synthetic constructor and save the reflection. I've been able to get good progress, but I think build tooling needs work before we could roll it out. In short - everything functionally works, but...
- On gradle
  - The generated bridge class does make it into the final jar. I can find it in the debugger and run it reflectively
  - Is partially indexed in the IDE out of the box. 
    - Visible but classpath/dependency errors
    - Fully visible if the output classes directory is added to the project sourcesets, unclear if that's necessary for compilation though
  - Compilation fails, and class doesn't appear to be part of the compilation classpath itself. Compilation fails on unresolved reference despite linking fine in the IDE.
  - Unclear if this is a bug with gradle, kotlin's gradle plugin, or even how kapt handles generated .class files. Going to ping some folks to try to figure out where to file this.
- On Maven
  - The generated bridge class is not indexed in the IDE, even if added to the source sets. 
  - Compilation fails on unresolved reference. Have not gotten as far as determining whether it makes it to the final jar.

ZacSweers:
The preceding wildcard isn't legal in proguard/R8, so we'll have to just keep all the constructors for now. 

Another thing that occurred to me is that we'd need to keep the `kotlin.jvm.internal.DefaultConstructorMarker` name or tie it with `identifiernamestring` in order to match the constructor, but when tried without it, it surprisingly Just Worked™️. After going down the rabbit hole I found [this in R8](https://r8.googlesource.com/r8/+/9661c4c4937daa8ce7cbf56bbba87f7806a9ce50/src/main/java/com/android/tools/r8/CompatProguardCommandBuilder.java), which I _think_ means R8 automatically adds `identifiernamestring` rules for common reflection APIs like the `Class.forName(...)` that we use and hence why it worked out of the box. As far as I can tell, this option isn't supported in Proguard though, nor can I find any official documentation anywhere online for `identifiernamestring` in general other than source code and some dagger comments.

So boils down to - do we want to preemptively include this in moshi's proguard file? Or just tell proguard users to manually include it?

```proguard
-keepnames class kotlin.jvm.internal.DefaultConstructorMarker
```

ZacSweers:
I'm going to go with including it for now, but can remove later if we decide to recommend manually including it

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -162,26 +177,24 @@ internal class AdapterGenerator(
...
 
   private fun jsonDataException(
-    description: String,
-    identifier: String,
-    condition: String,
-    reader: ParameterSpec
+      description: String,
+      identifier: String,
+      condition: String,
+      reader: ParameterSpec
```
JakeWharton: indent

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -228,65 +241,81 @@ internal class AdapterGenerator(
...
-        continue
-      }
-      if (property.hasDefault) {
-        hasOptionalParameters = true
-        continue
-      }
+    var useDefaultsConstructor = false
+    val parameterProperties = propertyList.asSequence()
+        .filter { it.hasConstructorParameter }
+        .onEach {
```
JakeWharton: There's only one real operator here. Would prefer
```kotlin
val parameterProperties = propertyList.filter { it.hasConstructorParameter }
val useDefaultsConstructor = parameterProperties.any { it.hasDefault }
```
to avoid the need to use side-effects.

moshi/src/main/resources/META-INF/proguard/moshi.pro
```diff
@@ -16,6 +16,16 @@
...
 -keepnames @com.squareup.moshi.JsonClass class *
 
+# Retain generated target class's synthetic defaults constructor and keep DefaultConstructorMarker's
+# name. We will look this up reflectively to invoke the type's constructor.
+#
+# We can't _just_ keep the defaults constructor because Proguard/R8's spec doesn't allow wildcard
+# matching preceding parameters.
+-keepnames class kotlin.jvm.internal.DefaultConstructorMarker
+-keepclassmembers @com.squareup.moshi.JsonClass class * {
+    <init>(...);
```
JakeWharton: Can you match on synthetic ones only?

moshi/src/main/resources/META-INF/proguard/moshi.pro
```diff
@@ -16,6 +16,16 @@
 # The name of @JsonClass types is used to look up the generated adapter.
 -keepnames @com.squareup.moshi.JsonClass class *
 
+# Retain generated target class's synthetic defaults constructor and keep DefaultConstructorMarker's
+# name. We will look this up reflectively to invoke the type's constructor.
+#
+# We can't _just_ keep the defaults constructor because Proguard/R8's spec doesn't allow wildcard
+# matching preceding parameters.
+-keepnames class kotlin.jvm.internal.DefaultConstructorMarker
+-keepclassmembers @com.squareup.moshi.JsonClass class * {
```
JakeWharton: Should we match on the metadata annotation as well to prevent affecting Java classes?

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -520,4 +531,91 @@ public static String typeAnnotatedWithAnnotations(Type type,
...
+  }
+
+  /**
+   * Creates an mask with bits set to indicate which indices of a default constructor's parameters
+   * are set.
+   *
+   * @param argPresentValues vararg of all present values (set or unset). Max allowable size is 32.
+   * @return the created mask.
+   */
+  public static int createDefaultValuesParametersMask(boolean... argPresentValues) {
```
JakeWharton: I vaguely remember commenting about moving this logic to the generated code which would eliminate the "is present" booleans and avoid the boolean array allocation to call this method. Was there a reason that didn't happen?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
```diff
@@ -29,24 +29,28 @@ import com.squareup.kotlinpoet.TypeSpec
...
 import javax.lang.model.element.TypeElement
 
+private val MOSHI_UTIL = Util::class.asClassName()
+
 /** Generates a JSON adapter for a target type. */
 internal class AdapterGenerator(
-  target: TargetType,
-  private val propertyList: List<PropertyGenerator>
+    target: TargetType,
+    private val propertyList: List<PropertyGenerator>
```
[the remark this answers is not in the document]
ZacSweers: No ktlint here! Still a maven project so no spotless available 😬, just checkstyle

moshi/src/main/resources/META-INF/proguard/moshi.pro
ZacSweers: Didn't even realize proguard allowed you to do that! Will fix

moshi/src/main/resources/META-INF/proguard/moshi.pro
ZacSweers: Good call, since this is suggested for keeping enum members too. Will fix

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: I misunderstood and thought you were just asking to make this argument a vararg. You think we should inline _all_ of this? It'd be a potentially really long cascade of if-checks

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
ZacSweers: Will fix

moshi/src/main/java/com/squareup/moshi/internal/Util.java
rharter: Could the mask itself move out to the adapter, then instead of property and propertySet, you'd just update the mask int when the value is read, right inside the when?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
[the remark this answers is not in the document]
JakeWharton: There seem to be at least two ktlint maven plugins from a quick Google. We should add one.

moshi/src/main/java/com/squareup/moshi/internal/Util.java
JakeWharton: Right I'm saying we don't use booleans and just use the int and set bits inside the generated code.

ZacSweers: I think I understand what mean now 👍. Will do but in a separate PR from the others above since it's a little more involved

ZacSweers: Note we can't entirely escape local booleans, but now they'd only be used for gating setting non-constructor `var` properties

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
ZacSweers: Part of #907

moshi/src/main/resources/META-INF/proguard/moshi.pro
ZacSweers: Part of #907

moshi/src/main/resources/META-INF/proguard/moshi.pro
ZacSweers: Part of #907

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: This gets a little tricky, because the mask only sets bits for parameters that are _not_ set. This means we'd likely have to manually set every bit first, and then unset them as we parse types. Worth the overhead?

rharter: Is it a problem to just invert the mask?

```kotlin
var mask = 1
                0 -> {
                    nullableFirstName = stringAdapter.fromJson(reader)
                    mask = mask and (0 shl 0)
                }
                1 -> {
                    nullableLastName = stringAdapter.fromJson(reader)
                    mask = mask and (0 shl 1)
                } 
                ...
```

ZacSweers: Nope that works, I had tried that and got test failures and assumed it was wrong. Turned out it was right, and something else had broken. Teaches me to read the test output better :)

Opened a PR: #908

Issue #263, opened 2017-02-16
Enum adapter stop reading name value from @Json in v1.4.0


cesarnorena:
I update from v1.3.1 to v1.4.0 and the enum adapter stop reading the name in json annotation. Now I get null for ExampleEntity.value

> @Json(name="value")

I have the next Json and the entity to store it:

```
{"value":1}

data class ExampleEntity(
    val value: ExampleEnum?
)
```

My Enum class is like this:

```
enum class ExampleEnum {
    @Json(name = "1") FIRST,
}
```

cesarnorena:
I suppose that this is happening after remove the next lines in v1.4.0 from "fromJson"

```
T constant = nameConstantMap.get(name);
if (constant != null) return constant;
```

cesarnorena:
```
Exception in thread "main" com.squareup.moshi.JsonDataException: Expected one of [0, 1] but was 0 at path $.value
	at com.squareup.moshi.StandardJsonAdapters$EnumJsonAdapter.fromJson(StandardJsonAdapters.java:244)
	at com.squareup.moshi.StandardJsonAdapters$EnumJsonAdapter.fromJson(StandardJsonAdapters.java:213)
	at com.squareup.moshi.JsonAdapter$2.fromJson(JsonAdapter.java:127)
	at com.squareup.moshi.ClassJsonAdapter$FieldBinding.read(ClassJsonAdapter.java:204)
	at com.squareup.moshi.ClassJsonAdapter.fromJson(ClassJsonAdapter.java:166)
	at com.squareup.moshi.JsonAdapter$2.fromJson(JsonAdapter.java:127)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:34)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:38)
```

NightlyNexus:
I do not think this worked in 1.3.1, either. It would have been a bug if it did.

I tried running
```java
@Test public void annotatedEnum() throws Exception {
  Moshi moshi = new Moshi.Builder().build();
  JsonAdapter<Roshambo> adapter = moshi.adapter(Roshambo.class).lenient();
  assertThat(adapter.fromJson("1")).isEqualTo(Roshambo.SCISSORS);
  assertThat(adapter.toJson(Roshambo.SCISSORS)).isEqualTo("1");
}
enum Roshambo {
  @Json(name = "1") SCISSORS
}
```
[...]

PR #897, opened 2019-09-03
Add EnumJsonAdapter.NullFallbackFactory

villesinisalo:
Provide an adapter factory for deserializing all unrecognized enum values within the same Moshi instance as `null` using existing `EnumJsonAdapter`  with fallback value.

**Why:** This is the default, out of the box behavior of Gson. One of the stated goals of Moshi is [similarity to Gson, but with fewer built-in adapters](https://github.com/square/moshi#borrows-from-gson). 

In an ideal world, one never needs to deserialize unrecognized enum values. But APIs change, and e.g. a user might be running an outdated version of a client. The server might have added new values to an existing enum that did not exist at the time the client was written. It's far easier to build your clients to fail gracefully on `null` enum values than to e.g. introduce a new JSON field every time you need to expand said enum. With Gson this is the default behavior, but with Moshi whole document serialization throws unless one adds an `EnumJsonAdapter` for *every single* enum type in the project that might be expanded in the future (this is especially daunting for a project that are migrating from Gson to Moshi), or figure how to write a custom `JsonAdapter.Factory`. Moshi should preferably just provide the latter as standard add-on component.

This behavior has [been](https://github.com/square/moshi/issues/218) [requested](https://github.com/square/moshi/pull/607#issuecomment-413057681) [multiple](https://github.com/square/moshi/pull/607#issuecomment-413302505) [times](https://github.com/square/moshi/issues/263#issuecomment-481305545), but the [complete](https://github.com/square/moshi/issues/218#issuecomment-262664970) [solution](https://github.com/Meisolsson/GitHubSdk/pull/25#issuecomment-462058587) was never delivered AFAIK.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
@@ -20,8 +20,15 @@ import com.squareup.moshi.JsonAdapter;
 import com.squareup.moshi.JsonDataException;
 import com.squareup.moshi.JsonReader;
 import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.Types;
+
 import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
 import java.util.Arrays;
+import java.util.Set;
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 
 /**
@@ -34,6 +41,16 @@ import javax.annotation.Nullable;
  *   Moshi moshi = new Moshi.Builder()
  *       .add(CurrencyCode.class, EnumJsonAdapter.create(CurrencyCode.class)
  *           .withUnknownFallback(CurrencyCode.USD))
+ *       .build();
+ * }</pre>
+ *
+ * If you want to map all unrecognized enum values within the same {@link com.squareup.moshi.Moshi
+ * Moshi instance} to null, you can add a {@link NullFallbackFactory}:
+ *
+ * <pre> {@code
+ *
+ *   Moshi moshi = new Moshi.Builder()
+ *       .add(new EnumJsonAdapter.NullFallbackFactory())
  *       .build();
  * }</pre>
  */
@@ -106,5 +123,30 @@ public final class EnumJsonAdapter<T extends Enum<T>> extends JsonAdapter<T> {
 
   @Override public String toString() {
     return "EnumJsonAdapter(" + enumType.getName() + ")";
+  }
+
+  /**
+   * A {@link JsonAdapter.Factory} that uses {@link EnumJsonAdapter} with {@link
+   * EnumJsonAdapter#withUnknownFallback(Enum) null fallback} to serialize all
+   * enums.
+   *
+   * The returned adapter is wrapped in {@link #nullSafe()} to allow both serializing and
+   * deserializing null values.
+   */
+  public static class NullFallbackFactory implements JsonAdapter.Factory {
+    @CheckReturnValue
+    @Nullable
+    @Override
+    public JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations, Moshi moshi) {
+      if (!annotations.isEmpty()) {
+        return null;
+      }
+      Class<?> rawType = Types.getRawType(type);
+      if (rawType.isEnum()) {
+        //noinspection unchecked
+        return new EnumJsonAdapter(rawType, null, true).nullSafe();
+      }
+      return null;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


ZacSweers:
I don't think this is a good idea, as it conflates unknown, absence, and nullability. This only really works if you make every enum property nullable (which may not make sense) _and_ there is no notion of absence. It is far better to use some "UNKNOWN" enum member and use the enum default support to indicate that so your code can be mindful of it.

In this case - you would have no idea if the enum property was unknown, just null, or absent after deserialization. It's a very broad strokes API that would create or facilitate more problems than it solves.

villesinisalo:
@ZacSweers note that I am not proposing an API change - this is an opt-in adapter factory for the adapters package. So users would be still making a conscious choice to enable this behavior over the default of throwing when the string can not be mapped to an enum value - same as when enabling deserialization of unknown enums to a fallback value or null per enum type.

I do understand your point of view, but don't really agree that it should be imposed on all users of Moshi. Consider the following scenarios:

 1. Your models are generated from a Swagger spec by [swagger-codegen](https://github.com/swagger-api/swagger-codegen) or similar. The idea of having a "fallback" enum value is rather novel, and I don't know a single code generator that would support it.
 2. Your models originate from a shared Java codebase with the server. Handling "UNKNOWN" enum values on the server rarely makes sense - it's often preferable to just return HTTP 400 or 500.
 3. You are transitioning from using Gson to Moshi. In Gson, all enum fields are effectively nullable, and your existing codebase hopefully already makes a consideration of this fact. Changing to using a fallback enum values could require substantial refactoring of working code when switching to Moshi.
 4. What happens when the server adds a new enum value `UNKNOWN`? You need to refactor all the client code that uses that as fallback enum value to make actual use of it.

(E.g. in my current codebase cases 2 & 3 above are in effect.)

In summary, I believe it's a matter of taste whether using a fallback value or null is preferable, and there are pros and cons to either approach. 

One possibility would be adding support for new `@JsonEnumFallback` enum value annotation to this adapter factory, so that users could selectively elect to specify a value to use as fallback for individual enum types in place of null. This would make all deserialized instances of such enum type non-nullable and I could avoid wrapping to `.nullSafe()` in the factory when a fallback value is provided. How does that sound?

ZacSweers:
I think there was a bit of a misunderstanding here - I know this isn't making it part of the default API, nor was that ever the implication of that being the main problem. It's making it a first-party option in any capacity that's the problem, as I don't think it passes the smell test as a first-party option. 

There are too many asterisks to its usage here to offer as a general-use tool. At best it recreates a bad GSON behavior that obfuscates why a property is null (was it absent? unknown? who knows!), and at worst it conflates the previous two + hard fails your property deserialization if you didn't remember to make it nullable. I think that's fine if you need it in your project to use it on your own, but it doesn't make sense for Moshi's first-party adapter suite.

Also - Another (very popular) code generator that supports the unknown case is Protocol Buffers. You can find more under the Enumeration section [here](https://developers.google.com/protocol-buffers/docs/proto). It's not JSON, but same philosophy with regards to enum handling in this case. It's definitely not a new issue nor new pattern.

rewgoes:
For my use case, the API returns a `List<Enum>` which is used to know which features are available to the user. As new features are added to the service, new `Enum` values are added/returned by the API.

In this case, users running old versions of the app won't have the new values related to the new features. So,  as I needed the same result parsed by GSON, which only returns the values that it recognizes, I've developed the following adapter:

```kotlin
class SkipBadEnumListAdapter(private val elementAdapter: JsonAdapter<Enum<*>>)
    : JsonAdapter<List<Enum<*>>>() {
    object Factory : JsonAdapter.Factory {
        override fun create(type: Type, annotations: MutableSet<out Annotation>, moshi: Moshi)
                : JsonAdapter<*>? {
            return if (annotations.isEmpty()
                    && Types.getRawType(type) == List::class.java
                    && type is ParameterizedType
                    && type.actualTypeArguments.size == 1
                    && Types.getRawType(type.actualTypeArguments[0]).isEnum) {
                val elementType = Types.collectionElementType(type, List::class.java)
                val elementAdapter = moshi.adapter<Enum<*>>(elementType)
                SkipBadEnumListAdapter(elementAdapter).nullSafe()
            } else {
                null
            }
        }
    }

    override fun fromJson(reader: JsonReader): List<Enum<*>>? {
        val result = mutableListOf<Enum<*>>()
        reader.beginArray()
        while (reader.hasNext()) {
            try {
                val peeked = reader.peekJson()
                elementAdapter.fromJson(peeked)?.let { result.add(it) }
            } catch (ignored: JsonDataException) {
                // Skip bad element ;)
            }
            reader.skipValue()
        }
        reader.endArray()
        return result

    }

    override fun toJson(writer: JsonWriter, value: List<Enum<*>>?) {
        if (value == null) {
        }
        writer.beginArray()
        for (i in value.indices) {
            elementAdapter.toJson(writer, value[i])
        }
        writer.endArray()
    }
}
```

PR #898, opened 2019-09-05
Don't reflect on kotlinx and update doc

ZacSweers:
This adds kotlinx to the exclusion list, and also updates the doc on ClassJsonAdapter


moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
```diff
@@ -39,9 +39,11 @@
  *
  * <ul>
  *   <li>android.*
+ *   <li>androidx.*
```
ZacSweers: note androidx is already excluded, just wasn't listed here before

PR #899, opened 2019-09-05
Clean up warnings in KotlinJsonAdapter

ZacSweers:
Misc idiomatic cleanups, nothing functional changing

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -142,8 +142,10 @@ internal class KotlinJsonAdapter<T>(
   }
 
   /** A simple [Map] that uses parameter indexes instead of sorting or hashing. */
-  class IndexedParameterMap(val parameterKeys: List<KParameter>, val parameterValues: Array<Any?>)
-    : AbstractMap<KParameter, Any?>() {
+  class IndexedParameterMap(
+    private val parameterKeys: List<KParameter>,
+    private val parameterValues: Array<Any?>
+  ) : AbstractMap<KParameter, Any?>() {
 
     override val entries: Set<Entry<KParameter, Any?>>
       get() {
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -213,6 +213,7 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       val parameter = parametersByName[property.name]
 
       if (Modifier.isTransient(property.javaField?.modifiers ?: 0)) {
+        @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
         if (parameter != null && !parameter.isOptional) {
           throw IllegalArgumentException(
               "No default value for transient constructor $parameter")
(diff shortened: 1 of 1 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -186,18 +189,18 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       // Fall back to a reflective adapter when the generated adapter is not found.
     }
 
-    if (rawType.isLocalClass) {
-      throw IllegalArgumentException("Cannot serialize local class or object expression ${rawType.name}")
+    require(!rawType.isLocalClass) {
```
rharter: Love this! Is this new in 1.3.50? I just discovered it the other day.

JakeWharton: The inspection is new, yes

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -248,9 +254,7 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
     for (parameter in constructor.parameters) {
       val binding = bindingsByName.remove(parameter.name)
-      if (binding == null && !parameter.isOptional) {
-        throw IllegalArgumentException("No property for required constructor ${parameter}")
-      }
+      require(!(binding == null && !parameter.isOptional)) { "No property for required constructor $parameter" }
```
JakeWharton:
```suggestion
      require(binding != null || parameter.isOptional) { "No property for required constructor $parameter" }
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -210,13 +213,15 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       val parameter = parametersByName[property.name]
 
       if (Modifier.isTransient(property.javaField?.modifiers ?: 0)) {
+        @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
```
JakeWharton: Down below you replaced a similar one

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -210,13 +213,15 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
         if (parameter != null && !parameter.isOptional) {
           throw IllegalArgumentException(
               "No default value for transient constructor $parameter")
         }
         continue
       }
 
+      @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
```
JakeWharton: Isn't it just `parameter == null || parameter.type == property.returnType`? That's not bad...

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Will fix in a follow-up, didn't know you were doing a pass before merging 😬

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Not quite - the null smartcast no longer works when this is done too. On mobile but in one of these I think it was fine, but in the other it meant nullchecking it's mention in the message

PR #900, opened 2019-09-08
Refactor missing/null properties to helper utils instead

ZacSweers:
This simplifies generated code quite a bit, and just uses a simple String.format rather than stringbuilder + avoiding string allocations

Before/after: https://gist.github.com/ZacSweers/90a4fb102ee6b9da1c5347f50b81d6af

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/AdapterGenerator.kt
@@ -31,7 +31,6 @@ import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
 import com.squareup.kotlinpoet.joinToCode
 import com.squareup.moshi.JsonAdapter
-import com.squareup.moshi.JsonDataException
 import com.squareup.moshi.JsonReader
 import com.squareup.moshi.JsonWriter
 import com.squareup.moshi.Moshi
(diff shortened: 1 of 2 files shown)


rharter:
I realised in reviewing this that the old one wasn't using the code gen knowledge we have, causing more allocations that required and looking overly complicated. Since we generate the code that prints the errors, the strings should have been generated by the generator, not at runtime.

Instead of

```
throw JsonDataException(StringBuilder("Non-null value '").append("lastName").append("' was null at ").append(reader.path).toString())
```

It could have been 

```
throw JsonDataException(String.format("Non-null value 'lastName' was null at %s", reader.getPath()))
```

Since we knew the property name at code gen time, there's no reason to append it at runtime.  This change cleans that up even more, so nice work.

JakeWharton:
@rharter actually the design was intentional!

For one, we don't care about allocation when we're about to throw and exception. Not only are exceptions extremely expensive to create themselves in terms of allocation, but they're meant to be created only in exceptional cases. Not something worth optimizing for.

But with regard to the string chunks, by using separate strings for the constant parts of the message and the dynamic parts, you're actually making the resulting APK smaller. The dex file format has a global string pool so the common strings are deduplicated and only paid for once. The property name is probably already free since it shares its name with the backing field.

More details at https://jakewharton.com/the-economics-of-generated-code/#string-duplication

JakeWharton:
> just uses a simple String.format rather than stringbuilder + avoiding string allocations

It was my understand that `String.format` is usually worse in these simple cases, again not that it matters.

rharter:
Ahh, yes, I seem to remember that discussion (or your blog post), now.

PR #901, opened 2019-09-08
Fast followups from #899

ZacSweers:
Followup fixes from the previous PR

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -213,18 +213,14 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       val parameter = parametersByName[property.name]
 
       if (Modifier.isTransient(property.javaField?.modifiers ?: 0)) {
-        @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
-        if (parameter != null && !parameter.isOptional) {
-          throw IllegalArgumentException(
-              "No default value for transient constructor $parameter")
+        require(parameter == null || parameter.isOptional) {
+          "No default value for transient constructor $parameter"
         }
         continue
       }
 
-      @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
-      if (parameter != null && parameter.type != property.returnType) {
-        throw IllegalArgumentException("'${property.name}' has a constructor parameter of type " +
-            "${parameter.type} but a property of type ${property.returnType}.")
+      require(parameter == null || parameter.type == property.returnType) {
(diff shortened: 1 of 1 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -213,18 +213,14 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
         }
         continue
       }
 
-      @Suppress("ReplaceGuardClauseWithFunctionCall") // More readable this way
-      if (parameter != null && parameter.type != property.returnType) {
-        throw IllegalArgumentException("'${property.name}' has a constructor parameter of type " +
-            "${parameter.type} but a property of type ${property.returnType}.")
+      require(parameter == null || parameter.type == property.returnType) {
+        "'${property.name}' has a constructor parameter of type ${parameter!!.type} but a property of type ${property.returnType}."
```
ZacSweers: Compiler can't smartcast this `parameter` here

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -254,7 +250,9 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
     for (parameter in constructor.parameters) {
       val binding = bindingsByName.remove(parameter.name)
-      require(!(binding == null && !parameter.isOptional)) { "No property for required constructor $parameter" }
+      require(binding != null || parameter.isOptional) {
```
rharter: `binding != null`? Is that right?

ZacSweers: Yeah this is just the flipped version of the `&&` before it. Autogenerated from intelliJ's inspection

rharter: Ugh, I find the `require` difficult to parse since it's basically an else clause holding the condition, but my brain wants to read it as an `if`. That's gonna take some getting used to.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
JakeWharton: File a bug?

ZacSweers: Can do, but wondering if this is just a known limitation of contracts for now (e.g. no compound expressions)

PR #903, opened 2019-09-09
Migrate to KotlinPoet-metadata

ZacSweers:
This is a preparation PR for the upcoming KotlinPoet-metadata APIs, which replace the use of `kotlin-metadata` in Moshi code gen and are built on top of Jetbrains' official `kotlinx-metadata-jvm` library.

Highlights:
- Far simpler implementation in `metadata.kt`, mostly code deletion and reworking to use kotlinpoet-metadata APIs
- Removed use of auto-common, kotlin-metadata dependencies
- Better error messages. Before - we would point to extracted synthetic fields/methods/etc for some error messages that weren't actually pointing to the correct locations because they assumed java. Now it just keeps it simple by referencing the class and the property with a description of the issue with it.
- Incidentally fixed a bug along the way where `private var` constructor properties were apparently not allowable.
- Minor cleanups of awkward, now-unused parts of the extracted code-gen API

Some oddities:
- For `@Json` and `@JsonQualifier` annotations, we still have to do manual parsing but with some different behavior
  - For `@Json` - this is given to us by kotlinpoet-metadata, but there is no way to cleanly grep for a member in kotlinpoet's `AnnotationSpec` API by name. So instead we manually scrape this off. 
    - This results in another thing we have to tweak later - kotlinpoet automatically escapes the value as needed. If we have, say, a name with a dollar sign in it, this would potentially become double escaped if we tried to render it to a `%S` template again. Now we take this value as-is, manually escape the default `property name` as needed, and manually insert the value into the generated class as a `"%L"` literal instead with quotes.
- For `@JsonQualifier`s, we are given the annotations from Kotlinpoet-metadata, but then need to look up that underlying `classname` via `Elements#getTypeElement()`, and then do the the checks on those results to see if they're annotated with `@JsonQualifier`

This revealed two issues in KotlinPoet-metadata:
1. We no longer need to try to shortcut our handling of `List<out Any>`, as kotlinx-metadata gives this information to us correctly. PR: https://github.com/square/kotlinpoet/pull/779
2. KotlinPoet-metadata doesn't support parameter annotations yet. PR: https://github.com/square/kotlinpoet/pull/780

CC @rharter @Egorand


kotlin/codegen/pom.xml
```diff
@@ -136,6 +147,12 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <args>
+            <!-- So we can declare ourselves as users of KotlinPoetMetadataPreview -->
+            <arg>-Xuse-experimental=kotlin.Experimental</arg>
```
Egorand: Will `-Xuse-experimental=com.squareup.kotlinpoet.KotlinPoetMetadataPreview` work as well?

ZacSweers: That shoooould work too, I guess we can do that since we only use one preview annotation

ZacSweers: 3206cdd

ZacSweers:
Rebased and built on kotlinpoet 1.4.0 final, this is ready to go 🏗

kotlin/codegen/pom.xml
```diff
@@ -136,6 +147,12 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <args>
+            <!-- So we can declare ourselves as users of KotlinPoetMetadataPreview -->
+            <arg>-Xuse-experimental=com.squareup.kotlinpoet.metadata.KotlinPoetMetadataPreview</arg>
```
swankjesse: 👍🏻

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -83,7 +81,8 @@ internal class AdapterGenerator(
       nameAllocator.newName("options"), JsonReader.Options::class.asTypeName(),
       KModifier.PRIVATE)
       .initializer("%T.of(${nonTransientProperties.joinToString(", ") {
-        CodeBlock.of("%S", it.jsonName).toString()
+        // We manually put in quotes because we know the jsonName is already escaped
```
swankjesse: Nice comment!

ZacSweers:
🎉

PR #909, opened 2019-09-11
Don't re-wrap nullsafe/nonnull adapters if they are already one

ZacSweers:
nullSafe() is often called defensively, which can lead to a lot of layers of wrapping and checking. This avoids that overhead where possible.


ZacSweers:
Derp, forgot. Added in b65827d

PR #911, opened 2019-09-11
Use raw type for kclass references in dynamic invocation code gen

### LINKED ISSUE #905: 1.9.0-SNAPSHOT: Error compiling generated adapter for generic class
Code like this is generated by the latest snapshot for generic class `Foo<T>`:
```kotlin
val localConstructor = this.constructorRef ?:
                Util.lookupDefaultsConstructor(Foo<T>::class.java).also {
                this.constructorRef = it }
```

Compiling this code results in error:
```
FooJsonAdapter.kt: Only classes are allowed on the left hand side of a class literal
```

ZacSweers:
Added a compile-only regression test. Didn't catch this before because we didn't have any test cases that were both generic _and_ used default values

Resolves #905


kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DefaultConstructorTest.kt
```diff
@@ -81,6 +81,13 @@ data class TestClass(
     val dynamicInlineOptional: Int = createInlineInt()
 )
 
+// Regression test for https://github.com/square/moshi/issues/905
+// Just needs to compile
+@JsonClass(generateAdapter = true)
+data class GenericTestClassWithDefaults(
```
technoir42: Missed `<T>`

ZacSweers: d'oh. And doing this locally reveals it's not actually fixed. Good catch!

Will wait until #908 is merged because the code gen changes there will make this easier

PR #913, opened 2019-09-11
Don't try to access unsafe APIs on android 29+

### LINKED ISSUE #872: Android Q compatibility policy violation
Device running android 9
Moshi version moshi-kotlin:1.8.0

Android P reports non sdk api uasge warnings when using the library:

The warnings are marked as light greylist and they are just warnings for now, but might be restricted in later versions as it is currently for the dark greylist warnings which are not going to work on android Q as announced.

See also https://developer.android.com/preview/restrictions-non-sdk-interfaces.html for a more detailed explanation.

StackTrace
```
StrictMode policy violation: android.os.strictmode.NonSdkApiUsedViolation: Lsun/misc/Unsafe;->theUnsafe:Lsun/misc/Unsafe;
        at android.os.StrictMode.lambda$static$1(StrictMode.java:428)
        at android.os.-$$Lambda$StrictMode$lu9ekkHJ2HMz0jd3F8K8MnhenxQ.accept(Unknown Source:2)
        at java.lang.Class.getDeclaredField(Native Method)
        at com.squareup.moshi.ClassFactory.get(ClassFactory.java:63)
        at com.squareup.moshi.ClassJsonAdapter$1.create(ClassJsonAdapter.java:79)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:137)
        at com.squareup.moshi.ClassJsonAdapter$1.createFieldBindings(ClassJsonAdapter.java:99)
        at com.squareup.moshi.ClassJsonAdapter$1.create(ClassJsonAdapter.java:82)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:137)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:97)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:67)
```

Thanks for support!

ZacSweers:
Resolves #872. Not sure if we want to try caching some of these lookups

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
@@ -78,7 +78,8 @@ abstract class ClassFactory<T> {
         final Method allocateInstance = unsafeClass.getMethod("allocateInstance", Class.class);
         return new ClassFactory<T>() {
           @SuppressWarnings("unchecked")
-          @Override public T newInstance() throws InvocationTargetException, IllegalAccessException {
+          @Override
+          public T newInstance() throws InvocationTargetException, IllegalAccessException {
             return (T) allocateInstance.invoke(unsafe, rawType);
           }
           @Override public String toString() {
--- a/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
@@ -54,29 +54,42 @@ abstract class ClassFactory<T> {
       // No no-args constructor. Fall back to something more magical...
     }
 
-    // Try the JVM's Unsafe mechanism.
-    // public class Unsafe {
-    //   public Object allocateInstance(Class<?> type);
-    // }
+    boolean tryUnsafe = true;
     try {
-      Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
-      Field f = unsafeClass.getDeclaredField("theUnsafe");
-      f.setAccessible(true);
-      final Object unsafe = f.get(null);
-      final Method allocateInstance = unsafeClass.getMethod("allocateInstance", Class.class);
-      return new ClassFactory<T>() {
-        @SuppressWarnings("unchecked")
-        @Override public T newInstance() throws InvocationTargetException, IllegalAccessException {
-          return (T) allocateInstance.invoke(unsafe, rawType);
-        }
-        @Override public String toString() {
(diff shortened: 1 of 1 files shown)


moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -54,29 +54,43 @@ abstract T newInstance() throws
...
-        }
-        @Override public String toString() {
-          return rawType.getName();
-        }
-      };
-    } catch (IllegalAccessException e) {
-      throw new AssertionError();
-    } catch (ClassNotFoundException | NoSuchMethodException | NoSuchFieldException ignored) {
-      // Not the expected version of the Oracle Java library!
+      Class<?> androidBuildClass = Class.forName("android.os.Build.VERSION");
```
JakeWharton: Shouldn't it be?
```suggestion
      Class<?> androidBuildClass = Class.forName("android.os.Build$VERSION");
```

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -54,29 +54,43 @@ abstract T newInstance() throws
...
+      Field sdkIntField = androidBuildClass.getDeclaredField("SDK_INT");
+      int sdk = (int) sdkIntField.get(null);
+      if (sdk >= 29) {
+        tryUnsafe = false;
+      }
+    } catch (Exception e) {
+      // Ignore, not running on android
```
JakeWharton: We usually just name the variable `ignore` to avoid needing a comment.

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -54,29 +54,43 @@ abstract T newInstance() throws
       // No no-args constructor. Fall back to something more magical...
     }
 
-    // Try the JVM's Unsafe mechanism.
-    // public class Unsafe {
-    //   public Object allocateInstance(Class<?> type);
-    // }
+    boolean tryUnsafe = true;
```
JakeWharton: Make it a static field?

ZacSweers: Would we need to make it nullable to know if it's been set or not?

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
ZacSweers: dcb9f15

PR #915, opened 2019-09-12
Update KotlinPoet to 1.3.0

### LINKED ISSUE #848: Generated serialization for field named with hot keywords
I use `class` as response field. In kotlin: 

```
@Keep
data class StaticLabelField(
        override val key: String,
        override val label: String,
        override val order: Int,
        override val condition: Condition? = null,
        val `class`: List<String>?
) : Field(), Validatable {
    override fun validator() = null
}
```
When using moshi to generate adapter code, it's not 'blocking/escaping' that field in `toJson` method: 

```
override fun toJson(writer: JsonWriter, value: Field.StaticLabelField?) {
    if (value == null) {
    }
    writer.beginObject()
    writer.name("key")
    stringAdapter.toJson(writer, value.key)
    writer.name("label")
    stringAdapter.toJson(writer, value.label)
    writer.name("order")
    intAdapter.toJson(writer, value.order)
    writer.name("condition")
    nullableConditionAdapter.toJson(writer, value.condition)
    writer.name("class")
    nullableListOfStringAdapter.toJson(writer, value.class) // <==== field with name 'class' has different result than intended.
    writer.endObject()
}
```

I worked it around by using `Json` annotation and different name for the field, but I think this is a bug that might be worth fixing.

ZacSweers:
Fixes #848 and adds a regression test case for it

### DIFF
--- a/kotlin/codegen/pom.xml
+++ b/kotlin/codegen/pom.xml
@@ -25,7 +25,7 @@
     <dependency>
       <groupId>com.squareup</groupId>
       <artifactId>kotlinpoet</artifactId>
-      <version>1.2.0</version>
+      <version>1.3.0</version>
     </dependency>
     <dependency>
       <groupId>net.ltgt.gradle.incap</groupId>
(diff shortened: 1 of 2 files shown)


ZacSweers:
CC @Egorand

PR #917, opened 2019-09-12
Report json name if different from property name in kotlin

### LINKED ISSUE #800: Codegen uses wrong name in generated exception message

ZacSweers:
Resolves #800 

Want to get feedback on the wording before going and adding tests. This also unifies reflection and code gen's error messaging

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/internal/Util.java
+++ b/moshi/src/main/java/com/squareup/moshi/internal/Util.java
@@ -49,7 +49,7 @@ public final class Util {
       = "Required value '%s' (JSON name '%s') missing at %s";
   private static final String UNEXPECTED_NULL_TEMPLATE = "Non-null value '%s' was null at %s";
   private static final String UNEXPECTED_NULL_WITH_NAME_TEMPLATE
-      = "Non-null value '%s' (json name '%s') was null at %s";
+      = "Non-null value '%s' (JSON name '%s') was null at %s";
   public static final Set<Annotation> NO_ANNOTATIONS = Collections.emptySet();
   public static final Type[] EMPTY_TYPE_ARRAY = new Type[] {};
   @Nullable private static final Class<?> DEFAULT_CONSTRUCTOR_MARKER;
--- a/moshi/src/main/java/com/squareup/moshi/internal/Util.java
+++ b/moshi/src/main/java/com/squareup/moshi/internal/Util.java
@@ -45,7 +45,11 @@ import static com.squareup.moshi.Types.supertypeOf;
 
 public final class Util {
   private static final String REQUIRED_PROPERTY_TEMPLATE = "Required property '%s' missing at %s";
+  private static final String REQUIRED_PROPERTY_NAME_TEMPLATE
+      = "Required property '%s' (json name '%s') missing at %s";
   private static final String UNEXPECTED_NULL_TEMPLATE = "Non-null value '%s' was null at %s";
+  private static final String UNEXPECTED_NULL_WITH_NAME_TEMPLATE
+      = "Non-null value '%s' (json name '%s') was null at %s";
   public static final Set<Annotation> NO_ANNOTATIONS = Collections.emptySet();
   public static final Type[] EMPTY_TYPE_ARRAY = new Type[] {};
   @Nullable private static final Class<?> DEFAULT_CONSTRUCTOR_MARKER;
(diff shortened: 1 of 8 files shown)


moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -567,17 +565,35 @@ public static String typeAnnotatedWithAnnotations(Type type,
     throw new IllegalStateException("No defaults constructor found for " + targetClass);
   }
 
-  public static JsonDataException missingProperty(String property, JsonReader reader) {
```
swankjesse: Because we call this from generated code we cannot delete it!

swankjesse: We can’t guarantee that all generated code is generated with the same version, or that the generated code matches the runtime version. Upgrading the runtime version shouldn’t be dangerous.

ZacSweers: would you rather just inline these directly?

JakeWharton: It creates a lot of code. I would just re-add this overload, pass `null` for the `jsonName`, and then add a null check so the first branch of the conditional is taken.

ZacSweers: Just to note - this method was only added since the last release in #900 and could be safely deleted if we want. That's why I was fine making the change here

JakeWharton: Sounds good. We should set up japicmp...

ZacSweers:
Most of the tests for this are identical in code gen and reflection, so going to take this opportunity to merge them into a parameterized shared test that checks both against the same code. We can start moving other shared behavior tests into this over time later

ZacSweers:
Done in 46a7659

PR #923, opened 2019-09-13
Added codegen support for inline classes

### LINKED ISSUE #776: Support Kotlin inline classes when they become stable
Moshi with Kotlin reflection wasn't working at all, so I tried to switch to Codegen. Everything is working good except Inline classes:
`inline class Seconds(val value: Long)`
with my adapter:
`class SecondsAdapter() {
    @FromJson
    fun fromJson(jsonReader: JsonReader, delegate: JsonAdapter<Seconds>): Seconds {
        val long = jsonReader.nextLong()
        return Seconds(long)
    }
}`
Error:
java.lang.ClassCastException: java.lang.Long cannot be cast to pl.amistad.stratapp.time.Seconds
Am I doing something wrong, or Moshi isn't ready yet for inline classes?

Wasabi375:
This was discussed in #776. The fix is really simple. I also added tests since I am using this in my own projects. 
I'm not sure if you want to pull this into master before inline classes become stable. To run the tests I enabled inline classes for the moshi-kotlin-test module. Also I don't think this will break unless they change the protobuff flag in the metadata declaring a class to be inline, which I think is very unlikely. 
Also this might not fix #776 completely since I haven't looked into using inline classes with the reflection based adapter.
___
For everyone else finding this, if this isn't merged into master [here](https://jitpack.io/#Wasabi375/moshi/inline_class-SNAPSHOT) is a link to jitpack where you can get the current version of the build. I will try to keep this up with master (as I am using it), but no promises :wink:

### DIFF
--- a/kotlin/tests/pom.xml
+++ b/kotlin/tests/pom.xml
@@ -117,6 +117,7 @@
           <args>
             <arg>-Werror</arg>
             <arg>-Xuse-experimental=kotlin.ExperimentalStdlibApi</arg>
+              <arg>-XXLanguage:+InlineClasses</arg>
           </args>
         </configuration>
       </plugin>
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
@@ -252,4 +252,21 @@ class DualKotlinTest(useReflection: Boolean) {
     assertThat(decoded.a).isEqualTo(null)
   }
 
+  @Test fun inlineClass() {
+    val adapter = moshi.adapter<InlineClass>()
+
+    val inline = InlineClass(5)
+
+    val expectedJson = """{"f":5}"""
+    assertThat(adapter.toJson(inline)).isEqualTo(expectedJson)
+
+    val testJson = """{"f":6}"""
+    val result = adapter.fromJson(testJson)!!
+    assertThat(result.f).isEqualTo(6)
+  }
+
 }
+
+// Has to be outside since inline classes are only allowed on top level
+@JsonClass(generateAdapter = true)
+inline class InlineClass(val f: Int)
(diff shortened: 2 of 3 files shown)


ZacSweers:
Happy to revisit this once inline classes are stable!

ZacSweers:
Three high level thoughts:
- This should have a test for a class with a property that is inline
- We'd want to support this in reflection too
- We are going to move to kotlinpoet's upcoming metadata support via #903

Wasabi375:
> We'd want to support this in reflection too

Well, if you are lucky I'll need reflection with inline classes in my projects. In that case I will probably add it, so there is a good chance I have this before inline classes are stable 😜

kotlin/tests/pom.xml
```diff
@@ -117,6 +117,7 @@
           <args>
             <arg>-Werror</arg>
             <arg>-Xuse-experimental=kotlin.ExperimentalStdlibApi</arg>
+              <arg>-XXLanguage:+InlineClasses</arg>
```
ZacSweers:
```suggestion
              <arg>-Xinline-classes</arg>
```

There's a newer arg for this

ZacSweers:
Commented on the issue, but if you want to rebase this I'm happy to merge it as a regression test. If possible, after you rebase can you move the test to the new `DualKotlinTest` to test this on both reflection and code gen? I thiiiiink it should Just Work with reflection as well.

ZacSweers:
@Wasabi375 are you planning to move this to the `DualKotlinTest` or would prefer that be a different PR?

kotlin/tests/pom.xml
Wasabi375: This fails with
```
[WARNING] Flag is not supported by this version of the compiler: -Xinline-classes
[ERROR] warnings found and -Werror specified
```

Wasabi375:
Before you merge this into master I would like to add a test case for this as well
> This should have a test for a class with a property that is inline

But that will have to wait until I come back from university later today

kotlin/tests/pom.xml
ZacSweers: Interesting, works for me with gradle.

ZacSweers:
Thanks!

PR #925, opened 2019-09-15
Introduce skipNextName() APIs to JsonReader/JsonWriter

### LINKED ISSUE #858: PolymorphicJsonAdapterFactory, .failOnUnknown() and Kotlin's Sealed Classes
Currently having an issue with `PolymorphicJsonAdapterFactory`, .failOnUnknown() and Kotlins `sealed class`s.
My situation: I have a JSON string with a list of objects which are subtypes of `MyClass` and i want to convert this to a Kotlin List of these subclasses. I also want to enable .failOnUnknown on my adapter when debugging (Due to still being in development and i want my tests to fail if the parsing encounters unexpected labels). 
The MyClass and it's subtypes
```
enum class MyType {
    ClassA,
    ClassB
}

sealed class MyClass(
        @Json(name = "type") val type: MyType
) {
    class A(
            @Json(name = "type") type: MyType,
            val value: Double
    ) : MyClass(type)
    
    class B(
            @Json(name = "type") type: MyType,
            val text: String
    ) : MyClass(type)
}
```
My adapter factory setup:
```
PolymorphicJsonAdapterFactory.of(MyClass::class.java, "type")
        .withSubtype(
                MyClass.A::class.java,
                MyType.ClassA.name
        )
        .withSubtype(
                MyClass.B::class.java,
                MyType.ClassB.name
        )
```
My JsonAdapter setup
```
Moshi
        .Builder()
        .add(adapterFactory)
        .build()
        .adapter<MyClass>(MyClass::class.java)
        .failOnUnknown()
```

Everything works well, if the first property of the JSON objects in the list is the type label `[{ "type": "ClassA", "value": 0.1 }]`, but if the JSON is structured like this: `[{ "value": 0.1 , "type": "ClassA" }]`, it fails due `failOnUnknown` being enabled and hitting the `value` label first, which doesn't exist in `MyClass`.

Any thoughts on this - is there a better approach? :)

### LINKED ISSUE #874: Feature Request: PolymorphicJsonAdapter flag to suppress automatic inclusion of label in toJson()
I've a series of classes that contains a `type` attribute (string). The Polymorphic adapter is great to parse and serialize those, but I sometimes need to use a specific adapter to do the same thing.

This means I need the autogenerated adapter for the actual type to write the `type` attribute (I cannot mark it as `@Transient`) but I do not want the Polymorphic adapter to produce a double `type` attribute.

```
interface DataItem {
  val type: String
}

data class Text(
  val text: String
): DataItem {
      override val type: String = "text"
}

data class Image(
  val url: String,
  val w: Int,
  val h: Int
): DataItem {
    override val type: String = "image"
}
```

If I setup the Polymorphic adapter for these types:

```
PolymorphicJsonAdapterFactory.of(DataItem::class.java, "type")
            .withSubtype(Text::class.java, "text")
            .withSubtype(Image::class.java, "image")
```

when I serialize I get:

```
{
    "type": "text",
    "type": "text",
    "text": "foo"
}
```

if I mark the `type` field as `@Transient` and then use the generated `Text` adapter directly I get

```
{
   "text": "foo"
}
```

I'd like to obtain
```
{
    "type": "text"
    "text": "foo"
}
```

in both cases.

Something like this?
```
PolymorphicJsonAdapterFactory.of(DataItem::class.java, "type")
            .withSubtype(Text::class.java, "text")
            .withSubtype(Image::class.java, "image")
            .skipLabelKeyInSerialization()
```

And the code in the PolymorphicAdapter could be changed to
```
      Class<?> type = value.getClass();
      int labelIndex = subtypes.indexOf(type);
      // ...
      JsonAdapter<Object> adapter = jsonAdapters.get(labelIndex);
      writer.beginObject();
      if (!skipLabelKey) {
          writer.name(labelKey).value(labels.get(labelIndex));
      }
      int flattenToken = writer.beginFlatten();
      adapter.toJson(writer, value);
      writer.endFlatten(flattenToken);
      writer.endObject();
    }
```

ZacSweers:
This is a proposal implementation for resolving two issues with PolymorphicJsonAdapter usage

The first is #858, and is resolved by marking a processed type label key to be skipped if the reader has failOnUnknown() set to true

The second is #874, and is resolved by marking that a type label has already been written and should be skipped if any delegate adapters try to write it as well. This implementation needs a bit more work to make it only apply on the first level object and not nested ones, but starting with the current implementation for review and feedback. I think this one is definitely worth pursuing because it affects any case where a subtype declares the type label.

Both implementations need more tests in the respective reader/writers, but will circle to those if we want to move forward with these APIs.

Resolves #858
Resolves #874

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -245,6 +245,9 @@ public final class PolymorphicJsonAdapterFactory<T> implements JsonAdapter.Facto
         reader.skipValue();
         return defaultValue;
       }
+      if (reader.failOnUnknown()) {
+        reader.skipNextName(labelKey);
+      }
       return jsonAdapters.get(labelIndex).fromJson(reader);
     }
 
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -291,6 +291,7 @@ public final class PolymorphicJsonAdapterFactory<T> implements JsonAdapter.Facto
       JsonAdapter<Object> adapter = jsonAdapters.get(labelIndex);
       writer.beginObject();
       writer.name(labelKey).value(labels.get(labelIndex));
+      writer.skipNextName(labelKey);
       int flattenToken = writer.beginFlatten();
       adapter.toJson(writer, value);
       writer.endFlatten(flattenToken);
--- a/adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
+++ b/adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
@@ -263,6 +263,23 @@ public final class PolymorphicJsonAdapterFactoryTest {
     assertThat(decoded.value).isEqualTo("Okay!");
   }
 
+  // Subtype does have a type, polymorphic adapter writes the type label name and value, subtype's
+  // type should be ignored in serialization to avoid duplication.
+  // Regression test for https://github.com/square/moshi/issues/874
+  @Test public void missingTypeLabel_toJson() throws IOException {
+    Moshi moshi = new Moshi.Builder()
+        .add(PolymorphicJsonAdapterFactory.of(Message.class, "type")
+            .withSubtype(MessageWithType.class, "success"))
+        .build();
+    JsonAdapter<Message> adapter = moshi.adapter(Message.class);
+
+    MessageWithType message = new MessageWithType("success", "Okay!");
+    String encoded = adapter.toJson(message);
+    assertThat(encoded).isEqualTo("{\"type\":\"success\",\"value\":\"Okay!\"}");
+    MessageWithType decoded = (MessageWithType) adapter.fromJson(encoded);
+    assertThat(decoded.value).isEqualTo("Okay!");
+  }
+
   interface Message {
   }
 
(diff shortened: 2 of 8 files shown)


timmc:
I think this has a recursion problem, in that it doesn't limit at what stack depth the label-skipping happens. See what happens if I add the following class and test to the unit tests:

```java
  static final class Chaff implements Message {
    final Map<String, String> aaa = new HashMap<String, String>(){{ put("type", "aaa"); }};
    final Map<String, String> zzz = new HashMap<String, String>(){{ put("type", "zzz"); }};
  }

  @Test public void recursionLabelChaff() throws IOException {
        .add(PolymorphicJsonAdapterFactory.of(Message.class, "type")
            .withSubtype(Chaff.class, "chaff"))
        .build();
    JsonAdapter<Message> adapter = moshi.adapter(Message.class);

    Chaff message = new Chaff();
    String encoded = adapter.toJson(message);
    // This line fails:
    assertThat(encoded).isEqualTo("{\"type\":\"chaff\",\"aaa\":{\"type\":\"aaa\"},\"zzz\":{\"type\":\"zzz\"}");
    Chaff decoded = (Chaff) adapter.fromJson(encoded);
    assertThat(decoded.aaa.get("type")).isEqualTo("aaa");
    assertThat(decoded.zzz.get("type")).isEqualTo("zzz");
  }
```

Output:

```
Expected :"{"type":"chaff","aaa":{"type":"aaa"},"zzz":{"type":"zzz"}"
Actual   :"{"type":"chaff","aaa":{},"zzz":{"type":"zzz"}}"
```

ZacSweers:
Yes, I mentioned in the original description that it's not implemented yet pending feedback on this proposal :)

> This implementation needs a bit more work to make it only apply on the first level object and not nested ones

timmc:
Ah jeez, I guess I skipped over that. -.- 

So with this implementation, you'd save off the setting in the scopes stack, then?

ZacSweers:
Yeah something like that would need to be done. Likely just clearing it when the scope changed

swankjesse:
I think this is a lot of management. Would it be better for impacted use cases to use a tree model instead of a streaming model? Then we could selectively strike the unwanted names.

I’m mostly reluctant about changes where every single read needs a new rarely-exercised branch. That seems fast but it might actually be devastating to our performance.

swankjesse:
One other option to consider – we could hide this in the failOnUnknown branch. So if we’re about to fail because there’s an unexpected name only then we pop the name we wanted to skip. It’s still a bit gnarly because of the stack management though.

ZacSweers:
@swankjesse has a better idea for solving this

timmc:
On stack management, I wonder if the `path` could be repurposed for this -- you're already managing an explicit stack.

PR #934, opened 2019-09-25
Proposal: object support via ObjectJsonAdapter

ZacSweers:
This is a proposal implementation of a new `ObjectJsonAdapter` that can be used to handle serialization of Kotlin `object` types. This is useful for sentinel types, and I've included a neat demo under `examples` in 9a10638.

In short - the idea is that there are cases where you either receive types with a single sentinel semantic value, or you can have cases of canary keys whose' values are just `{}` to show they are present. This is hard to model nicely in Java, but Kotlin's `objects` fit this space nicely.


ZacSweers:
Open question - should we support JsonQualifiers? My sense is no but just want to validate that in case there's a use case I'm missing.

ZacSweers:
And if not - should we error on objects with qualifiers to let them know they don't do what they expect?

dlew:
My understanding of JsonQualifiers is to let one define a custom adapter. In this case the adapter can only ever do one thing, which is decode to a specific `object`. It doesn't really make sense to me that you'd want to do a custom adapter for when you're purposefully constraining yourself to a singleton?

Providing a nice error message would probably be appreciated in that case, though.

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -642,14 +642,27 @@ class KotlinJsonAdapterTest {
...
     val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+    val adapter = moshi.adapter<ObjectDeclaration>()
+    assertThat(adapter.toJson(ObjectDeclaration)).isEqualTo("{}")
+    assertThat(adapter.fromJson("{}")).isSameAs(ObjectDeclaration)
+  }
+
+  @Test fun objectDeclarations_withJsonContent() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+    val adapter = moshi.adapter<ObjectDeclaration>()
+    assertThat(adapter.fromJson("""{"a":6}""")).isSameAs(ObjectDeclaration)
```
rharter: The unfortunate part about this is that in this case reading and writing is now asymmetric. Since we can't hide the ignored content somewhere in the instance if you read this from an external source, then try to send it back, they'll get something different then what you started with.  

I'm not sure how that could be mitigated unless there was a requirement that the `ObjectDeclaration` conformed to some sort of interface that allowed storing that extra data.  So the question is, is that important?  If so, then I think we should fail in this case, only supporting empty sentinel types.  If not, we can ignore it.

rharter: Unless, of course, that's what `failOnUnknown()` is for.  I'm not familiar with that.

ZacSweers:
Hmm, thinking on this more - this only works right now if the type is used for a property inside another kotlin class. There's no support for a top-level object type. This might be fine, but it begs the question of how we can support this in code gen if the enclosing type doesn't have a generated adapter. 

Something like @rharter's suggestion above with an opt-in `ObjectDeclaration` could help cover that case too, but it has a different issue: how do you get the singleton instance without reflection?

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
ZacSweers: When I talked about this idea with @swankjesse some time back, the idea was that `failOnUnknown()` would be usable to enforce this. That would break the example in this case, but that's a separate [known issue](https://github.com/square/moshi/pull/925). The only reason not to just immediately skip processing the keys entirely (which are always skipped here) is so that `failOnUnknown()` can catch them. If we can resolve #925 (Jesse has an idea for this), then we could even just always enable that in this adapter's scope.

As for asymmetric - I think it's only asymmetric at the JSON level and not the Moshi/adapter level. There's a precedent for this already for unknown JSON keys (i.e. enable `failOnUnknown()` if you care) and transient properties for data that never make it to the written JSON.

ZacSweers:
We could just reuse `@JsonClass(generateAdapter = true)` and generate a class like this

```kotlin
class CustomObjectJsonAdapter : ObjectJsonAdapter(CustomObject)
```

Then in the reflective lookup - if the looked up adapter class is assignable from ObjectJsonAdapter, invoke a no-arg constructor and return that.

Globegitter:
@ZacSweers Any progress on this - this would be really useful for one of our use cases.

ZacSweers:
Nothing currently. Out of curiosity, are any of your use cases outside of a polymorphic context?

Globegitter:
I am not sure i fully understand, my use case is exactly the same as in your example.

ZacSweers:
The example is using the polymorphic adapter so I’m assuming that means you aren’t looking for this outside of that context. That’s kind of what I’ve been leaning toward - object support being more of a possible feature of a polymorphic adapter rather than standalone support.

Then in the polymorphic adapter, one would just specify the subtype object instances rather than the subtype classes.

moshi/src/main/java/com/squareup/moshi/internal/ObjectJsonAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+    reader.beginObject();
+    while (reader.hasNext()) {
+      reader.skipName();
+      reader.skipValue();
+    }
+    reader.endObject();
+    return instance;
+  }
+
+  @Override public void toJson(JsonWriter writer, @Nullable T value) throws IOException {
```
Iljo: Should there be some null checks here, as in other adapters? Something like:
``` kotlin
if (value == null) throw new NullPointerException("value was null! Wrap in .nullSafe() to write nullable values.");
```

PR #941, opened 2019-09-27
Finish readJsonString

Dkhusainov:
This PR finishes `readJsonString` implementation(though it might be incorrect/not catch some cases). @swankjesse could you check it out?


ZacSweers:
Sorry for the delay on this. We're going to go with a better/more flexible solution described in https://github.com/square/moshi/issues/675#issuecomment-536130040 (PRs welcome if anyone wants to contribute it sooner!)

Issue #100, opened 2015-10-14
JsonWriter.value() for (char[], offset, length) or CharSequence ? 


Lakedaemon:
JsonReader & JsonWriter handles utf8 values (and bytes) internally...which is good for performance reasons

Does it make sense to have a public api for utf8 ByteString/chars/CharSequence values for JsonReader/Jsonwriter, to prevent object allocation/garbage collection/unnecessary transformations ?

swankjesse:
Neat idea. Would probably be binary-incompatible to just use CharSequence everywhere. Sigh.

PR #943, opened 2019-09-30
Add JsonWriter.valueSink for streaming data

JakeWharton:
Unlike `value(BufferedSource)`, this retains the push-based nature of JsonWriter and allows layering other sinks (such as a base64-encoding Sink for binary data). Since `value(BufferedSource)` is trivially replaced by reading the source into the value sink, it is deprecated.

Potentially helps out #100 a bit.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
@@ -17,9 +17,11 @@ package com.squareup.moshi;
 
 import java.io.IOException;
 import javax.annotation.Nullable;
+import okio.Buffer;
 import okio.BufferedSink;
-import okio.BufferedSource;
+import okio.Okio;
 import okio.Sink;
+import okio.Timeout;
 
 import static com.squareup.moshi.JsonScope.DANGLING_NAME;
 import static com.squareup.moshi.JsonScope.EMPTY_ARRAY;
@@ -28,6 +30,7 @@ import static com.squareup.moshi.JsonScope.EMPTY_OBJECT;
 import static com.squareup.moshi.JsonScope.NONEMPTY_ARRAY;
 import static com.squareup.moshi.JsonScope.NONEMPTY_DOCUMENT;
 import static com.squareup.moshi.JsonScope.NONEMPTY_OBJECT;
+import static com.squareup.moshi.JsonScope.STREAMING_VALUE;
 
 final class JsonUtf8Writer extends JsonWriter {
 
@@ -273,16 +276,35 @@ final class JsonUtf8Writer extends JsonWriter {
     return this;
   }
 
-  @Override public JsonWriter value(BufferedSource source) throws IOException {
+  @Override public BufferedSink valueSink() throws IOException {
     if (promoteValueToName) {
       throw new IllegalStateException(
-          "BufferedSource cannot be used as a map key in JSON at path " + getPath());
+          "BufferedSink cannot be used as a map key in JSON at path " + getPath());
     }
     writeDeferredName();
     beforeValue();
-    sink.writeAll(source);
-    pathIndices[stackSize - 1]++;
-    return this;
+    pushScope(STREAMING_VALUE);
+    return Okio.buffer(new Sink() {
+      @Override public void write(Buffer source, long byteCount) throws IOException {
+        sink.write(source, byteCount);
+      }
+
+      @Override public void close() {
+        if (peekScope() != STREAMING_VALUE) {
+          throw new AssertionError();
+        }
+        stackSize--; // Remove STREAMING_VALUE from the stack.
+        pathIndices[stackSize - 1]++;
+      }
+
+      @Override public void flush() throws IOException {
+        sink.flush();
+      }
+
+      @Override public Timeout timeout() {
+        return Timeout.NONE;
+      }
+    });
   }
 
   /**
@@ -380,6 +402,7 @@ final class JsonUtf8Writer extends JsonWriter {
    */
   @SuppressWarnings("fallthrough")
   private void beforeValue() throws IOException {
+    int nextTop;
     switch (peekScope()) {
       case NONEMPTY_DOCUMENT:
         if (!lenient) {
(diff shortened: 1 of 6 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -273,16 +276,34 @@ private void writeDeferredName() throws IOException {
...
+        }
+        stackSize--; // Remove STREAMING_VALUE from the stack.
+        pathIndices[stackSize - 1]++;
+      }
+
+      @Override public void flush() {
+      }
+
+      @Override public Timeout timeout() {
+        return Timeout.NONE;
```
swankjesse: should this be the sink’s timeout?

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
```diff
@@ -226,21 +230,35 @@ public Object root() {
...
     }
-    pathIndices[stackSize - 1]++;
-    return this;
+    pushScope(STREAMING_VALUE);
+
+    final Buffer buffer = new Buffer();
+    return Okio.buffer(new ForwardingSink(buffer) {
+      @Override public void close() throws IOException {
+        if (peekScope() != STREAMING_VALUE || stack[stackSize] != null) {
+          throw new AssertionError();
```
swankjesse: Need to handle repeated close

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -273,16 +276,34 @@ private void writeDeferredName() throws IOException {
...
-    return this;
+    pushScope(STREAMING_VALUE);
+    return Okio.buffer(new Sink() {
+      @Override public void write(Buffer source, long byteCount) throws IOException {
+        sink.write(source, byteCount);
+      }
+
+      @Override public void close() {
+        if (peekScope() != STREAMING_VALUE) {
+          throw new AssertionError();
```
swankjesse: Need to handle repeated close

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
```diff
@@ -273,16 +276,34 @@ private void writeDeferredName() throws IOException {
...
+        }
+        stackSize--; // Remove STREAMING_VALUE from the stack.
+        pathIndices[stackSize - 1]++;
+      }
+
+      @Override public void flush() {
```
swankjesse: flush the underlying sink!

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -360,12 +360,32 @@ public final void setSerializeNulls(boolean serializeNulls) {
 
   /**
    * Writes {@code source} directly without encoding its contents.
-   * Since no validation is performed, {@link #setSerializeNulls} and other writer configurations
-   * are not respected.
    *
-   * @return this writer.
+   * @see #valueSink()
+   * @deprecated Use: {@code try (BufferedSink sink = writer.valueSink()) { source.readAll(sink): }}
```
swankjesse: I wouldn’t bother to deprecate; we’ll never remove it and it just punishes users for no reason

swankjesse: (We made a mistake by offering this bad API, but our users don’t need to adjust anything because there’s no reason we’d ever want to remove this.)

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -360,12 +360,32 @@ public final void setSerializeNulls(boolean serializeNulls) {
...
+    return this;
+  }
+
+  /**
+   * Returns a {@link BufferedSink} into which arbitrary data can be written without any additional
+   * encoding. You <b>must</b> call {@link BufferedSink#close()} before interacting with this
+   * {@code JsonWriter} instance again.
+   * <p>
+   * Since no validation is performed, options like {@link #setSerializeNulls} and other writer
+   * configurations are not respected (if the content is itself JSON).
```
swankjesse: Please remove `If this content is itself JSON`, the comment must be JSON or this document is malformed!

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
JakeWharton: I thought it weird that someone streaming a value could potentially change the timeout or deadline of the enclosing JSON write.

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.java
JakeWharton: Handled automatically by `Okio.buffer`

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
JakeWharton: Handled automatically by `Okio.buffer`. Test added.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
JakeWharton: I thought it would be weird that you could do this since you cannot do it anywhere else. I don't particularly mind hooking it up, but it didn't seem like something you should have access to.

swankjesse: We could potentially do the thing where you can constrain the timeouts but you can’t relax them?

JakeWharton: SGTM. Is there an example of this somewhere?

moshi/src/test/java/com/squareup/moshi/PromoteNameToValueTest.java
```diff
@@ -351,4 +351,21 @@
     writer.endObject();
     assertThat(factory.json()).isEqualTo("{\"a\":\"a value\"}");
   }
+
+  @Test public void writerValueSinkFails() throws Exception {
```
swankjesse: I think this is fair

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java
swankjesse: Neat

Issue #921, opened 2019-09-13
internal property without backing field leads to NPE - kotlin code gen


Wasabi375:
```kotlin
@JsonClass(generateAdapter = true)
class Test  {

	@Transient
	private var foo: Int = 5

	internal var bar
		get() = foo
		set(f: Int) { foo = f}
}
```
This crashes with a NPE during code generation. The problem as far as I can tell is that `TargetType.get(...)` generates a property for the Test class with backing field, getter and setter set to `null`.  That leads to an NPE when `TargetProperty.element` is called. The relevant part of the stacktrace is 

> kotlin.KotlinNullPointerException
        at com.squareup.moshi.kotlin.codegen.TargetProperty.getElement(TargetProperty.kt:61)
        at com.squareup.moshi.kotlin.codegen.TargetProperty.jsonQualifiers(TargetProperty.kt:124)
        at com.squareup.moshi.kotlin.codegen.TargetProperty.generator(TargetProperty.kt:95)
        at com.squareup.moshi.kotlin.codegen.JsonClassCodegenProcessor.adapterGenerator(JsonClassCodegenProcessor.kt:103)
        at com.squareup.moshi.kotlin.codegen.JsonClassCodegenProcessor.process(JsonClassCodegenProcessor.kt:89)
        at org.jetbrains.kotlin.kapt3.base.incremental.IncrementalProcessor.process(incrementalProcessors.kt)

I was looking into how to fix this, but I am not entirely sure. The problem is that kotlin obfuscates the names of  internal functions/properties. For properties it obfuscates getters/setters. This means internal properties with backing fields are fine, since the backing field does not get obfuscated. But as soon as there is no backing field or the property is backed by a Delegate the code gen crashes.

ZacSweers:
Thanks for the report. I'll look at this once the kotlinpoet-metadata move is complete, which should better inform things areas this

Wasabi375:
Seems like the move to kotlinpoet-metadata fixed all my problems I have with moshi. This is also no longer an issue. I kinda wish I started using moshi after this change, would have saved me a lot of time :wink: Oh well, I need it for a hobby project so it's not like it cost me anything :smile:

If you want I can again create a PR with a test case.

ZacSweers:
Test case would be good to prevent any regressions!

PR #947, opened 2019-10-03
added test for internal property without backing field (#921)

Wasabi375:
Test for #921, the fix for the bug seems to be part of #903


ZacSweers:
Thanks!

PR #952, opened 2019-10-15
Refactor ClassFactory to avoid trying certain APIs on known android versions that don't support them

### LINKED ISSUE #949: 1.9.0 - UnsupportedOperationException when deserialising java class with non-empty constructor
We have (had) some legacy java DTO code:
```java
public class ProblemDTO {
    private String message;

    ProblemDTO(String message) {
        this.message = message;
    }
}
```

The Moshi crashes on API 29 when trying to deserialise it:
```
 Caused by: java.lang.UnsupportedOperationException: ObjectStreamClass.getConstructorId(Class<?>) is not supported on SDK 28
        at java.io.ObjectStreamClass.getConstructorId(ObjectStreamClass.java:2421)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.squareup.moshi.ClassFactory.get(ClassFactory.java:110)
        at com.squareup.moshi.ClassJsonAdapter$1.create(ClassJsonAdapter.java:96)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:138)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:98)
        at com.squareup.moshi.Moshi.adapter(Moshi.java:72)
```

An example project is here - https://github.com/emartynov/MoshiTest

The fix for the issue was the conversion of java DTO to Kotlin.

ZacSweers:
This improves the android support in `ClassFactory` to not try grey-listed reflection APIs on 28+, and also gates gingerbread-only APIs to only try below API 10. This leverages some of the prior android support we'd added to support it.

Resolves #949, sort of. The workaround posted in it is the right solution, this just makes it not try something we know won't work.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
@@ -34,8 +34,10 @@ abstract class ClassFactory<T> {
   abstract T newInstance() throws
       InvocationTargetException, IllegalAccessException, InstantiationException;
 
+  private static volatile boolean isAndroid = false;
   private static volatile boolean androidChecked = false;
   private static volatile boolean tryUnsafe = true;
+  private static volatile int androidSdkInt = -1;
 
   public static <T> ClassFactory<T> get(final Class<?> rawType) {
     // Try to find a no-args constructor. May be any visibility including private.
@@ -57,18 +59,8 @@ abstract class ClassFactory<T> {
       // No no-args constructor. Fall back to something more magical...
     }
 
-    if (!androidChecked) {
-      try {
-        Class<?> androidBuildClass = Class.forName("android.os.Build$VERSION");
-        Field sdkIntField = androidBuildClass.getDeclaredField("SDK_INT");
-        int sdk = (int) sdkIntField.get(null);
-        if (sdk >= 29) {
-          tryUnsafe = false;
-        }
-      } catch (Exception ignored) {
-      } finally {
-        androidChecked = true;
-      }
+    if (!androidChecked && isAndroid()) {
+      tryUnsafe = !isAndroidSdkAtLeast(29);
     }
     if (tryUnsafe) {
       // Try the JVM's Unsafe mechanism.
@@ -150,5 +142,29 @@ abstract class ClassFactory<T> {
     }
 
     throw new IllegalArgumentException("cannot construct instances of " + rawType.getName());
+  }
+
+  private static boolean isAndroid() {
+    if (!androidChecked) {
+      try {
+        Class<?> androidBuildClass = Class.forName("android.os.Build$VERSION");
+        isAndroid = true;
+        Field sdkIntField = androidBuildClass.getDeclaredField("SDK_INT");
+        androidSdkInt = (int) sdkIntField.get(null);
+      } catch (Exception ignored) {
+      } finally {
+        androidChecked = true;
+      }
+    }
+    return isAndroid;
+  }
+
+  private static boolean isAndroidSdkAtLeast(int target) {
+    if (isAndroid()) {
+      int sdk = androidSdkInt;
+      return sdk != -1 && sdk >= target;
+    } else {
+      return true;
+    }
   }
 }
--- a/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassFactory.java
@@ -109,7 +109,8 @@ abstract class ClassFactory<T> {
         newInstance.setAccessible(true);
         return new ClassFactory<T>() {
           @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


ZacSweers:
I'll try this in an app targeting API 28+ to test this.

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -98,57 +90,91 @@ public T newInstance() throws InvocationTargetException, IllegalAccessException
...
-        @SuppressWarnings("unchecked")
-        @Override public T newInstance() throws InvocationTargetException, IllegalAccessException {
-          return (T) newInstance.invoke(null, rawType, Object.class);
-        }
-        @Override public String toString() {
-          return rawType.getName();
-        }
-      };
-    } catch (Exception ignored) {
+    if (isBelowAndroidApi(10)) {
```
ZacSweers:
```suggestion
    if (isBelowAndroidApi(9)) {
```

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -98,57 +90,91 @@ public T newInstance() throws InvocationTargetException, IllegalAccessException
...
-        @SuppressWarnings("unchecked")
-        @Override public T newInstance() throws InvocationTargetException, IllegalAccessException {
-          return (T) newInstance.invoke(null, rawType, Object.class);
-        }
-        @Override public String toString() {
-          return rawType.getName();
-        }
-      };
-    } catch (Exception ignored) {
+    if (isBelowAndroidApi(9)) {
```
ZacSweers: Note I can't test this because it doesn't seem like it's possible to download pre-JB android versions anymore via ADB manager

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -57,18 +59,8 @@ abstract T newInstance() throws
...
-        int sdk = (int) sdkIntField.get(null);
-        if (sdk >= 29) {
-          tryUnsafe = false;
-        }
-      } catch (Exception ignored) {
-      } finally {
-        androidChecked = true;
-      }
+    if (!androidChecked && isAndroid()) {
+      tryUnsafe = isBelowAndroidApi(29);
```
dlew: The code is setting up a complex volatile state here that is a bit hard to follow. E.g., one might wonder (as I did) why it's not just calling `isAndroid()`, since that method  *also* checks `!androidChecked` ahead of time.

What if we just sussed all these variables out at once? Like, write a method that resolves `androidChecked`, `isAndroid`, `tryUnsafe`, and `androidSdkInt` in one fell swoop. Then the code becomes easier to follow.

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -98,57 +90,91 @@ public T newInstance() throws InvocationTargetException, IllegalAccessException
...
+          }
+          @Override public String toString() {
+            return rawType.getName();
+          }
+        };
+      } catch (Exception ignored) {
+      }
     }
 
     throw new IllegalArgumentException("cannot construct instances of " + rawType.getName());
```
dlew: Since we have no solution for API 29+, maybe we should throw a more instructive exception in that case (pointing them to the workaround, i.e. "you are on API 29+, you need an empty constructor")?

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
```diff
@@ -98,57 +90,91 @@ public T newInstance() throws InvocationTargetException, IllegalAccessException
...
+        androidSdkInt = (int) sdkIntField.get(null);
+      } catch (Exception ignored) {
+      } finally {
+        androidChecked = true;
+      }
+    }
+    return isAndroid;
+  }
+
+  private static boolean isBelowAndroidApi(int target) {
```
dlew: Minor, but I found this method name confusing - what is "is" here?

ZacSweers: Cleaned up in 415ffb0

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
ZacSweers: Good call - 40b2ed3

moshi/src/main/java/com/squareup/moshi/ClassFactory.java
ZacSweers: Good idea - e793ecf and dcc427c

ZacSweers:
Going to do one more check on API 28 and API 27 devices before merging

ZacSweers:
Abandoning for now since we opted to revert covering these

dlew:
Just for future knowledge - the revert was in #964.

PR #954, opened 2019-10-17
Implement JsonReader#valueSource()

ZacSweers:
This is a reader analog for #943. The tests need a bit more fleshing out, but want to start with this for review.

At a high level
- JsonUtf8Reader works by basically replacing `skipValue()` with one that optionally accepts a `Sink`. `skipValue()` calls this method with a black hole, and `valueSource()` calls this with a `Buffer` that the returned `BufferedSource` ultimately consumes. A lot of `skip()`/`readByte()` calls are reworked to defer to this sink, so blackhole behavior just skips under the hood like before and valueSource() keeps the data in its buffer. I don't know if there's a lazier way to push this along though, as right now it eagerly reads the value like skipValue does.
- JsonValueReader works by just emitting utf8 values directly to a buffer that the returned source consumes. One interesting caveat here is that when creating value maps, it always uses doubles for numbers, which could create asymmetric inputs/outputs.

There's probably a couple places these need to check scope in other methods, I'm in unfamiliar territory with this PR so feedback welcome

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -959,6 +959,9 @@ final class JsonUtf8Reader extends JsonReader {
   }
 
   @Override public void close() throws IOException {
+    if (peekScope() == STREAMING_VALUE) {
+      throw new IllegalStateException("Sink from valueSource() was not closed");
+    }
     peeked = PEEKED_NONE;
     scopes[0] = JsonScope.CLOSED;
     stackSize = 1;
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -22,9 +22,16 @@ import javax.annotation.Nullable;
 import okio.Buffer;
 import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
+import okio.Sink;
+import okio.Source;
+import okio.Timeout;
+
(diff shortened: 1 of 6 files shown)


ZacSweers:
@JakeWharton has this stacked locally already on top of #943

tmm1:
hey @ZacSweers, whatever happened here? are you using this patch?

@dkhusainov are you using the patch in production?

ZacSweers:
You already asked on the other issue yesterday, please don't duplicate discussions and be patient

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -860,23 +876,31 @@ private String nextUnquotedValue() throws IOException {
...
+
+  private void skipQuotedValue(ByteString runTerminator, Sink sink) throws IOException {
     while (true) {
       long index = source.indexOfElement(runTerminator);
       if (index == -1L) throw syntaxError("Unterminated string");
 
       if (buffer.getByte(index) == '\\') {
-        buffer.skip(index + 1);
+        sink.write(buffer, index + 1);
         readEscapeCharacter();
```
tmm1, line 889: Should `readEscapeCharacter` have `sink` passed in ala https://github.com/infotech-group/moshi/commit/1d052e71e163b99a6f4b233f951edf9b71f646f8#diff-ba52b67c0349f268e571d765eab69d95

ZacSweers: Not all of those tests appeared to pass when I tried them in #1219

PR #957, opened 2019-10-19
Generate optimized toString()

ZacSweers:
This optimizes `toString()` functions to emit separate strings for the enclosing prefix and the adapter name

Example:

```kotlin
override fun toString(): String = buildString {
    append("GeneratedJsonAdapter(").append("SmokeTestType").append(")") }
```

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -40,6 +40,8 @@ import java.lang.reflect.Constructor
 import java.lang.reflect.Type
 
 private val MOSHI_UTIL = Util::class.asClassName()
+private const val TO_STRING_PREFIX = "GeneratedJsonAdapter("
+private const val TO_STRING_SIZE_BASE = TO_STRING_PREFIX.length + 1 // 1 is the closing paren
 
 /** Generates a JSON adapter for a target type. */
 internal class AdapterGenerator(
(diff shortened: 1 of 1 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -158,8 +159,13 @@ internal class AdapterGenerator(
...
         .addModifiers(KModifier.OVERRIDE)
         .returns(String::class)
-        .addStatement("return %S",
-            "GeneratedJsonAdapter(${originalTypeName.rawType().simpleNames.joinToString(".")})")
+        .addStatement(
+            "return %M·{ append(%S).append(%S).append(%S) }",
+            MemberName("kotlin.text", "buildString"),
+            "GeneratedJsonAdapter(",
+            originalTypeName.rawType().simpleNames.joinToString("."),
+            ")"
```
JakeWharton: Use a char?

(In the generated code, not here in the source)

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -158,8 +159,13 @@ internal class AdapterGenerator(
     return FunSpec.builder("toString")
         .addModifiers(KModifier.OVERRIDE)
         .returns(String::class)
-        .addStatement("return %S",
-            "GeneratedJsonAdapter(${originalTypeName.rawType().simpleNames.joinToString(".")})")
+        .addStatement(
+            "return %M·{ append(%S).append(%S).append(%S) }",
```
JakeWharton: Can also pass the exact size as the first argument to `buildString` to avoid expansion / over-allocation.

ZacSweers:
Addressed both above in e1873e2

Now looks like this

```kotlin
override fun toString(): String = buildString(35) {
    append("GeneratedJsonAdapter(").append("SmokeTestType").append(')') }
```

Issue #962, opened 2019-10-26
R8 shakes values(), which is required for the standard enum serializer


jzbrooks:
Our team is in the process of moving from hand-rolled enum adapters to using the standard enum adapter. It seems like others might run into something similar. We use kotlin codegen for our non-enum classes and the built-in handling for enums.

```
    java.lang.AssertionError: impossible
        at java.lang.Enum$1.create(Enum.java:272)
        at java.lang.Enum$1.create(Enum.java:262)
        at libcore.util.BasicLruCache.get(BasicLruCache.java:58)
        at java.lang.Enum.getSharedConstants(Enum.java:289)
        at java.lang.Class.getEnumConstantsShared(Class.java:2423)
        at java.lang.Class.getEnumConstants(Class.java:2409)
        at com.squareup.moshi.B$k.<init>(StandardJsonAdapters.java:3)
        at com.squareup.moshi.B$b.a(StandardJsonAdapters.java:24)
        at com.squareup.moshi.A.a(Moshi.java:27)
        at <internal stuff>
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at com.squareup.moshi.E.a.a(Util.java:103)
        at com.squareup.moshi.B$b.a(StandardJsonAdapters.java:22)
        at com.squareup.moshi.A.a(Moshi.java:27)
        at l.J.a.a.a(MoshiConverterFactory.java:13)
        at l.E.a(Retrofit.java:32)
        at l.C$a.a(RequestFactory.java:298)
        at l.F.a(ServiceMethod.java:1)
        at l.E.a(Retrofit.java:11)
        at l.E$a.invoke(Retrofit.java:5)
        at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
        at $Proxy0.a(Unknown Source)
        at f.c.a.c.a(Repository.kt:77)
        at <internal stuff>
        at j.p.g.a.a.a(ContinuationImpl.kt:3)
        at kotlinx.coroutines.N.run(Dispatched.kt:15)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.a(CoroutineScheduler.kt:54)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.a(CoroutineScheduler.kt:4)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$a.run(CoroutineScheduler.kt:40)
[...]

PR #963, opened 2019-10-26
Keep synthesized values() for EnumJsonAdapter

jzbrooks:
Addresses #962 

I removed the function modifiers mentioned in the issue as they didn't seem to be specifically related to the problem at hand. I also added it to the other enum related rule, rather than using the `enum *` matcher.

One less than ideal thing is that it would keep a hand written unused method with the same signature for Java consumers. I suspect that's a small edge case, but if @JakeWharton or others have a suggestion to improve the rule to avoid it, it could be valuable. As far as I can tell the Kotlin compiler doesn't generate a `synthetic` modifier for the generated method, which seems unfortunate.

### DIFF
--- a/moshi/src/main/resources/META-INF/proguard/moshi.pro
+++ b/moshi/src/main/resources/META-INF/proguard/moshi.pro
@@ -8,9 +8,11 @@
 -keep @com.squareup.moshi.JsonQualifier interface *
 
 # Enum field names are used by the integrated EnumJsonAdapter.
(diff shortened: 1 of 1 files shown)


JakeWharton:
All enums are required to generate this API because it's how `Enum.valueOf` works, but it's also just a generally useful API.

The default rules that AGP wants you to use (but you shouldn't) keeps all of these methods on every enum, so this targeted version is fine.

PR #967, opened 2019-10-27
Use generatedJsonAdapterName() in code gen for better behavior parity

ZacSweers:
It's not the prettiest form right now, but it ensures we're using the same expectations as the runtime API expects


swankjesse:
Not sure I get it. Wouldn't a test case be a nicer way to keep these two from diverging?

ZacSweers:
This is in code gen, so all the existing code gen tests are the test case no?

ZacSweers:
@swankjesse let me know if the above makes sense, I really think this would be a good way to bolt down our tests around this

ZacSweers:
Closed out because the `bestGuess()` usage isn't great and we get good coverage but the existing tests

PR #969, opened 2019-10-28
Promote Kotlin type-inferring APIs to the main Moshi package

swankjesse:
These don't require kotlin-reflect after all.


moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -72,6 +76,23 @@
...
   }
 
+  /**
+   * Returns a JSON adapter for {@code kType}, creating it if necessary. Note that while nullability
+   * of {@code kType} itself is handled, nested types (such as in generics) are not resolved.
+   */
+  @CheckReturnValue public <T> JsonAdapter<T> adapter(KType kType) {
+    JsonAdapter<T> adapter = adapter(toType(kType));
+    if (adapter instanceof NullSafeJsonAdapter || adapter instanceof NonNullJsonAdapter) {
+      // TODO CR - Assume that these know what they're doing? Or should we defensively avoid
```
ZacSweers: Worth probably deciding this while we're at it

swankjesse: Yep, I think we add a `TypeAdapter.Factory` that takes a KType and makes the adapters responsible for this decision.

ZacSweers:
Continued in #1202

PR #972, opened 2019-10-28
Update to KotlinPoet 1.4.2

### LINKED ISSUE #968: internal abstract property leads to IllegalArgumentException in kotlinpoet.metadata
```kotlin
abstract class InternalAbstractProperty {
  internal abstract val test: String
}

class Impl(override val test: String) : InternalAbstractProperty()
```
also reported here: https://github.com/square/kotlinpoet/issues/820

ZacSweers:
Fixes #968 

We have to manually unwrap typealiases now, as kotlinpoet now (correctly) reports the aliases and just leaves a tag to retrieve the underlying type as needed.


ZacSweers:
Seems sonatype is having trouble again... we can retry this after it's had a bit more time with the kotlinpoet release (I was working against snapshot for the impl)

PR #976, opened 2019-10-29
Fix support for classes w/ multiple constructors in code gen

### LINKED ISSUE #975: kotlin codgen: multiple constructors with default lead to crash
```kotlin
class Foo(val f: Int)

class Bar(val f: Foo = Foo(5), val b: Int) {
	constructor(f: Int, b: Int = 6): this(Foo(f), b)
}
@ExperimentalStdlibApi
fun main() {
	val moshi = Moshi.Builder().build()
	moshi.adapter<Bar>().fromJson("""{"b":6}""")
}
```
fails with
```
Exception in thread "main" java.lang.IllegalArgumentException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at moshiTest.BarJsonAdapter.fromJson(BarJsonAdapter.kt:55)
	at moshiTest.BarJsonAdapter.fromJson(BarJsonAdapter.kt:17)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:40)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:43)
	at moshiTest.MainKt.main(main.kt:46)
	at moshiTest.MainKt.main(main.kt)
```
The problem is that kotlin generates multiple methods with the `DefaultConstructorMarker` argument. That leads to a chance of the `fromJson` code choosing the wrong constructor in `Util.lookupDefaultsConstructor`.

ZacSweers:
This fixes #975 by directly looking up the constructor via `getDeclaredConstructors` and the exact parameter types + mask and default constructor marker.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -20,6 +20,7 @@ import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.CodeBlock
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.INT
 import com.squareup.kotlinpoet.KModifier
 import com.squareup.kotlinpoet.MemberName
 import com.squareup.kotlinpoet.NameAllocator
@@ -31,6 +32,7 @@ import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.TypeVariableName
 import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
+import com.squareup.kotlinpoet.joinToCode
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.JsonReader
 import com.squareup.moshi.JsonWriter
@@ -48,6 +50,14 @@ internal class AdapterGenerator(
     target: TargetType,
     private val propertyList: List<PropertyGenerator>
 ) {
+
+  companion object {
+    private val DEFAULT_CONSTRUCTOR_EXTRA_PARAMS = arrayOf(
+        CodeBlock.of("%T::class.javaPrimitiveType", INT),
+        CodeBlock.of("%T.DEFAULT_CONSTRUCTOR_MARKER", Util::class)
+    )
+  }
+
   private val nonTransientProperties = propertyList.filterNot { it.isTransient }
   private val className = target.typeName.rawType()
   private val visibility = target.visibility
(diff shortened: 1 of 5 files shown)


ZacSweers:
Considered other options but felt this was best

Pros:
- Proguard-safe
- No more runtime `Util` requirement, all code is directly in code gen
Cons:
- A bit dense in generated code

Other options considered:

- Look up constructor by jvm method signature
  - Extremely not proguard safe, checks every possible constructor via string comparison
- Add `@MoshiConstructor` annotation to designate which constructor should be used
  - Could make proguard rules easier, but unclear if this annotation is actually copied to the defaults constructor
  - Still requires iterating all constructors searching for it
  - Error prone - what happens if someone mistakenly applies it to multiple constructors?
  - Requires users to make code changes to work, when the regression right now was that it worked without configuration before.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -98,6 +109,25 @@ internal class AdapterGenerator(
       .initializer("null")
       .build()
 
+//  private val getOrFindConstructorRefFunction = FunSpec.builder(
```
swankjesse: remove

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
```diff
@@ -49,10 +53,41 @@ internal fun TypeName.defaultPrimitiveValue(): CodeBlock =
...
-      UNIT, Void::class.asTypeName() -> throw IllegalStateException("Parameter with void or Unit type is illegal")
+      UNIT, Void::class.asTypeName(), NOTHING -> throw IllegalStateException("Parameter with void, Unit, or Nothing type is illegal")
       else -> CodeBlock.of("null")
     }
 
+internal fun TypeName.asTypeBlock(): CodeBlock {
+  when (this) {
+    is ParameterizedTypeName -> {
+      return if (rawType == ARRAY) {
+        CodeBlock.of("%T::class.java", copy(nullable = false))
```
swankjesse: does this do the right thing? `List<String>::class.java` ?

ZacSweers: This is only for arrays, which are special. For other generic types it just takes the raw type.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: x

ZacSweers: 7e53156

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
ZacSweers: Resolving this since I think it was just a misunderstanding and want to build the multi-mask support on top of this, can revisit if you think the array handling is still problematic!

PR #978, opened 2019-10-30
Support 32+ parameter constructors in code gen

### LINKED ISSUE #977: data classes with over 32 properties need multiple masks
1.3.60-eap-76 with version 1.9.0 of codegen I started getting `java.lang.IllegalArgumentException: Wrong number of arguments` for data classes that have more than 32  properties.

the decompiled byte code for the synthetic default constructor expects an additional mask for each 32 count boundary crossed.

this issue is a follow up to https://github.com/square/moshi/issues/960

ZacSweers:
Resolves #977 using the methodology roughly described here: https://github.com/square/moshi/issues/977#issuecomment-547691422

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -40,7 +40,6 @@ import com.squareup.moshi.Moshi
 import com.squareup.moshi.internal.Util
 import java.lang.reflect.Constructor
 import java.lang.reflect.Type
-import kotlin.math.ceil
 
 private val MOSHI_UTIL = Util::class.asClassName()
 private const val TO_STRING_PREFIX = "GeneratedJsonAdapter("
@@ -198,7 +197,7 @@ internal class AdapterGenerator(
     }
 
     // Calculate how many masks we'll need. Round up if it's not evenly divisible by 32
-    val maskCount = ceil(propertyList.count { it.hasConstructorParameter }.toFloat().div(32)).toInt()
+    val maskCount = (propertyList.count { it.hasConstructorParameter } + 31) / 32
     // Allocate mask names
     val maskNames = Array(maskCount) { index ->
       nameAllocator.newName("mask$index")
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -197,7 +197,12 @@ internal class AdapterGenerator(
     }
 
     // Calculate how many masks we'll need. Round up if it's not evenly divisible by 32
-    val maskCount = (propertyList.count { it.hasConstructorParameter } + 31) / 32
+    val propertyCount = propertyList.count { it.hasConstructorParameter }
+    val maskCount = if (propertyCount == 0) {
+      0
+    } else {
+      (propertyCount + 31) / 32
+    }
     // Allocate mask names
     val maskNames = Array(maskCount) { index ->
       nameAllocator.newName("mask$index")
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -260,7 +260,7 @@ internal class AdapterGenerator(
         if (property.hasConstructorDefault) {
           val inverted = (1 shl maskIndex).inv()
           result.addComment("\$mask = \$mask and (1 shl %L).inv()", maskIndex)
-          result.addStatement("%1L = %1L and %2L", maskNames[maskNameIndex], inverted)
+          result.addStatement("%1L = %1L and 0x%2L.toInt()", maskNames[maskNameIndex], Integer.toHexString(inverted))
         } else {
           // Presence tracker for a mutable property
           result.addStatement("%N = true", property.localIsPresentName)
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -40,6 +40,7 @@ import com.squareup.moshi.Moshi
 import com.squareup.moshi.internal.Util
 import java.lang.reflect.Constructor
 import java.lang.reflect.Type
+import kotlin.math.ceil
 
 private val MOSHI_UTIL = Util::class.asClassName()
 private const val TO_STRING_PREFIX = "GeneratedJsonAdapter("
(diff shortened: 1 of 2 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -198,10 +197,18 @@ internal class AdapterGenerator(
       }
     }
 
-    val maskName = nameAllocator.newName("mask")
+    // Calculate how many masks we'll need. Round up if it's not evenly divisible by 32
+    val maskCount = ceil(propertyList.count { it.hasConstructorParameter }.toFloat().div(32)).toInt()
```
swankjesse: What's with the weird math? You should never need floats to do int math.

Conventional int math looks like this:
val maskCount = (propertyList.count { it.hasConstructorParameter } + 31) / 32

swankjesse: Also what's the behavior on zero parameters? 0 masks or 1?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -211,12 +218,31 @@ internal class AdapterGenerator(
...
+    //
+    // Track important indices for masks. Masks generally increment with each parameter (including
+    // transient).
+    //
+    // Mask name index is an index into the maskNames array we initialized above.
+    //
+    // Once the maskIndex reaches 32, we've filled up that mask and have to move to the next mask
+    // name. Reset the maskIndex relative to here and continue incrementing.
+    //
+    var maskIndex = 0
```
swankjesse: Don't you want just one variable, maskableIndex? Then you can do (maskableIndex / 32) and (maskableIndex % 32) to get the bits you want.

swankjesse: Tracking it this way feels awkward because it's less random-access to go from maskableIndex to the specific bits you want to adjust.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -234,7 +260,7 @@ internal class AdapterGenerator(
         if (property.hasConstructorDefault) {
           val inverted = (1 shl maskIndex).inv()
           result.addComment("\$mask = \$mask and (1 shl %L).inv()", maskIndex)
-          result.addStatement("%1L = %1L and %2L", maskName, inverted)
+          result.addStatement("%1L = %1L and %2L", maskNames[maskNameIndex], inverted)
```
swankjesse: It'd be nice to emit `inverted` in hex? `0x80` for bit 4 instead of 16?

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codegen/MultipleMasksTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    val arg44: Long = 44,
+    val arg45: Long = 45,
+    val arg46: Long = 46,
+    val arg47: Long = 47,
+    val arg48: Long = 48,
+    val arg49: Long = 49,
+    val arg50: Long = 50,
+    val arg51: Long = 51,
+    val arg52: Long = 52,
+    @Transient val arg53: Long = 53,
```
swankjesse: Nice.

swankjesse: Add a parameter with no default value?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: I was just rounding up, that approach works too. Good point on 0 parameters, will add a check for that.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: Either way works, I just thought this was a bit more readable. Which do you rather?

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/codegen/MultipleMasksTest.kt
ZacSweers: e303f66

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: b2b0a38

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: 497fdba for the math
7f4532e for the empty check. Worth mentioning that even before, it would just ignore the mask names in that case since it wouldn't use the dynamic invocation in that case

ZacSweers:
Going to merge this now to unblock #979. Can followup the `maskableIndex` bit if you want to go with that

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -198,10 +196,23 @@ internal class AdapterGenerator(
       }
     }
 
-    val maskName = nameAllocator.newName("mask")
+    // Calculate how many masks we'll need. Round up if it's not evenly divisible by 32
+    val propertyCount = propertyList.count { it.hasConstructorParameter }
+    val maskCount = if (propertyCount == 0) {
```
swankjesse: I don’t think you need the if here, 31/32 = 0 in int math

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
swankjesse: I think it’s less readable, but that’s cause I do lots of code like this and using divide+mod at read time feels natural to me.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -234,7 +264,7 @@ internal class AdapterGenerator(
         if (property.hasConstructorDefault) {
           val inverted = (1 shl maskIndex).inv()
           result.addComment("\$mask = \$mask and (1 shl %L).inv()", maskIndex)
```
swankjesse: I don’t think we need the comment? I don’t think it clarifies

PR #982, opened 2019-10-30
Handle non-property parameters in kotlin code gen

### LINKED ISSUE #979: Code gen: classes with non-property parameters  don't work with mask generation
This is a regression from 1.8.0, though I'm fairly certain this was working by coincidence in 1.8.0. We should try to fix this. All parameters (property or not) participate in mask generation and constructor invocation. 

We should handle these similarly to how we handle transient properties: increment the mask index, use them in constructor lookup, use a default value in invocation, but otherwise ignore them.

ZacSweers:
This adds a `FromJsonComponent` sealed type for describing the combination of properties, parameters, and parameter properties. This allows us to handle parameter-only types separately (needed for mask index calculation). Resolves #979

Same suggestions welcome... I couldn't think of what else to call this.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -28,6 +28,7 @@ import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.ParameterizedTypeName
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
 import com.squareup.kotlinpoet.PropertySpec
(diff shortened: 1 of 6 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -196,8 +202,34 @@ internal class AdapterGenerator(
...
+    for ((index, parameter) in targetConstructorParams) {
+      val property = propertiesByIndex[index]
+      if (property == null) {
+        components += ParameterOnly(parameter)
+      } else {
+        components += ParameterProperty(parameter, property)
+      }
+    }
+
+    // Now add the remaining properties that aren't parameters
```
swankjesse: I like this design

PR #992, opened 2019-10-31
Fix nullable properties

### LINKED ISSUE #990: 1.9.1 codegen no nullableListOfStringAdapter
Hi,

First of all, tanks for the fantastic work on the 1.9.x release! I'm not sure if this is an issue or if I've simply misunderstood what the expected behavior is.

I have a data class with a `List<String>?` which would produce the following line in the code-gen generated adapter on 1.8.0:
`14 -> genres = nullableListOfStringAdapter.fromJson(reader)`

With 1.9.1 code-gen produces this:

`        14 -> genres = listOfStringAdapter.fromJson(reader) ?: throw Util.unexpectedNull("genres",
            "genres", reader)`

This results in `com.squareup.moshi.JsonDataException: Required value 'genres' missing at $.programs[0].event`.

Have I missed something in the 1.9.X release for marking lists as optional?

ZacSweers:
We unfortunately regressed here by not propagating nullability to the unwrapped alias or parameterized type.

Resolves #990

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -52,6 +52,7 @@ import java.lang.annotation.ElementType
 import java.lang.annotation.Retention
 import java.lang.annotation.RetentionPolicy
 import java.lang.annotation.Target
+import java.util.TreeSet
 import javax.annotation.processing.Messager
 import javax.lang.model.element.ElementKind
 import javax.lang.model.element.TypeElement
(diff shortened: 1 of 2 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -322,10 +322,11 @@ private fun String.escapeDollarSigns(): String {
 private fun TypeName.unwrapTypeAlias(): TypeName {
   return when (this) {
     is ClassName -> {
-      tag<TypeNameAliasTag>()?.type ?: this
+      tag<TypeNameAliasTag>()?.type?.copy(nullable = isNullable, annotations = annotations) ?: this
```
cketti: I'm not sure this does the right thing for e.g. 

```kotlin
typealias Foo = Bar?
data class Bar(val whatever: Int)

data class Test(val foo: Foo)
```

`isNullable` for `Foo` is `false`. When resolving `Foo` to `Bar?` the value is `true`. The `copy()` then overwrites the correctly resolved value with `false`. (I may be wrong. I haven't tested this.)

ZacSweers: Oh that's a good catch. Tried a few cases

```kotlin
data class Bar(val whatever: Int)
typealias Foo = Bar?

// foo is a nullable Bar
data class Test(val foo: Foo?) // <- will lint `?` as redundant
```

```kotlin
data class Bar(val whatever: Int)
typealias Foo = Bar

// foo is a nullable Bar
data class Test(val foo: Foo?)
```

```kotlin
data class Bar(val whatever: Int)
typealias Foo = Bar?

// foo is a nullable Bar
data class Test(val foo: Foo)
```

So all three of the above cases, my takeaway is that we should mark this as nullable if either `foo's` declared or underlying types are nullable. Does that sound right?

ZacSweers: Implemented in 6d131d1 and test in b3fffe2

PR #993, opened 2019-10-31
Check against non-nullable known primitives when rendering class blocks

### LINKED ISSUE #991: Incorrect reading of the type in codegen
Hi
When I use the new codegen version 1.9.1, I get the NoSuchMethodException error. Occurs only in classes that have boxed primitives nullable values for example `Int?`. I discovered that this has to do with the new way of creating objects. When we add, for example, `Int?` this generated adapter instead of creating a constructor for "java.lang.Integer" still uses the primitive value of `int` causing the NoSuchMethodException error.

ZacSweers:
`Int?` will not equal `Int` in KotlinPoet, so this was always falling through to the default `::class.java` code.

Resolves #991

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
@@ -71,7 +71,9 @@ internal fun TypeName.asTypeBlock(): CodeBlock {
       return bound.asTypeBlock()
     }
     is ClassName -> {
-      return when (this) {
+      // Check against the non-nullable version for equality, but we'll keep the nullability in
+      // consideration when creating the CodeBlock if needed.
+      return when (copy(nullable = false)) {
         BOOLEAN, CHAR, BYTE, SHORT, INT, FLOAT, LONG, DOUBLE -> {
           if (isNullable) {
             // Remove nullable but keep the java object type
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
@@ -435,6 +435,35 @@ class DualKotlinTest(useReflection: Boolean) {
       val wildcardOut: GenericClass<out TypeAlias>,
       val complex: GenericClass<GenericTypeAlias>
   )
+
+  // Regression test for https://github.com/square/moshi/issues/991
+  @Test fun nullablePrimitiveProperties() {
+    val adapter = moshi.adapter<NullablePrimitives>()
+
+    @Language("JSON")
+    val testOutputJson = """{"oh_boy":"testing","that_is_a_test":3,"test_again":0}"""
+
+    assertThat(adapter.serializeNulls().toJson(NullablePrimitives("testing", 3)))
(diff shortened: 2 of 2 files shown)


kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
```diff
@@ -436,6 +436,35 @@ class DualKotlinTest(useReflection: Boolean) {
...
+
+    @Language("JSON")
+    val testInputJson = """{"oh_boy":"testing","that_is_a_test":3}"""
+    val result = adapter.fromJson(testInputJson)!!
+    assertThat(result.ohboy).isEqualTo("testing")
+    assertThat(result.thatIsATest).isEqualTo(3)
+    assertThat(result.testAgain).isEqualTo(0)
+  }
+
+  // public Testy(String var1, Integer var2, int var3, int var4, DefaultConstructorMarker var5)
```
cketti: The comment refers to `Testy`, the original class name in the bug report. The test itself uses `NullablePrimitives`.

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
```diff
@@ -436,6 +436,35 @@ class DualKotlinTest(useReflection: Boolean) {
...
+  @JsonClass(generateAdapter = true)
+  data class NullablePrimitives(
```
cketti: I think this test would be easier to understand if the class contained only a single `Int?` property. Or maybe one nullable property for each primitive type.

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
ZacSweers: Ah this was accidentally left over, will remove

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
ZacSweers: Good call, I'll switch this out to be the different primitive types

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
ZacSweers: 7b4e825

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
ZacSweers: 1e22cb2

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
```diff
@@ -71,7 +71,9 @@ internal fun TypeName.asTypeBlock(): CodeBlock {
       return bound.asTypeBlock()
     }
     is ClassName -> {
-      return when (this) {
+      // Check against the non-nullable version for equality, but we'll keep the nullability in
```
swankjesse: Neat!

PR #994, opened 2019-10-31
Don't allow toJson() serialization on polymorphic default values

### LINKED ISSUE #989: Polymorphic adapter will try to serialize the default instance
Should we error here if we have a non-null default instance that comes through?

ZacSweers:
This type explicitly indicates an unknown value. This is an explosive case, but an alternative could be to just write null into the writer?

Resolves #989

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -275,6 +275,9 @@ public final class PolymorphicJsonAdapterFactory<T> implements JsonAdapter.Facto
     }
 
     @Override public void toJson(JsonWriter writer, Object value) throws IOException {
+      if (defaultValueSet && value == defaultValue) {
+        throw new IllegalArgumentException("Cannot serialize default value instance: " + value);
+      }
       Class<?> type = value.getClass();
       int labelIndex = subtypes.indexOf(type);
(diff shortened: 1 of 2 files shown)


rharter:
This error surprises me. I would have expected the default value write to fail when looking for the subtype and tell the user to register it [here](https://github.com/square/moshi/blob/6323a0b7c82b84c2e9128341533a1399c072b040/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java#L282). Is that not the case?

ZacSweers:
Nope! Default values are special in that they're just a single instance, and don't have an associated type label. It's only for when a type label that _is_ present is unrecognized during deserialization, and not intended for serialization.

rharter:
Yeah, but at the linked line it looks like if an unknown subtype (unlabelled type) is encountered it throws an IAE saying that the user needs to register a type for that subtype. 

I think the actual issue in the test code is that the default value **is** a registered type (an `Error` type). I think this is reasonable for that use case.

ZacSweers:
Ahh I see what you mean now. I suppose that actually covers this case then. Thoughts on keeping this check anyway just to make the error message more clear for default value instances? Otherwise the "register a subtype for this type" might be misleading

rharter:
Agreed!

swankjesse:
I don’t agree with the policy here. Why can’t the fallback value also occur organically?

Also this is particularly problematic if the client reads a value from the server, modifies it, and sends it back.

PR #999, opened 2019-11-02
log exceptions if throw during codegen

Wasabi375:
This just logs any exception that is thrown during codgen if it is not caught otherwise and rethrows it. That way the message of the exception is visible without running gradle with the `--stacktrace` option. 
I kinda whish this were there. While debugging as well as when finding new bugs this would be nice, because I won't need to rerun the build (which takes some time in my project) just to see the error message.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessor.kt
@@ -95,35 +95,40 @@ class JsonClassCodegenProcessor : AbstractProcessor() {
   }
 
   override fun process(annotations: Set<TypeElement>, roundEnv: RoundEnvironment): Boolean {
-    for (type in roundEnv.getElementsAnnotatedWith(annotation)) {
-      if (type !is TypeElement) {
-        messager.printMessage(
-            Diagnostic.Kind.ERROR, "@JsonClass can't be applied to $type: must be a Kotlin class",
-            type)
-        continue
-      }
-      val jsonClass = type.getAnnotation(annotation)
-      if (jsonClass.generateAdapter && jsonClass.generator.isEmpty()) {
-        val generator = adapterGenerator(type) ?: continue
-        generator
-            .generateFile {
-              it.toBuilder()
-                  .apply {
-                    generatedType?.asClassName()?.let { generatedClassName ->
-                      addAnnotation(
-                          AnnotationSpec.builder(generatedClassName)
-                              .addMember("value = [%S]",
-                                  JsonClassCodegenProcessor::class.java.canonicalName)
-                              .addMember("comments = %S", "https://github.com/square/moshi")
-                              .build()
-                      )
-                    }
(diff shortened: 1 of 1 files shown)


ZacSweers:
Logging errors to messager does fail the build, I'm not sure what this is accomplishing that that doesn't? You have to invoke the same kapt process either way, messager will just collect all the errors and share them together at the end (and fail that build step).

Wasabi375:
The default gradle settings don't show the exception message if a build fails with an exception. Instead it logs:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':engine:kaptKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution
   > java.lang.reflect.InvocationTargetException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```
This changes just logs the exception message so it will be displayed by default. To see the full stacktrace gradle would still require the '--stacktrace' option.
It's just a convenience (for me). Right now every time I found a bug, first thing I had to do is rerun the build, just to see the exception message, which takes an additional 30secs up to a minute for me.

ZacSweers:
> The default gradle settings don't show the exception message if a build fails with an exception. Instead it logs:

The exceptions are usually printed before this message. Scroll up in your build trace?

Wasabi375:
Normal log messages are printed before this message. This is fine for invalid user code (using moshi in the wrong way). But in the case of a bug in moshi leading to an exception you get something like this
```
>gradlew :engine:clean :engine:build

> Task :engine:kaptGenerateStubsKotlin
w: ATTENTION!
This build uses unsafe internal compiler arguments:

-XXLanguage:+InlineClasses

This mode is not recommended for production use,
as no stability/compatibility guarantees are given on
compiler or generated code. Use it at your own risk!


> Task :engine:kaptKotlin FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':engine:kaptKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution
   > java.lang.reflect.InvocationTargetException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/5.1.1/userguide/command_line_interface.html#sec:command_line_warnings
```
This change would just add a log entry with the exception message, in this case
```
java.lang.IllegalArgumentException: abstract function withNewPath cannot have code
```

Wasabi375:
Maybe it's not the best idea though, as it might comunicate that it's a user error instead of a bug in moshi.

ZacSweers:
That seems like a bug with kapt then, it should print messager logs and include all the relevant element information.

Here's an example in my own project: https://github.com/zacsweers/catchup

image

If you can narrow down what's specific in your project, I think you should file this on the Kotlin issue tracker as a kapt issue

PR #1010, opened 2019-11-08
Fix incorrect type variance being applied on generated adapters 

### LINKED ISSUE #1009: Generated adapters redeclare type variable variance
Carrying over from https://github.com/square/moshi/issues/1008#issuecomment-550513501. These should just be a no-variant typevariable

ZacSweers:
Resolves #1009. Also fixes a discovered boxed types bug along the way and refactors the type mapping to a new `mapTypes` API for reuse.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
@@ -138,3 +138,9 @@ internal fun <T: TypeName> TypeName.mapTypes(target: KClass<T>, transform: T.() 
     else -> throw UnsupportedOperationException("Type '${javaClass.simpleName}' is illegal. Only classes, parameterized types, wildcard types, or type variables are allowed.")
   }
 }
+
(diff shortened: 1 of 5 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
```diff
@@ -95,3 +99,48 @@ internal fun KModifier.checkIsVisibility() {
...
+        else -> throw UnsupportedOperationException("Not possible.")
+      }
+    }
+    else -> throw UnsupportedOperationException("Type '${javaClass.simpleName}' is illegal. Only classes, parameterized types, wildcard types, or type variables are allowed.")
+  }
+}
+
+internal fun TypeName.stripTypeVarVariance(): TypeName {
+  return mapTypes<TypeVariableName> {
+    TypeVariableName(name = name, bounds = bounds.map { it.mapTypes(TypeVariableName::stripTypeVarVariance) }, variance = null)
```
swankjesse: style nit: we wrap at 100 columns in Moshi style

PR #1014, opened 2019-11-11
Update to Kotlin compile testing 1.2.4

ZacSweers:
This allows us to un-toe-hold a test from before


kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessorTest.kt
```diff
@@ -366,11 +365,84 @@ class JsonClassCodegenProcessorTest {
           """
     ))
     assertThat(result.exitCode).isEqualTo(KotlinCompilation.ExitCode.OK)
-    val adapterSource = result.generatedFiles.find { it.name == "PersonJsonAdapter.kt" }!!
 
+    // We're checking here that we only generate one `stringAdapter` that's used for both the
+    // regular string properties as well as the the aliased ones.
+    val adapterClass = result.classLoader.loadClass("PersonJsonAdapter").kotlin
```
ZacSweers: So there's two approaches demoed here. One is to load the class directly and check that we only have one adapter property. The other is a classic "verify the generated source" approach. The latter is more explicit, the former is simpler. What do you think?

swankjesse: I don’t like this style of test because it asserts a specific implementation instead of a specific behavior. Simpler would be to classload and assert the types of the fields?

```
assertThat(declaredFieldsOf(PersonJsonAdapter::class).map { it.type }).containsExactly(
    JsonReader::Options::class,
    parameterizedType(JsonAdapter::class.java, String::class.java)
)
```

ZacSweers: That's basically what the first example is. Do you want me to make the API look more like that example?

ZacSweers: Done in 856744a

Issue #1016, opened 2019-11-12
Kotlin codegen throws on nullable List and Map


StefMa:
I have basically the same issue as #526.
My model looks like this:
```kotlin
data class Entity(
    @Json(name = "class") val classes: List<String>? = null,
    val properties: Map<String, Any>? = null,
}
```

But the generated (adapter) code looks like this:
```kotlin
private val listOfStringAdapter: JsonAdapter<List<String>> =
      moshi.adapter(Types.newParameterizedType(List::class.java, String::class.java), emptySet(),
      "classes")

private val mapOfStringAnyAdapter: JsonAdapter<Map<String, Any>> =
      moshi.adapter(Types.newParameterizedType(Map::class.java, String::class.java,
      Any::class.java), emptySet(), "properties")
```

which leads tot he fact that the `fromJson` has this code:
```kotlin
classes = listOfStringAdapter.fromJson(reader) ?: throw Util.unexpectedNull("classes", "class", reader)
properties = mapOfStringAnyAdapter.fromJson(reader) ?: throw Util.unexpectedNull("properties", "properties", reader)
```
which will throw if `properties` or `classes` are `null` in the json.

I use version 1.9.1

ZacSweers:
Let's not duplicate the bugs, it's fixed in master and will be in the next release

StefMa:
Hmm - as I can see it #526 is already released?
It is a issue from May 18. Since then there were couple of releases. (See the [Changelog](https://github.com/square/moshi/blob/master/CHANGELOG.md)).

I guess this is a regression...

StefMa:
Sry for the trouble.
I found out that my implementation was wrong 😓 .
This issue is not valid anymore.

PR #1017, opened 2019-11-12
Try to reproduce moshi bug

StefMa:
See https://github.com/square/moshi/issues/1016


StefMa:
Strange. This seems to working 😕

PR #1024, opened 2019-11-15
Update to Kotlin 1.3.60

ZacSweers:


ZacSweers:
Need to wait for https://github.com/tschuchortdev/kotlin-compile-testing/pull/31 to be released first since kotlin-compile-testing uses some private APIs that broke here

ZacSweers:
Folded into #1014

PR #1026, opened 2019-11-17
Handle enclosing visibility + cache metadata parsing

### LINKED ISSUE #1022: Generated JsonAdapters for nested data classes ignore parent class visibility modifier
Hi!

I am using `moshi-kotlin-codegen:1.8.0` to generate adapters for the following classes:
```kotlin
internal data class Parent(    
  val type: Int,
  val name: String? = null) {

  data class NestedClass(
    val nestedProperty: String
  )
}
```

Expected:
moshi generates two JsonAdapters, both with `internal` visibility modifier, because visibility of nested classes cannot exceed visibility of the enclosing class.

Actual:
moshi generates ParentJsonAdapter as `internal`, but NestedClassJsonAdapter as `public`.

ZacSweers:
This resolves #1022 and also became a good opportunity to refactor metadata parsing a bit. This allows us to put a cache in front of it now, which should give us a nice little performance win.

Also added an extra check + test for private classes

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -206,13 +206,26 @@ internal fun targetType(messager: Messager,
       properties.putIfAbsent(name, property)
     }
   }
+  val visibility = kotlinApi.modifiers.visibility()
+  // If any class in the enclosing cl
+  val resolvedVisibility = if (visibility == KModifier.INTERNAL) {
+    // If our nested type is already internal, no need
+    visibility
+  } else {
+    // Implicitly public.
+    val forceInternal = generateSequence<Element>(element) { it.enclosingElement }
+        .filterIsInstance<TypeElement>()
+        .map { cachedClassInspector.toImmutableKmClass(it.metadata) }
+        .any { it.isInternal }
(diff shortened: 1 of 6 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -176,46 +178,62 @@ internal fun targetType(messager: Messager,
...
+          cachedClassInspector.toTypeSpec(supertype.element)
+        }
+      }
+  for (supertypeApi in superTypes.values) {
+    val supertypeProperties = declaredProperties(constructor, supertypeApi)
     for ((name, property) in supertypeProperties) {
       properties.putIfAbsent(name, property)
     }
   }
+  val visibility = kotlinApi.modifiers.visibility()
```
swankjesse: standalone method for this?

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessorTest.kt
```diff
@@ -182,6 +182,20 @@ class JsonClassCodegenProcessorTest {
         "error: @JsonClass can't be applied to LocalClass: must not be local")
   }
 
+  @Test fun privateClassesNotSupported() {
```
swankjesse: nice

PR #1034, opened 2019-11-21
Exclude unused transitive dependency

cesarsotovalero:
Hello, I noticed that the transitive dependency `kotlin-stdlib-common` is in the dependency tree of modules  `moshi-kotlin`, `moshi-kotlin-codegen`, and `moshi-kotlin-tests` but it is not used. Therefore, excluding this transitive dependency from `kotlin-stdlib` will make these modules smaller and Maven dependency trees clearer.

### DIFF
--- a/kotlin/codegen/pom.xml
+++ b/kotlin/codegen/pom.xml
@@ -21,6 +21,12 @@
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
       <artifactId>kotlin-stdlib</artifactId>
+      <exclusions>
+        <exclusion>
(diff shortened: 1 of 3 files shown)


JakeWharton:
I don't see much value in trying to play whack-a-mole with transitive dependencies. For one, this prevents the artifact from participating in proper dependency resolution by consumers which can produce version mismatches in the case that the dependency is needed (such as in multiplatform projects).

swankjesse:
Agreed. If you want to shrink, shrink with ProGuard.

Issue #634, opened 2018-08-24
Not respecting nullability when serializing using generated adapter


konmik:
@JsonClass(generateAdapter = true)
    class Child()

    class Parent(val child: Child?)

    moshi.adapter(Parent::class.java).toJson(Parent(null))

Caused by: java.lang.NullPointerException: value was null! Wrap in .nullSafe() to write nullable values.
        at com.booking.pulse.features.donotdisturb.ChildJsonAdapter.toJson(ChildJsonAdapter.kt:42)
        at com.booking.pulse.features.donotdisturb.ChildJsonAdapter.toJson(ChildJsonAdapter.kt:13)
        at com.squareup.moshi.ClassJsonAdapter$FieldBinding.write(ClassJsonAdapter.java:203)
        at com.squareup.moshi.ClassJsonAdapter.toJson(ClassJsonAdapter.java:172)
        at com.squareup.moshi.JsonAdapter$2.toJson(JsonAdapter.java:144)
        at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:52)
        at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:58)

NightlyNexus:
The adapter for Parent don't know about its property's nullability unless you install the reflective Kotlin adapter or generate a Kotlin adapter for Parent.

konmik:
I understand this.

The issue is that parent's  runtime adapter makes assumption on child's adapter nullability. We should probably add an annotation to solve the issue, add default behavior or something like that.

konmik:
Solution 1: `@JsonClass(generateAdapter = true, nullable = true)`

Solution 2: `@field:Json(name = "time_until", nullable = true)`

Solution 3: when generic serializer finds a generated adapter for a field it marks field as nullable by default.

remcomokveld:
I am also running into this when trying to parse a `List<Parent?>`

I think creating all generatedAdapters with `nullSafe` would be okay for this right? if the child may not be null kotin will throw the NPE when the parent class' constructor is being called.

There is one problem with this approach, and that is that `List<Parent>` is no longer guaranteed to only contain non nulls, but I would think that that needs to be handled in special CollectionJsonAdapter for kotlin collections.

remcomokveld:
The problem I am running into now is that I have a list which can contain null items.

With Kotlin's nullability features this should be possible to express by having 
```
@JsonClass(generateAdapter = true)
data class Item(val itemId: String)
```

[...]

PR #1045, opened 2019-12-01
Bug nullable collection

dkonurov:
I wanna help. I tried first support @Nullable and @NonNull annotation in package jetbrains with target TYPE_USE or TYPE_PARAMETER but it's work above java 8. After added annotation @NullableValues and @NonNullValues for collection and support first in java. And After this annotation in koltin-codegen and KotlinJsonAdapter in class parameter only. #634  try to fix this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,89 @@
 Change Log
 ==========
 
+## Version 1.9.2
+
+_2019-11-17_
+
+ * Fix: Generate correct adapters for several special cases including reified inline types, public
+   classes enclosed in internal classes, deprecated types with `-Werror`, primitives in type
+   parameters, nullables in type parameters, and type aliases in type parameters.
+
+
+## Version 1.9.1
+
+_2019-10-30_
+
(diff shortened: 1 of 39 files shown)


swankjesse:
I'm sorry we didn't signal earlier on the issue report, but I think doing any annotations on the enclosing type is a non-starter because it doesn't compose. If we're going to treat `T` and `T?` as distinct types, it needs to be a part of the `Type` class. The challenge here is not to special-case collections but to find a mechanism that works elsewhere as well.

```
class Order<T>(
    val items: List<T>,
    val price: Money
)
```

```
class Project(
  val supplies: Order<Supply?>,
  val snacks: Order<Snack>,
  val dueDate: Date
)
```

dkonurov:
@swankjesse Hello. Sorry for my bad English. Thank's you for answer. But when I started work with this issue, I think work with type. After  I understand that Java doesn't know about nullable or nonnull type in enclosing class. In Java we can only add annotation on class or type of generic. And If we want add this function in java and in kotlin, we have no other way out how to add annotations. And Your Example work with Kotlin class in generic or reflect. But may be I don't understand something.

PR #1050, opened 2019-12-09
Create logic to register base sub types

Tulioh:
I had a problem when trying to deserialize an interface that is a sub-type of `baseType` and is used as an interface in another `subType` of `baseType`.

This is the exception that is thrown:
`java.lang.IllegalArgumentException: No JsonAdapter for interface com.squareup.moshi.adapters.PolymorphicJsonAdapterFactoryTest$BlockingMessage (with no annotations) for interface com.squareup.moshi.adapters.PolymorphicJsonAdapterFactoryTest$BlockingMessage blockingMessage for class com.squareup.moshi.adapters.PolymorphicJsonAdapterFactoryTest$MessageWithBlockingMessage for interface com.squareup.moshi.adapters.PolymorphicJsonAdapterFactoryTest$Message`

Here is an example of the problem for better understanding:

```java
// My BaseType
interface Message {
}

// My BaseSubType
interface BlockingMessage extends Message {
}

// BaseSubType implementation
class BlockingMessageType implements BlockingMessage {
  final String value;

  BlockingMessageType(String value) {
    this.value = value;
  }
}

// SubType of BaseType that uses BaseSubType
class MessageWithBlockingMessage implements Message {
  final BlockingMessage blockingMessage;

  MessageWithBlockingMessage(BlockingMessage blockingMessage) {
    this.blockingMessage = blockingMessage;
  }
}
```

My implementation solves that problem by adding a extra parameter to the `PolymorphicJsonAdapterFactory` named `baseSubTypes` using the method `withBaseSubType`. With this method I can register extra sub-types (`BlockingMessage`) of class `Message`.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -108,6 +108,7 @@ import javax.annotation.Nullable;
  */
 public final class PolymorphicJsonAdapterFactory<T> implements JsonAdapter.Factory {
   final Class<T> baseType;
+  final List<Class<? extends T>> baseSubTypes;
   final String labelKey;
   final List<String> labels;
   final List<Type> subtypes;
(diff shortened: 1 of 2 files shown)


ZacSweers:
I don't think I see the value this adds. Why not just use `BlockingMessage` when registering your Polymorphic adapter?

tuliopereirazup:
Do you mean `withSubtype(BlockingMessageType::class, "blockingMessageType")`?
Or create a new Adapter for `BlockingMessage` sub-classes?

Tulioh:
Do you mean `withSubtype(BlockingMessageType::class, "blockingMessageType")`?
Or create new adapter for `BlockingMessage` sub-types?

ZacSweers:
Either/or

PR #1053, opened 2019-12-11
Don't linebreak keys with spaces

### LINKED ISSUE #1052: Kotlin Codegen doesn't always handle spaces gracefully in the JSON field names
Came across this whilst trying to answer a question for someone on Stack Overflow.

Using the following:

```kotlin
data class Metadata(
    @Json(name = "1. Information") val information: String,
    @Json(name = "2. Symbol") val symbol: String,
    @Json(name = "3. Last Refreshed") val lastRefreshed: String,
    @Json(name = "4. Interval") val interval: String,
    @Json(name = "5. Output Size") val size: String,
    @Json(name = "6. Time Zone") val timeZone: String
)
```

The generated adapter starts as so:
```kotlin
@Suppress("DEPRECATION", "unused", "ClassName", "REDUNDANT_PROJECTION")
class MetadataJsonAdapter(
  moshi: Moshi
) : JsonAdapter<Metadata>() {
  private val options: JsonReader.Options = JsonReader.Options.of("1. Information", "2. Symbol", "3.
      Last Refreshed", "4. Interval", "5. Output Size", "6. Time Zone")
  // ....
}
```

Note the new line in `private val options` broken within the string field name - This code obviously does not compile!

ZacSweers:
Resolves #1052

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -126,7 +126,7 @@ internal class AdapterGenerator(
           JsonReader.Options::class.asTypeName(),
           nonTransientProperties
               .map {
-                // We manually put in quotes because we know the jsonName is already escaped
+                // We manually put in quotes because we know the jsonName is already escaped.
                 val whitespaceSafeName = it.jsonName.replace(" ", "·")
                 CodeBlock.of("\"$whitespaceSafeName\"")
               }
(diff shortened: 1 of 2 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -121,10 +121,17 @@ internal class AdapterGenerator(
...
-      .initializer("%T.of(${nonTransientProperties.joinToString(", ") {
-        // We manually put in quotes because we know the jsonName is already escaped
-        CodeBlock.of("\"%L\"", it.jsonName).toString()
-      }})", JsonReader.Options::class.asTypeName())
+      .initializer(
+          "%T.of(%L)",
+          JsonReader.Options::class.asTypeName(),
+          nonTransientProperties
+              .map {
+                // We manually put in quotes because we know the jsonName is already escaped
```
oldergod: missing a period

ZacSweers:
```suggestion
                // We manually put in quotes because we know the jsonName is already escaped.
```

PR #1059, opened 2020-01-08
Improve the way Json.name is fetched for codegen

ansman:
This depends on https://github.com/square/kotlinpoet/pull/865 being released

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -125,11 +125,7 @@ internal class AdapterGenerator(
           "%T.of(%L)",
           JsonReader.Options::class.asTypeName(),
           nonTransientProperties
-              .map {
-                // We manually put in quotes because we know the jsonName is already escaped.
-                val whitespaceSafeName = it.jsonName.replace(" ", "·")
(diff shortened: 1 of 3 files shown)


ansman:
Yeah, I suppose it's slightly less gross now

pom.xml
```diff
@@ -37,7 +37,7 @@
     <okio.version>1.16.0</okio.version>
     <okio2.version>2.1.0</okio2.version>
     <kotlin.version>1.3.60</kotlin.version>
-    <kotlinpoet.version>1.4.4</kotlinpoet.version>
+    <kotlinpoet.version>1.5.0-SNAPSHOT</kotlinpoet.version>
```
ansman: Obviously this shouldn't be merged until a stable version is out

ansman:
I'll investigate why it failed. I'll also add a test for json names that contain `"` which currently doesn't work

pom.xml
```diff
@@ -37,7 +37,7 @@
     <okio.version>1.16.0</okio.version>
     <okio2.version>2.1.0</okio2.version>
     <kotlin.version>1.3.60</kotlin.version>
-    <kotlinpoet.version>1.4.4</kotlinpoet.version>
+    <kotlinpoet.version>1.5.0</kotlinpoet.version>
```
ZacSweers: I would imagine this brings in a lot of other breaking changes in metadata with the file facade support

ZacSweers:
Thanks!

PR #1063, opened 2020-01-11
Check types parameter size in init

### LINKED ISSUE #932: Check expected types size in generic generated json adapters
Needs this fixed in KotlinPoet first: https://github.com/square/kotlinpoet/issues/800

Then we can put an initializer first like this

```kotlin
init {
    require(types.size == 1) {
      "TypeVariable mismatch: Expecting 1 type(s) for generic type variables [T], but received ${types.size} with values $types"
    }
  }
```

And yield a more meaningful error message. Helps clarify cases like #931

ZacSweers:
Resolves #932

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -18,6 +18,7 @@ package com.squareup.moshi.kotlin.codegen.api
 import com.squareup.kotlinpoet.ARRAY
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.CodeBlock
+import com.squareup.kotlinpoet.CodeBlock.Companion
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.INT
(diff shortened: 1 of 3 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -162,6 +163,22 @@ internal class AdapterGenerator(
 
     if (typeVariables.isNotEmpty()) {
       result.addTypeVariables(typeVariables.map { it.stripTypeVarVariance() as TypeVariableName })
+      // require(types.size == 1) {
+      //   "TypeVariable mismatch: Expecting 1 type(s) for generic type variables [T], but received ${types.size} with values $types"
+      // }
+      result.addInitializerBlock(CodeBlock.builder()
+          .beginControlFlow("require(types.size == %L)", typeVariables.size)
+          .addStatement(
+              "buildString·{·append(%S).append(%L).append(%S).append(%S).append(%S).append(%L)·}",
```
Egorand: Won't the generated code look like this?
```kotlin
require(types.size == 1) {
    buildString { append("TypeVariable mismatch: Expecting ").append(1).append(" type for generic type variables [").append("T").append("], but received ").append(typesParamName.size) }
}
```
A bit concerned about this being a very long single line.

ZacSweers: Yeah, but in the printed error message it looks fine.

PR #1064, opened 2020-01-11
Fix dualkotlintest

ZacSweers:
This wasn't actually running reflection before because the reflection adapter will use generated adapters anyway if available. I've added an overload that allows controlling this behavior (default is enabled), as I'm not sure how else to set this up.

This also makes an opportunistic optimization to defer checking for generated classes until just before the ClassJsonAdapter, as it avoids unnecessary reflection checks before other non-generatable-types like collections/maps/arrays.

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -177,7 +177,13 @@ internal class KotlinJsonAdapter<T>(
   }
 }
 
-class KotlinJsonAdapterFactory : JsonAdapter.Factory {
+/**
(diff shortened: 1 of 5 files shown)


rharter:
If this is just for the test, I wonder if there's a way to either make the test more representative by breaking out the functional reflection test into it's own module that doesn't include codegen, or using some internal constant or something that the compiler can optimize away, so we don't add an if check and constructor param confusion, adding complexity to the code for a test case.

ZacSweers:
Yeah it's kind of just for the test. Breaking out tests would be a lot of duplication. Maybe just a simple internal backdoor but there's no package private in Kotlin :/

rharter:
I haven't found a good way to do this in Kotlin.  Wish I had some idea to
help.

ZacSweers:
I've got another idea. Involves a backdoor but it's at least not public API

PR #1067, opened 2020-01-14
Generate proguard rules for code gen on demand

### LINKED ISSUE #1065: Generate proguard rules in code gen directly
This way we can generate exact matching rules for what's needed rather than the current blanketed ones.

- Collect rules on a per-processor basis
- For each class
  ```
  -keepnames class {the target class}
  -if class {the target class}
  -keep class {the generated adapter} {
      <init>({exact constructor});
      private final {adapter fields}
  }
  ```
- For each qualifier
  ```
  -keep interface {the qualifier class}
  ```

ZacSweers:
Resolves #1065 

This removes all the generic proguard configuration for code gen in favor of rules for generated adapters and the models they encode for. This allows us to be precise and only keep exactly what's needed.

High level coverage:
 * Keeping the target class name to Moshi's reflective lookup of the adapter.
 * Keeping the generated adapter class name + public constructor for reflective lookup.
 * Keeping any used JsonQualifier annotations and the properties they are attached to.
 * If the target class has default parameter values, also keeping the associated synthetic constructor as well as the `DefaultConstructorMarker` type Kotlin adds to it.

Each rule is intended to be as specific and targeted as possible to reduce footprint, and each is conditioned on usage of the original target type via Proguard/R8's `if` conditional checking.
 
To keep this processor as an `ISOLATING` incremental processor, we generate one file per target class with a deterministic name with an appropriate originating element.

Examples can be found in the added `Processor should generate comprehensive proguard rules` test in `JsonClassProcessorTest`.


kotlin/codegen/pom.xml
```diff
@@ -47,6 +47,11 @@
       <artifactId>kotlinpoet-classinspector-elements</artifactId>
       <version>${kotlinpoet.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.ow2.asm</groupId>
```
ZacSweers: Much of the asm bits used are derived from prior work done in the unfinished bridge generator: https://github.com/square/moshi/blob/z/generateBytecode/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/BridgeGenerator.kt

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -87,7 +93,14 @@ internal fun primaryConstructor(kotlinApi: TypeSpec, elements: Elements): Target
     )
   }
 
-  return TargetConstructor(parameters, primaryConstructor.modifiers.visibility())
+  val kmConstructorSignature = primaryConstructor.tag<ImmutableKmConstructor>()?.signature?.toString()
+      ?: run {
+        messager.printMessage(ERROR, "No KmConstructor found for primary constructor.",
```
rharter: Is this something that is expected to happen (in cases of misconfiguration, or something)? If so, is there a better error message that users might understand?  I'm not sure this would be decipherable by an end user, or tell them how to fix the problem.

rharter: I see below that they'll also receive "No primary constructor found on $element", which seems like a more understandable message.  Is this message required, then?  If it doesn't provide more specific information about how to solve the issue then I'd probably vote for leaving it off to avoid the issues filed referencing this because someone only looked at the first error message.

ZacSweers: Yeah I'm not sure what to do with these, because they're kind of things we know won't happen but don't want to just drop in a `!!` without something more descriptive. If it can gracefully skip it I've tried to do that, but fall back to a hard error that we'd recognize if they report it if it's something truly not something we can work with, like the `error()` case. Could add a request in a followup to file on Moshi if they see it?

PR #1074, opened 2020-01-15
Add README section for using reflect/code gen combined

### LINKED ISSUE #822: Add docs on how to use Moshi with annotations but moshi reflect for IDE builds
https://github.com/JakeWharton/butterknife/tree/master/butterknife-reflect got me thinking that moshi should be able to have a reflect artifact for IDE builds as well. Since moshi already has a reflect artifact, maybe it would be beneficial to add a blurb on how to use reflect on IDE and annotation processing on everything else.

ZacSweers:
Resolves #822. Definitely open to feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,7 +591,7 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
 both private and protected properties. If you have configured both, generated adapters will be used
 on types that are annotated `@JsonClass(generateAdapter = true)`.
 
-#### Getting the best developer experience out of both reflection and code gen
+#### Getting the best developer experience using both reflection and code gen
 
 If you want to to use code gen but also want to avoid the build time cost of annotation processing, 
 one pattern you can use is to use reflection for dev builds and code gen for production builds.
--- a/README.md
+++ b/README.md
@@ -594,7 +594,7 @@ on types that are annotated `@JsonClass(generateAdapter = true)`.
 #### Getting the best developer experience using both reflection and code gen
 
 If you want to use code gen but also want to avoid the build time cost of annotation processing, 
-one pattern you can use is to use reflection for dev builds and code gen for production builds.
+one approach might be to use reflection for development builds and code gen for production builds.
 
 In an Android project, such a setup could look like so:
 
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ on types that are annotated `@JsonClass(generateAdapter = true)`.
 
 #### Getting the best developer experience using both reflection and code gen
 
-If you want to to use code gen but also want to avoid the build time cost of annotation processing, 
+If you want to use code gen but also want to avoid the build time cost of annotation processing, 
 one pattern you can use is to use reflection for dev builds and code gen for production builds.
 
 In an Android project, such a setup could look like so:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -591,6 +591,40 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
 both private and protected properties. If you have configured both, generated adapters will be used
 on types that are annotated `@JsonClass(generateAdapter = true)`.
 
+#### Getting the best developer experience out of both reflection and code gen
+
+If you want to to use code gen but also want to avoid the build time cost of annotation processing,
```
rharter:
```suggestion
If you want to use code gen but also want to avoid the build time cost of annotation processing, 
```

README.md
```diff
@@ -591,6 +591,40 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
 both private and protected properties. If you have configured both, generated adapters will be used
 on types that are annotated `@JsonClass(generateAdapter = true)`.
 
+#### Getting the best developer experience out of both reflection and code gen
```
rharter:
```suggestion
#### Getting the best developer experience using both reflection and code gen
```

README.md
```diff
@@ -591,6 +591,40 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
...
+
+If you want to to use code gen but also want to avoid the build time cost of annotation processing, 
+one pattern you can use is to use reflection for dev builds and code gen for production builds.
```
rharter: The repetition of `use` reads a bit weird.

```suggestion
one approach might be to use reflection for development builds and code gen for production builds.
```

README.md
```diff
@@ -591,6 +591,40 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
...
+dependencies {
+  debugImplementation "com.squareup.moshi:moshi-kotlin:{version}"
+  kaptRelease "com.squareup.moshi:moshi-kotlin-codegen:{version}"
+}
+```
+
+Then in a debug-only source set, contribute the reflective `KotlinJsonAdapter` to your app's Moshi
+instance creation.
+
+Another similar pattern in Android is to use reflect only for IDE builds, and reflect for command
```
rharter:
```suggestion
Another similar pattern in Android is to use reflect only for IDE builds, and code gen for command
```

rharter: I would add one or two sentences before this alternate approach explaining why one might want to do this:

> While this can speed up development builds by removing the annotation processing step required for code gen, it can cause inconsistencies between development and release builds, which might be an issue if you use your development builds for internal distribution and testing.

That could be considered out of scope for the readme, but I think you could argue that any of this subjective material could be.

README.md
```diff
@@ -591,6 +591,40 @@ for better performance and to avoid the `kotlin-reflect` dependency; prefer refl
...
+line and CI builds. Gradle configuration for this could look like so:
+
+```gradle
+dependencies {
+  if (properties.containsKey('android.injected.invoked.from.ide')) {
```
rharter: Using this approach how would someone contribute the `KotlinJsonAdapter` only for builds like this without using a build variant?  Is the suggestion to contribute it regardless and just let it defer to code gen in CI builds?

README.md
ZacSweers: toutou

PR #1076, opened 2020-01-15
Remove kotlin-reflect, replace with kotlinx-metadata-jvm

### LINKED ISSUE #307: Can't moshi-kotlin use kotlinx.reflect.lite instead of heavyweight kotlin-reflect?
Here's a lightweight version of kotlin-reflect: https://github.com/Kotlin/kotlinx.reflect.lite
My question is: Can it be used for Moshi to bring the library size down? If yes, this issue is a feature request for it.

ZacSweers:
Kotlin-reflect has been a long-standing thorn for `KotlinJsonAdapter` for a number of reasons:

- It has a huge impact on binary size, both in terms of the library and the compiler impact. When removing it from the slack code base recently, it saved us nearly 10k methods just by removing it (even though only 2 were used!)
- It's super slow and heavyweight
- It has a number of tricky proguard considerations

This PR replaces kotlin-reflect entirely with kotlinx-metadata-reflect, the low-level metadata library we use in code gen (via kotlinpoet-metadata). It's also what the `kotlinx.reflect.lite` project is in the process of being [migrated to](https://github.com/Kotlin/kotlinx.reflect.lite/pull/12), and used in a number of other major toolchains now (Dagger, R8, etc).

Effectively, this skips past `kotlinx.reflect.lite` and replaces `kotlin-reflect` with something closer to an "ultralight" setup, set up specific to Moshi's needs. It's a significant change to the adapter, but the core logic remains the same with one notable exception (documented at the end).

This supersedes #307, but effectively resolves #307 in spirit.

### Stats

Quick androidx-benchmark on my serialization benchmarks project shows this is ~21.5% faster for buffer reads as well! https://github.com/ZacSweers/json-serialization-benchmarking/pull/10

```
AndroidBenchmark.moshi_kotlin_codegen_buffer_toJson[minified=true]                   5,052,032  ns
AndroidBenchmark.moshi_kotlin_codegen_string_toJson[minified=true]                   7,072,449  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_toJson[minified=true]       8,096,823  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_toJson[minified=true]                9,461,355  ns
AndroidBenchmark.moshi_kotlin_reflective_string_toJson[minified=true]               11,308,542  ns

AndroidBenchmark.moshi_kotlin_codegen_buffer_fromJson[minified=true]                 7,553,282  ns
AndroidBenchmark.moshi_kotlin_codegen_string_fromJson[minified=true]                 8,746,668  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_fromJson[minified=true]     9,167,501  ns
AndroidBenchmark.moshi_kotlin_codegen_buffer_fromJson[minified=false]               10,039,950  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_fromJson[minified=true]             10,454,168  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_fromJson[minified=false]   11,691,721  ns
AndroidBenchmark.moshi_kotlin_codegen_string_fromJson[minified=false]               11,716,408  ns
AndroidBenchmark.moshi_kotlin_reflective_string_fromJson[minified=true]             11,734,532  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_fromJson[minified=false]            12,930,158  ns
AndroidBenchmark.moshi_kotlin_reflective_string_fromJson[minified=false]            14,747,397  ns
```

### Two open questions
* The behavior change is that moshi-kotlin still allowed extending from and encoding non-kotlin superclasses and their properties. I think this was an oversight when we stopped supporting mixing them in Moshi 1.8, as this doesn't match the behavior of code gen.
* ~~What are the appropriate proguard rules for this?~~

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
@@ -1,0 +1,74 @@
+package com.squareup.moshi.kotlin.reflect
+
+import kotlinx.metadata.jvm.JvmFieldSignature
+import kotlinx.metadata.jvm.JvmMethodSignature
+import java.lang.reflect.Constructor
+import java.lang.reflect.Field
+import java.lang.reflect.Method
+
+internal val Class<*>.descriptor: String
+  get() {
+    return when {
+      isPrimitive -> when (kotlin) {
+        Byte::class -> "B"
+        Char::class -> "C"
+        Double::class -> "D"
+        Float::class -> "F"
+        Int::class -> "I"
+        Long::class -> "J"
+        Short::class -> "S"
+        Boolean::class -> "Z"
+        Void::class -> "V"
+        else -> throw RuntimeException("Unrecognized primitive $this")
+      }
+      isArray -> name.replace('.', '/')
+      else -> "L$name;".replace('.', '/')
+    }
+  }
+
+internal val Method.descriptor: String
+  get() {
+    return buildString {
+      append('(')
+      parameterTypes.joinTo(this, separator = "", transform = { it.descriptor })
+      append(')')
+      append(returnType.descriptor)
+    }
+  }
+
+/**
+ * Returns the JVM signature in the form "$Name$MethodDescriptor", for example: `equals(Ljava/lang/Object;)Z`.
+ *
+ * Useful for comparing with [JvmMethodSignature].
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Method.jvmMethodSignature: String get() = "$name$descriptor"
+
+internal val Constructor<*>.descriptor: String
+  get() {
+    return buildString {
+      append('(')
+      parameterTypes.joinTo(this, separator = "", transform = { it.descriptor })
+      append(')')
+      append('V')
+    }
+  }
+
+/**
+ * Returns the JVM signature in the form "<init>$MethodDescriptor", for example: `"<init>(Ljava/lang/Object;)V")`.
+ *
+ * Useful for comparing with [JvmMethodSignature].
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Constructor<*>.jvmMethodSignature: String get() = "<init>$descriptor"
+
+/**
+ * Returns the JVM signature in the form "$Name:$FieldDescriptor", for example: `"value:Ljava/lang/String;"`.
+ *
+ * Useful for comparing with [JvmFieldSignature].
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Field.jvmFieldSignature: String get() = "$name:${type.descriptor}"
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
@@ -27,14 +27,7 @@ internal val Class<*>.descriptor: String
   }
 
 internal val Method.descriptor: String
-  get() {
-    return buildString {
-      append('(')
-      parameterTypes.joinTo(this, separator = "", transform = { it.descriptor })
-      append(')')
-      append(returnType.descriptor)
-    }
-  }
+  get() = parameterTypes.joinToString(separator = "", prefix = "(", postfix = ")${returnType.descriptor}") { it.descriptor }
 
 /**
  * Returns the JVM signature in the form "$Name$MethodDescriptor", for example: `equals(Ljava/lang/Object;)Z`.
@@ -46,14 +39,7 @@ internal val Method.descriptor: String
 internal val Method.jvmMethodSignature: String get() = "$name$descriptor"
 
 internal val Constructor<*>.descriptor: String
-  get() {
-    return buildString {
-      append('(')
-      parameterTypes.joinTo(this, separator = "", transform = { it.descriptor })
-      append(')')
-      append('V')
-    }
-  }
+  get() = parameterTypes.joinToString(separator = "", prefix = "(", postfix = ")V") { it.descriptor }
 
 /**
  * Returns the JVM signature in the form "<init>$MethodDescriptor", for example: `"<init>(Ljava/lang/Object;)V")`.
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -546,13 +546,24 @@ private class JvmSignatureSearcher(clazz: Class<*>) {
 
   private val methodSignatureCache = mutableMapOf<String, Method>()
   private val fieldSignatureCache = mutableMapOf<String, Field>()
-  private val methodIterator by lazy(NONE) { clazz.allMethods().iterator() }
-  private val fieldIterator by lazy(NONE) { clazz.allFields().iterator() }
+  private val declaredMethodsIterator by lazy(NONE) { clazz.declaredMethods.iterator() }
+  private val declaredFieldsIterator by lazy(NONE) { clazz.declaredFields.iterator() }
+  private val methodIterator by lazy(NONE) { clazz.methods.iterator() }
+  private val fieldIterator by lazy(NONE) { clazz.fields.iterator() }
 
   fun findMethod(signature: JvmMethodSignature): Method? {
     val signatureString = signature.asString()
     val cached = methodSignatureCache[signatureString]
     if (cached != null) return cached
+
+    while (declaredMethodsIterator.hasNext()) {
+      val next = declaredMethodsIterator.next()
+      val nextSignature = next.jvmMethodSignature
+      methodSignatureCache[nextSignature] = next
+      if (nextSignature == signatureString) {
+        return next
+      }
+    }
 
     while (methodIterator.hasNext()) {
       val next = methodIterator.next()
@@ -571,6 +582,15 @@ private class JvmSignatureSearcher(clazz: Class<*>) {
     val cached = fieldSignatureCache[signatureString]
     if (cached != null) return cached
 
+    while (declaredFieldsIterator.hasNext()) {
+      val next = declaredFieldsIterator.next()
+      val nextSignature = next.jvmFieldSignature
+      fieldSignatureCache[nextSignature] = next
+      if (nextSignature == signatureString) {
+        return next
+      }
+    }
+
     while (fieldIterator.hasNext()) {
       val next = fieldIterator.next()
       val nextSignature = next.jvmFieldSignature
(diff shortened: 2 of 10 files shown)


kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -202,55 +227,129 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+        generateSequence(rawType) { it.superclass }
+            .mapNotNull { it.toKmClass(false) }
+            .flatMap { it.properties.asSequence() }
+            .filterNot { Flag.IS_PRIVATE(it.flags) }
+            .filter { Flag.Property.IS_VAR(it.flags) }
+
+    for (property in allPropertiesSequence.distinctBy { it.name }) {
+      val propertyField = property.fieldSignature?.let { signature ->
+        val signatureString = signature.asString()
+        rawType.declaredFields.find { it.jvmFieldSignature == signatureString }
```
ZacSweers: I think I need to do an allFields() style function for this as well to be safe.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -282,6 +379,277 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      throw IllegalStateException("Cannot serialize class with metadata $classMetadata")
+    } else {
+      return null
+    }
+  }
+
+  return classMetadata.toKmClass()
+}
+
+private fun Class<*>.allMethods(): Sequence<Method> {
```
ZacSweers: I wonder if this is worth sticking a lazily-filled cache in front of

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Method.jvmMethodSignature: String get() = "$name$descriptor"
+
+internal val Constructor<*>.descriptor: String
+  get() {
+    return buildString {
+      append('(')
+      parameterTypes.joinTo(this, separator = "", transform = { it.descriptor })
```
udalov: Minor, but you can use one-liners for this
```
internal val Constructor.descriptor: String
    get() = parameterTypes.joinToString(separator = "", prefix = "(", postfix = ")V") { it.descriptor }
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -202,55 +227,129 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
     val bindingsByName = LinkedHashMap<String, KotlinJsonAdapter.Binding<Any, Any?>>()
 
-    for (property in rawTypeKotlin.memberProperties) {
-      val parameter = parametersByName[property.name]
+    // TODO this doesn't cover platform types
+    val allPropertiesSequence = kmClass.properties.asSequence() +
+        generateSequence(rawType) { it.superclass }
+            .mapNotNull { it.toKmClass(false) }
+            .flatMap { it.properties.asSequence() }
+            .filterNot { Flag.IS_PRIVATE(it.flags) }
```
udalov:
```suggestion
            .filterNot { Flag.IS_PRIVATE(it.flags) || Flag.IS_PRIVATE_TO_THIS(it.flags) }
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -282,6 +379,277 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
     val options = JsonReader.Options.of(*nonTransientBindings.map { it.name }.toTypedArray())
-    return KotlinJsonAdapter(constructor, bindings, nonTransientBindings, options).nullSafe()
+    return KotlinJsonAdapter(constructorData, bindings, nonTransientBindings, options).nullSafe()
+  }
+
+  private infix fun KmType?.isEqualTo(other: KmType?): Boolean {
+    return when {
+      this === other -> true
+      this != null && other != null -> {
+        abbreviatedType == other.abbreviatedType
```
udalov: Note that this will imply that types which are represented with the same runtime type are different, if they have a different abbreviation. Not sure if you'd like this, but if you would, then never mind:
```
typealias ListOfString = List<String>
typealias MyList<T> = List<T>
```
KmType of `ListOfString` isEqualTo KmType of `MyList<String>` -> `false`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -282,6 +379,277 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+  }
+
+  private infix fun KmTypeProjection?.isEqualTo(other: KmTypeProjection?): Boolean {
+    return when {
+      this === other -> true
+      this != null && other != null -> {
+        variance == other.variance
+            && type isEqualTo other.type
+      }
+      this == null && other == null -> true
```
udalov: I suppose this is already checked by `this === other -> true`?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -282,6 +379,277 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+}
+
+internal data class ParameterData(
+    val km: KmValueParameter,
+    val index: Int,
+    val rawType: Class<*>,
+    val annotations: List<Annotation>
+) {
+  val name get() = km.name
+  val isOptional get() = Flag.ValueParameter.DECLARES_DEFAULT_VALUE(km.flags)
```
udalov: Note that semantics of "is optional" are bit different from "declares default value". Not sure how important it is for your use case:
```
interface A {
    fun foo(x: String = "...")
}

interface B : A {
    override fun foo(x: String)
}
```
`x` in `B.foo` is optional, although it does _not_ declare default value.

[`KParameter.isOptional`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-parameter/is-optional.html) implements the "is optional" semantics.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
ZacSweers: Good call! 6cff563

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Hmm, that's a good point. Since we're just doing this for checking equality of the parameter and property. I guess since it's a legal assignment we can drop the alias check. Done in 36f945b

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Hmm, good point! I added the this === other check later 🙃 d3e6e36

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Interesting, that's a good point. I think we can stick with just our default check since this is only ever checked on constructor parameters. Updated the naming of this though to reflect the default value semantic though! 164bd08

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: a5a693b

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: da04f76

ZacSweers:
Updated with modernized proguard rules for this approach that also leverage R8's features where appropriate (based off of this PR for kotlin-reflect rules https://github.com/JetBrains/kotlin/pull/2893)

@swankjesse want to get your sign-off before merging, especially with the platform types behavior change.

kotlin/reflect/src/main/resources/META-INF/com.android.tools/proguard/moshi-kotlin.pro
```diff
@@ -0,0 +1,26 @@
...
+
+# Keep names of Kotlin classes but allow shrinking if unused
+-keepnames @kotlin.Metadata class *
+
+# Keep fields, constructors, and methods in Kotlin classes.
+-keepclassmembers @kotlin.Metadata class * {
+  <fields>;
+  <init>(...);
+  <methods>;
+}
```
ZacSweers: @udalov @jsjeon I wonder if these should be added to the kotlin-reflect rules as well

udalov: For some reason I thought that the `-keep class kotlin.Metadata` rule would also keep all members. If that's not true, then your proposal seems reasonable.

ZacSweers: This is slightly different, and rather it applies this rule only to classes annotated with Metadata, not the metadata annotation class itself

ZacSweers: thinking on it more, I think it's probably best to omit these for now and have the guidance be that users apply keep rules as needed for their exact cases. Same with kotlin-reflect - the rules in place now make sure _its_ machinery works, and users can just make sure they handle appropriate rules for classes that they plan to reflect on

ZacSweers: Cleaned up in ad47a56

jsjeon: Oops, I should have paid attention to github alarms more often and carefully. Sorry for the late game.

@udalov 
> I thought that the -keep class kotlin.Metadata rule would also keep all members.

That's not true. The rule doesn't specify members, and thus all members are still eligible for shrinking and minification. Since members, such as `k`, `d1`, etc., are read all together (via metadata library), the correct rule is `-keep class kotlin.Metadata { *; }`.

But, as @ZacSweers already mentioned, rules here point to classes with metadata annotation, not the metadata definition itself.

Regarding having similar rules in kotlin-reflect, I would say no, because it's still up to users to add such rules if some classes (w/ or w/o metadata annotation) have reflective uses. Glad to see that you ended up with the same conclusion.

ZacSweers:
Just a note for posterity - @swankjesse requested this not be merged until he can review as well

ZacSweers:
Will reopen a new PR, I can't push to this branch anymore since I stepped back from the project 😬

ZacSweers:
Continued here #1183

Issue #784, opened 2019-01-07
PolymorphicJsonAdapterFactory withDefaultSubtype and withFallbackSubtype support


DeweyReed:
Issue #756(How to handle new polymorphic label in backward compatible way?) is folded into #739(PolymorphicJsonAdapterFactory could have a fallback class like with enum) which is closed by the PR #778.

The newly added method `withDefaultValue` falls back to default value(like an error message) but can't fall back to a default type(or label). 

Code example: 

now we can do this

```Kotlin
PolymorphicJsonAdapterFactory.of(Weather.class, "type")
        .withSubtype(Sunny.class, "sunny")
        .withSubtype(Rain.class, "rain")
        .withDefaultValue(Sunny(...))
```

we can't do this without breaking backward compatility

```Kotlin
PolymorphicJsonAdapterFactory.of(Weather.class, "type")
        .withSubtype(Sunny.class, "sunny")
        .withSubtype(Rain.class, "rain")
        .withDefaultSubtype(Rain.class)
        // or this
        //.withDefaultLabel("sunny")
```

NightlyNexus:
`withDefaultSubtype` sounds good to me. open to a PR, unless anyone has other concerns.

ZacSweers:
I worry this would be kind of a footguns since it assumes they all have compatible data fields for the default subtype. I've always just seen unknown fallbacks used with basically like a singleton "unknown" instance to signal it was unrecognized, otherwise you could have problems trying to deserialized unknown data and hoping it works

alex-tavella:
Hey guys, do you have any plan on having this? It would be really useful for a scenario kinda similar to the one @NightlyNexus pointed out.  `.withDefaultValue(value)` won't be suffice to us since we want to send analytics events on failing types (e.g. every occurrence of `DefaultUnknownPromoBanner` with its labelkey).

ZacSweers:
For the example given, you can just make a singleton instance of `DefaultUnknownPromoBanner` and provide that. If it's not decoding _anything_, there's not much different between having an instance vs asking moshi to reflectively deserialize nothing. The API overhead of different fallback strategies is a bit gnarly if we just add custom types, so if we add more in this space I think it's far more likely that we'd add some sort of flexible callback like so:

```kotlin
.withUnknownHandler { label: String, reader: JsonReader, moshi: Moshi -> 
 // Do whatever you want here, return a subtype of the base type though 
}
```

alex-tavella:
[...]

PR #1094, opened 2020-02-09
Extend PolymorphicJsonAdapterFactory to allow custom handling of unknown labels

macisamuele:
This PR includes an implementation proposal for #784.

The main idea is to allow users of the `JsonAdapterFactory` to provide a `JsonAdapter` that will deal with the `reader` in case of an unknown/unmapped label.

Doing this has the advantage to give freedom to developers to re-use already existing components and tools to build their logic.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -95,46 +95,52 @@ import javax.annotation.Nullable;
  *   <li>Each type identifier must be unique.
  * </ul>
  *
- * <p>For best performance type information should be the first field in the object. Otherwise Moshi
- * must reprocess the JSON stream once it knows the object's type.
+ * <p>For best performance type information should be the first field in the object. Otherwise
+ * Moshi must reprocess the JSON stream once it knows the object's type.
  *
- * <p>If an unknown subtype is encountered when decoding, this will throw a {@link
- * JsonDataException}. If an unknown type is encountered when encoding, this will throw an {@link
- * IllegalArgumentException}. If the same subtype has multiple labels the first one is used when
- * encoding.
+ * <p>If an unknown subtype is encountered when decoding:
+ *  <ul>
+ *    <li> if {@link #withDefaultValue(Object)} is used, then {@code defaultValue} will be returned
+ *    <li> if {@link #withFallbackJsonAdapter(JsonAdapter)} is used, then the
+ *        {@code fallbackJsonAdapter.fromJson(reader)} result will be returned
+ *    <li> otherwise a {@link JsonDataException} will be thrown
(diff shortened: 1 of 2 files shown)


macisamuele:
Ping @JakeWharton, @ZacSweers as last developers taking care of merging PRs.

As mentioned in the issue I'm working on https://github.com/Yelp/swagger-gradle-codegen/issues/103 and having this merged would simplify a lot the process as well as it would solve #784 by providing flexibility to developers of managing the "edge-case" of unknown labels as they like.

I'd like to have some feedbacks around this PR or at least an ETA 😄 .

ZacSweers:
Sorry for the delay. In the future - tagging individual users based on perceived activity isn't the best approach and would appreciate if you don't do that.

I like the idea behind this approach, but have a few thoughts:
- Should we deprecate `withDefaultValue()`? Or just keep it as a shorthand for that specific case?
- I don't think we should expose an entire adapter in the API, as we _only_ want this for reads. This gets to a larger point though, maybe we should have read and write interfaces (this could play well into some cases we want to support in the future)

macisamuele:
@ZacSweers , sorry for the direct ping. I apologize for that.

> Should we deprecate `withDefaultValue()`?

Ideally yes, as already visible on the PR the method does return a `JsonAdapter`  that essentially does very little other than returning the value.
I would be more than happy to annotate the method as deprecated, I was mostly looking for a feedback from the owners before doing it.


>  I don't think we should expose an entire adapter in the API, as we only want this for reads. This gets to a larger point though, maybe we should have read and write interfaces (this could play well into some cases we want to support in the future)

I understand your point, but consider the following case in https://github.com/Yelp/swagger-gradle-codegen/issues/103
```kotlin
class Animal(open var type: String)
data class Dog(override var type: String, var name: String): Animal(type=type)
data class Cat(override var type: String, var size: Int): Animal(type=type)
```
If we would allow "custom"  handling on unrecognized labels we might have `{"type":"unknown"}` that will be converted (via the usage of the `PolymorphicJsonAdapterFactory`) into `Animal(type="unknown")`. At this point _how can we re-serialize the kotlin-object?_

The above example is mostly meant to provide a realistic use-case where we don't only need read but also writes.

Said so I'm open to ideas to implement it differently, but I thought that giving the freedom to inject a `JsonAdapter` as fallback was the easiest to implement, to understand (from a user point-of-view) and more importantly the one that allows the greatest customization.


Before I do update the PR by annotating `withDefaultValue` as deprecated I'd like to understand if the approach proposed in this PR is a viable approach.

macisamuele:
Are there any updates on this?

swankjesse:
@macisamuele I like the idea, lets go forward with this.

swankjesse:
Small follow-up: https://github.com/square/moshi/pull/1182

swankjesse:
Nice PR by the way. I apologize for being slow-to-respond. The last several months have been... hectic.

PR #1098, opened 2020-02-15
Add JsonWriter.jsonValue API

JakeWharton:
This is symmetric with JsonReader.readJsonValue in that it writes an Object comprised of maps, lists, and some scalars.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonWriter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonWriter.java
@@ -19,6 +19,8 @@ import java.io.Closeable;
 import java.io.Flushable;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import okio.BufferedSink;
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -386,6 +388,57 @@ public final JsonWriter value(BufferedSource source) throws IOException {
...
+   * @return this writer.
+   * @see JsonReader#readJsonValue()
+   */
+  public final JsonWriter jsonValue(@Nullable Object value) throws IOException {
+    if (value instanceof Map<?, ?>) {
+      beginObject();
+      for (Map.Entry<?, ?> entry : ((Map<?, ?>) value).entrySet()) {
+        Object key = entry.getKey();
+        if (!(key instanceof String)) {
+          throw new IllegalArgumentException(
```
swankjesse: what does this do if key is null?

JakeWharton: NPE creating message. Will fix and test later today. Should also test arrays and maps with null values.

Issue #345, opened 2017-08-26
Moshi-kotlin does not work with proguard


grandstaish:
When ProGuard is enabled the call [`rawType.kotlin.primaryConstructor`](https://github.com/square/moshi/blob/e8a2596841f67fb6423d01416dc2e25dfafa91fa/kotlin/src/main/java/com/squareup/moshi/KotlinJsonAdapter.kt#L160) in `KotlinJsonAdapterFactory` returns `null` so it exits early. 

I created a sample repository here which demonstrates the issue: https://github.com/grandstaish/moshi-kotlin-reflection

The only way I can get Moshi to work currently is to uncomment [this line](https://github.com/grandstaish/moshi-kotlin-reflection/blob/da1ddb408e2f7bd6e48a51c70bf9bcac7668be79/app/proguard-rules.pro#L28) in proguard (`-keep class kotlin.** { *; }`). However, keeping literally everything in kotlin is going way overboard! Is there a better solution to this? 

I have tried to figure this out for too long now and I'm going to have to pass this issue onto the experts, sorry! 😞

swankjesse:
Oooh, yes that’s tough. Not sure exactly what we’ll need to do.

JakeWharton:
Keep Kotlin's metadata annotation. That's all that should be needed.

grandstaish:
I tried keeping the Metadata annotation, but it didn't help here unfortunately. My sample currently has the rules listed in the Moshi README.md, including [keeping the methods of the Metadata annotation](https://github.com/grandstaish/moshi-kotlin-reflection/blob/da1ddb408e2f7bd6e48a51c70bf9bcac7668be79/app/proguard-rules.pro#L23).

grandstaish:
Solved with help from @udalov  🎉 🎉 

The required rules for kotlin are:

```
-dontwarn kotlin.reflect.jvm.internal.**
-keep class kotlin.reflect.jvm.internal.ReflectionFactoryImpl { *; }
-keep class kotlin.Metadata { *; }
```

Also it might be worth mentioning that you need to keep the constructors and fields of your model classes too:

```
-keepclassmembers class my.models.package.** {
  <init>(...);
  <fields>;
}
```

Do you want me to create a PR to update the README.md?

ktchernov:
Keeping Kotlin meta data is not enough. And keeping all constructors and fields of model classes is really annoying to do. Unless you move all your model classes into one package.

[...]

PR #1116, opened 2020-03-27
Update moshi-kotlin.pro

Iljo:
In order to get moshi-kotlin working with reflection, while using R8, this line needs to be added. 
Maybe this line seams obvious, but it could save some hours to someone.

Solution is actually based on answer https://github.com/square/moshi/issues/345#issuecomment-325683339

And it may solve issue https://github.com/square/moshi/issues/1095

### DIFF
--- a/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro
+++ b/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro
@@ -1,5 +1,6 @@
 -keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoaderImpl
 
+-keep class kotlin.Metadata { *; }
 -keepclassmembers class kotlin.Metadata {
     public <methods>;
 }


kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro
```diff
@@ -1,5 +1,6 @@
 -keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoaderImpl
 
+-keep class kotlin.Metadata { *; }
 -keepclassmembers class kotlin.Metadata {
```
rharter, line 4: I'm not as versed in proguard rules as @JakeWharton, but I think the added line effectively invalidates this one by keeping everything in the Metadata class, which is more broad than Moshi needs.

JakeWharton: That is right

Iljo: Ups... True. To optimize it, I suggest changing original keep option for `kotlin.Metadata`, to:
```
-keep class kotlin.Metadata {
    public <methods>;
}
```
Because if I don't preserve class, I get exception:
```
java.lang.IllegalArgumentException: Cannot serialize Kotlin type my.package.CensoredClass$CensoredSubclass. 
  Reflective serialization of Kotlin classes without using kotlin-reflect has undefined and unexpected behavior. 
  Please use KotlinJsonAdapter from the moshi-kotlin artifact or use code gen from the moshi-kotlin-codegen artifact.
```

Only way that I can explain the error being fixed by this new suggestion, is by the sentence of the [proguard manual](https://www.guardsquare.com/en/products/proguard/manual/usage#keepoptions) manual for `keepclassmembers`:

> Specifies class members to be preserved, if their classes are preserved as well. 

Which I would transfer to: `keepclassmembers`  has no effect if you don't keep the class. 

I might be missing something, because the more I dig into proguard, the less I'm confident with it. Anyway,
just wanted to share something that worked for me.

ZacSweers:
See my comment here - https://github.com/square/moshi/issues/345#issuecomment-665412421

I think this shouldn't be merged vs waiting for the new rules in 1.4. If anything, once 1.4 is out we can drop quite a bit from the embedded rules in this project

Iljo:
OK, I'll close the pull request then.

Issue #1119, opened 2020-04-09
Document thread-safety


GrahamLea:
As mentioned in  #468, it would be nice if the docs made it clear which classes are thread safe. Seems like a simple change so I'll make a PR.

JakeWharton:
Thanks for the help!

PR #1120, opened 2020-04-09
Document thread safety

GrahamLea:
As per #1119.

I've just done Moshi, Moshi.Builder and JsonAdapter. Happy to do more on request.

Note that Moshi.Builder wasn't thread-safe, so I trivially made it so by synchronizing its list of factories.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -31,10 +31,10 @@ import okio.BufferedSource;
 /**
  * Converts Java values to JSON, and JSON values to Java.
  *
- * JsonAdapter instances provided by Moshi are thread-safe, meaning multiple threads can safely use a single instance
+ * <p>JsonAdapter instances provided by Moshi are thread-safe, meaning multiple threads can safely use a single instance
  * concurrently.
  *
- * Custom JsonAdapter implementations should be designed to be thread-safe.
+ * <p>Custom JsonAdapter implementations should be designed to be thread-safe.
  */
 public abstract class JsonAdapter<T> {
   @CheckReturnValue public abstract @Nullable T fromJson(JsonReader reader) throws IOException;
--- a/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
@@ -31,8 +31,8 @@ import okio.BufferedSource;
 /**
  * Converts Java values to JSON, and JSON values to Java.
  *
- * <p>JsonAdapter instances provided by Moshi are thread-safe, meaning multiple threads can safely use a single instance
- * concurrently.
+ * <p>JsonAdapter instances provided by Moshi are thread-safe, meaning multiple threads can safely
+ * use a single instance concurrently.
  *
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
```diff
@@ -30,6 +30,11 @@
 
 /**
  * Converts Java values to JSON, and JSON values to Java.
+ *
+ * JsonAdapter instances provided by Moshi are thread-safe, meaning multiple threads can safely use a single instance
```
JakeWharton, line 34: Missing `<p>` tags to separate paragraphs.

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -39,6 +39,8 @@
 
 /**
  * Coordinates binding between JSON values and Java objects.
+ *
+ * Moshi instances are thread-safe, meaning multiple threads can safely use a single instance concurrently.
```
JakeWharton, line 43: Missing `<p>`

moshi/src/main/java/com/squareup/moshi/Moshi.java
```diff
@@ -187,8 +189,13 @@ private Object cacheKey(Type type, Set<? extends Annotation> annotations) {
...
   }
 
+  /**
+   * Builder class for preparing instances of Moshi.
+   *
+   * Builder instances are thread-safe, meaning multiple threads can safely use a single instance concurrently.
+   */
   public static final class Builder {
-    final List<JsonAdapter.Factory> factories = new ArrayList<>();
+    final List<JsonAdapter.Factory> factories = Collections.synchronizedList(new ArrayList<JsonAdapter.Factory>());
```
JakeWharton, line 198: I can't think of a real reason to support concurrent use of a builder.

moshi/src/main/java/com/squareup/moshi/JsonAdapter.java
GrahamLea: Fixed

moshi/src/main/java/com/squareup/moshi/Moshi.java
GrahamLea: Fixed

moshi/src/main/java/com/squareup/moshi/Moshi.java
GrahamLea: Me neither. It just seemed like there's no good reason for it not to be thread-safe when the things around it are deliberately designed to be, and it's trivial to make it so. 🤷🏻‍♂️

GrahamLea:
Lines were too long in the previous build.

moshi/src/main/java/com/squareup/moshi/Moshi.java
JakeWharton: I don't want to commit to thread-safety in the builder because it gives off the wrong impression. Factory order is extremely important so accessing it concurrently means you're opening yourself up to non-deterministic ordering. If you need concurrent access (which will be no one in practice) you can synchronize on the instance you're using.

GrahamLea: Okay

PR #1121, opened 2020-04-09
Improve error when incorrectly trying to use a collection class like ArrayList instead of List

angusholder:
A new user of Moshi might try to make a class like

    data class Response (
        val users: ArrayList<User>
    )

Then when trying to create a JsonAdapter for it, they get an unhelpful error
message like

    Platform class java.util.ArrayList in java.util.ArrayList<com.example.User>
    requires explicit JsonAdapter to be registered

which doesn't explain what they should do to fix the problem. In fact what it
hints towards is that the user should implement a JsonAdapter for ArrayList,
when really they just need to change the type to List!

Now the error looks like

    No JsonAdapter for java.util.ArrayList<com.example.User>, you should use
    List instead of ArrayList (Moshi only supports the collection interfaces
    by default).

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
+++ b/moshi/src/main/java/com/squareup/moshi/ClassJsonAdapter.java
@@ -23,6 +23,8 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Can this go into the failure condition that already exists? As implemented right now we pay for these checks on every adapter creation that gets this far down the list.

angusholder:
I've pushed a change now

JakeWharton:
Thank you!

PR #1123, opened 2020-04-12
Introduce the concept of readOnly/writeOnly generated adapters

Tolriq:
This is a proposal for discussion.

I wanted a quick test to evaluate the gains, and open discussion.
My application talks to many different APIs that have many endpoints with many many properties, but 95% of them are only read and never written.

Applying this to my main app, gains 300KB in the final dex size, and 120KB to the APK size, those are not small numbers, so it would be nice to have something to handle this use case.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -54,7 +54,9 @@ private const val TO_STRING_SIZE_BASE = TO_STRING_PREFIX.length + 1 // 1 is the 
 /** Generates a JSON adapter for a target type. */
 internal class AdapterGenerator(
     private val target: TargetType,
-    private val propertyList: List<PropertyGenerator>
+    private val propertyList: List<PropertyGenerator>,
+    private val readOnly: Boolean,
+    private val writeOnly: Boolean
 ) {
 
   companion object {
@@ -239,15 +241,17 @@ internal class AdapterGenerator(
       }
     }
 
-    result.addProperty(optionsProperty)
+    if (!writeOnly) {
+      result.addProperty(optionsProperty)
+    }
     for (uniqueAdapter in nonTransientProperties.distinctBy { it.delegateKey }) {
       result.addProperty(uniqueAdapter.delegateKey.generateProperty(
           nameAllocator, typeRenderer, moshiParam, uniqueAdapter.name))
     }
 
     result.addFunction(generateToStringFun())
-    result.addFunction(generateFromJsonFun(result))
-    result.addFunction(generateToJsonFun())
+    result.addFunction(generateFromJsonFun(result, writeOnly))
+    result.addFunction(generateToJsonFun(readOnly))
 
     return result.build()
   }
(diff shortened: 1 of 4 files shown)


JakeWharton:
This is exceedingly asymmetric. Presumably the opposite case exists too?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -541,26 +542,28 @@ internal class AdapterGenerator(
...
-    result.endControlFlow()
-
-    result.addStatement("%N.beginObject()", writerParam)
-    nonTransientProperties.forEach { property ->
-      // We manually put in quotes because we know the jsonName is already escaped
-      result.addStatement("%N.name(%S)", writerParam, property.jsonName)
-      result.addStatement("%N.toJson(%N, %N.%N)",
-          nameAllocator[property.delegateKey], writerParam, valueParam, property.name)
+    if (ignoreToJson) {
+      result.addComment("ignoreToJson specified!")
```
JakeWharton, line 551: This should throw

Tolriq: Absolutely (as said in description) but I found ```JsonDataException``` description very precise and not matching this case, and did not know what to throw so went with comment first to see real life gains.
Will update PR with JsonDataException as in all cases the gains are too big for ignoring them and worth maintaining a fork if necessary. (Let me know if you have better idea for what to throw)

Tolriq:
PR updated went with the readOnly/writeOnly names and throw JsonDataException. (Should maybe throw something else or update the description of JsonDataException)

Tests are incomplete as I don't know how to check that the proper throw code is generated :(

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
JakeWharton: I would think `UnsupportedOperationException` is probably most appropriate here

sam33rdhakal:
@JakeWharton is this going to be merged anytime soon ?

swankjesse:
I don’t like it because it fails in awkward ways. In particular, if you generate the wrong kind of adapter it fails too late, at runtime. It’s also potentially annoying to do well in practice, because you might want to strip the write-only path in production, but use it in tests.

Have you considered configuring ProGuard config to prune the unwanted methods by name? That’s very simple! You could potentially even generate `.pro` config files with annotations:

```
  @WriteOnly
  data class Person(...)
```

Tolriq:
If you feed wrong data to the adapter it also fails at runtime.
About tests having them fails is the way to know there's a configuration problem before runtime.
Proguard rules would fails in even worse ways at runtime and since most don't apply proguard / R8 in debug would see the issue even later.

So I kindly disagree but anyway due to the delays, I'm in prod with that since a very long time, so I guess I'll have to keep a fork.

swankjesse:
SGTM. A fork is a nice fix here.

Tolriq:
I'll also disagree as optimisation is a global need and not something specific to me but won't insist :)

Issue #1126, opened 2020-04-13
Add the ability to track data mapping issues


LachlanMcKee:
The `EnumJsonAdapter` provides the ability to specify an unknown fallback value. This is a great feature that can help avoid bugs.

However I feel that there are some scenarios where the developer may want to know when this happens. Perhaps that is an unexpected value that should be handled differently than others.

The `EnumJsonAdapter` does this fallback silently, and does not keep track of this.

My proposal is to add mechanism to the `JsonReader` that can be used to track these types of data mapping issues.

I raised the following PR to demonstrate this: https://github.com/square/moshi/pull/1125

Is this something worthwhile for the project? Another way I could solve this would be to wrap the `JsonReader` in a special reader that does this.

I am interested in also expanding upon this by adding the ability to filter out invalid elements from List/Sets and also log this.

I have another branch that expands upon this PR that can be found here: https://github.com/LachlanMcKee/moshi/commit/a6aa429030ba1325e8d6028669c8d6bd8720eadc (note: this code is not production ready, just wanted to get something working to demonstrate the concept)

rharter:
Thanks for proposing this and sharing an example PR. I agree with you about the problem statement, but I'm not sure the proposed solution is ideal.

Adding custom functionality into the JsonReader for a specific class of object, Enums, violates some of Moshi's compostable nature, and that could easily be handled by the EnumJsonAdapter itself. Additionally, adding things to the reader to handle debug cases isn't ideal. 

Something I'd be more in favor of is adding an option to the EnumJsonAdapter itself to provide this functionality. Having a SAM interface that allowed the user to provide a custom action for unmatched strings would allow the user to do this without adding weight to the JsonReader.

```
EnumJsonAdapter.of(MyEnum.class)
  .onUnknownValue((name) -> {
    Log.d(TAG, "Unknown JSON type: " + name);
  })
  ...
```

While that's an option, I'm not sure I'm entirely convinced. what I've done in the past is only add unknown value support in production and let it crash in internal builds, or log unknown instances at the consumption point. Given the debug nature of this is might also make sense to just wrap the EnumJsonAdapter in another super thin one that reports unknown values, but only in debug builds.

LachlanMcKee:
Hi @rharter. Thanks for getting back to me.

I was considering this as a legitimate production feature, not a debugging feature.

I mentioned in my issue further down that this could be extended to also tracking other invalid data such as filtered list elements.

[...]

PR #1131, opened 2020-04-20
[Using delegation] Introduced DataMappingMismatchLog to track mapping errors

LachlanMcKee:
Relates to issue #1126.

This is a similar change to the other PR, however the logic has been moved out of the standard `JsonReader`, and moved into a more specific implementation.

To facilitate this, the `JsonReader` abstract class has now had to become an interface, and to maintain compatibility of the `JsonReader of(BufferedSource source)`  static method, the Java version has been increased to Java 8 (from Java 7).

Retrofit and several other square libraries have already increased their JVM version, so I thought this may be a possible approach.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
@@ -15,8 +15,12 @@
  */
 package com.squareup.moshi.adapters;
 
-import com.squareup.moshi.*;
-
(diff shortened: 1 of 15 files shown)


LachlanMcKee:
@swankjesse I have replied to your comment on the issue, and raised https://github.com/square/moshi/pull/1209 as a possible alternative using listeners instead (the developer can choose whether the listener should contain state or not)

LachlanMcKee:
Closing this PR as the tags approach will be suitable enough for my requirements, and require less major changes in the library.

Issue #1132, opened 2020-04-20
Change JsonReader to be an interface


LachlanMcKee:
Based on another issue that I have recently raised, I have found that it is much easier to extend the functionality of the `JsonReader` through delegation.

Unfortunately it is not currently possible to create a wrapper `JsonReader` that delegate to another `JsonReader` due to internal logic. The delegate would also contain unnecessary data as it would need to instantiate a standard `JsonReader`.

By making the `JsonReader` class and interface, it would become very easy to create a delegating reader.

Note: Unfortunately to maintain compatibility, the JVM would need to updated to 1.8 due to needing a static interface method.

Here is a link to the PR: https://github.com/square/moshi/issues/1132

LachlanMcKee:
Any thoughts on this? Other than the change to JDK 8, it seems fairly harmless (as it shouldn't break API compatibility), and gives the library more extensibility through composition.

I believe a lot of the other square libraries are using JDK 8 as well now?

swankjesse:
It breaks API compatibility because JVM bytecode bakes in whether it’s calling an interface or a class method.

Perhaps a better question – what behavior do you want to implement via delegating? We could potentially support something like interceptors on the read and write path.

LachlanMcKee:
Hey @swankjesse, sorry for the delay getting back to you. I would like to be able to add support for tracking data-mapping issues similar to how I showed in this PR: https://github.com/square/moshi/pull/1131.

I believe by making the reader an interface, it would allow developers to create their own implementations of the the reader where they can retain state (like tracking data mapping issues during parsing) that the core library itself should not need to care about.

By using global interceptors, it becomes difficult to tie a data-mapping issue to a specific parsing that has taken place. Having context of which parse failed allows you to attach extra metrics should you wish to log this to for analysis (i.e. the URL of the request that had the issue).

For example, I have done something similar to a library that I have built to generate Gson TypeAdapters (and would like to migrate to Moshi if possible). I have then made some changes to the retrofit gson converter that allows we to seamlessly log any data-mapping issues along with the URL. Each of these requests also has a unique id in the header that allows us to match the backend logs.

------

Can you think of a way that interceptors could be added, but still retain the ability to do the above? I suppose the interceptors themselves could have state?

LachlanMcKee:
@swankjesse I have created a new PR with a possible solution that does not require delegation (https://github.com/square/moshi/pull/1209).

[...]

PR #1133, opened 2020-04-20
Made JsonReader an interface to allow composition

LachlanMcKee:
See: https://github.com/square/moshi/issues/1132


swankjesse:
Closing to pick up discussion on the linked issue. There might be another way to solve this problem?

PR #1139, opened 2020-05-03
WIP: Add Optional Type

dkonurov:
Add only for reflect kotlin type, and add new Type in default moshi repository. It's MVP only and I add this logic ONLY for reflect kotlin. If my idea is correct I add this logic to codegen. I updated last idea and add general Type of NonNull or Nullable type for alway Parametrized Type in Kotlin and Java. @ZacSweers @swankjesse pls review me

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -286,34 +286,23 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
       if (arrayType == null) {
         resolved
       } else {
-        val optionalType = if (arrayType.type?.isMarkedNullable == false) {
-          Types.ofNonNull(resolved.componentType)
-        } else {
-          Types.ofNullable(resolved.componentType)
-        }
-        Types.arrayOf(optionalType)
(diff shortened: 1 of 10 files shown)


ZacSweers:
This isn’t reviewable in its current state. You should revert all the unrelated formatting changes you made, make sure your new code matches the style of the project, and explain what problem this is trying to accomplish as it’s not clear from the PR description or code. Moshi supports default parameter values as its optional support already.

dkonurov:
Thank you so much for quick answer. I fix code style and I wanna fixed bug: #634

ZacSweers:
There are a few multiple irrelevant formatting changes. But more importantly - that issue is about nullability in collections. This PR seems to conflate optionality with nullability, which are different things.

I also think it's the wrong approach. As Jesse mentioned in the issue, this would need to be done I a way that's controlled by the internal type model. Maybe a solution is something like Gson's `TypeToken` or adopting Kotlin's `KType` in a Kotlin rewrite.

To save yourself time in the future, I would recommend keeping implementation discussions in the issue for now until maintainers give the thumbs up on a specific approach.

dkonurov:
Thank you for answer. Sorry for my bad English. I don't know how and where I can discuss about correct decision that issue. And can you close this PR? or I can push next commit there? I wanna help :-) but I newbie in open source coding. I'm so sorry.

ZacSweers:
Discuss in the issue you linked. You can close this PR yourself

Issue #1115, opened 2020-03-27
Do not reference kotlin.Metadata


mopsalarm:
After updating from moshi 1.8.0 to 1.9.2 `kotlin.Metadata` annotations are suddenly kept by r8, which increases apk size on android unnecessarily. My guess is the class lookup in https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/internal/Util.java#L57

This is really not required if you are only using generated adapters for kotlin classes and don't want to use the `ClassJsonAdapter` at all.

rharter:
Hmm, the only keep rule for `kotlin.Metadata` that we add is in the kotlin-reflect artifact.  Codegen also keeps the synthetic constructors marked with that, but I wouldn't expect that would keep the annotation.  Can you ensure you don't have the reflect artifact?

mopsalarm:
Yes I am not using the reflect artifact. The line of code I highlighted in the issue is not present in the 1.8.0 release, so my guess is that r8 is smart enough to realize that the kotlin.Metadata class is used via r reflection and keeps it because of that.

JakeWharton:
Ask it? `-whyareyoukeeping class kotlin.Metadata`

mopsalarm:
Seems like my guess was correct:
moshi 1.9.2 

```kotlin.Metadata
|- is reflected from:
|  void com.squareup.moshi.internal.Util.<clinit>()
|- is reachable from:
|  com.squareup.moshi.internal.Util
|- is referenced from:
|  com.squareup.moshi.JsonAdapter com.squareup.moshi.Moshi.adapter(java.lang.reflect.Type,java.util.Set,java.lang.String)
|- is invoked from:
|  void com.pr0gramm.app.api.pr0gramm.Api_TagDetails_TagInfoJsonAdapter.<init>(com.squareup.moshi.Moshi)
```

Moshi 1.8.0

```Nothing is keeping kotlin.Metadata```

**Comment by Tolriq** [comment]
Faced the same after finding a workaround for my generics issue.
Adding 
```
-assumevalues public class com.squareup.moshi.internal.Util {
  private static final * METADATA return null;
}
```
To R8 config does workaround this.

Tolriq:
FYI: Asked R8 team https://issuetracker.google.com/issues/152811099

Seems only a workaround is available for now.

Thomas-Vos:
[...]

PR #1152, opened 2020-06-04
Fix R8 keeping all Kotlin metadata when using Moshi.

Tolriq:
R8 now have automatic detection of Class.forName and auto generate keep rules.

This makes application built with Moshi to keep unwanted Kotlin Metadata that have a huge impact on APK size.

After discussion with R8 team, there's currently no solution to avoid this other than using the trick in this PR.

Fixes https://github.com/square/moshi/issues/1115

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/internal/Util.java
+++ b/moshi/src/main/java/com/squareup/moshi/internal/Util.java
@@ -54,7 +54,7 @@ public final class Util {
     Class<? extends Annotation> metadata = null;
     try {
       //noinspection unchecked
-      metadata = (Class<? extends Annotation>) Class.forName("kotlin.Metadata");
+      metadata = (Class<? extends Annotation>) Class.forName(getKotlinMetadataClassName());
     } catch (ClassNotFoundException ignored) {
     }
     METADATA = metadata;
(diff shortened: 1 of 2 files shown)


moshi/pom.xml
```diff
@@ -34,6 +39,14 @@
     </dependency>
   </dependencies>
 
+  <repositories>
```
JakeWharton, line 42: This will prevent us from releasing to Maven central and cannot be used

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -69,6 +70,11 @@
     DEFAULT_CONSTRUCTOR_MARKER = defaultConstructorMarker;
   }
 
+  @Keep /* Kept to avoid optimizing out reflection. */
```
JakeWharton, line 73: You can use our existing rules file to keep this rather than using the dependency.

Thomas-Vos:
Could this pull request be merged, please? This decreases APK size a lot.

Thomas-Vos:
Just tried this pull request in my project and can confirm the issue is fixed, thanks @Tolriq.

Issue #609, opened 2018-08-01
Exclude property without backing field from deserialization


alexbchr:
To remove a property from deserialization with Moshi, I need to use the Kotlin's `@Transient` annotation. However, I use a property without a backing field (with custom getter and setter), and therefore doesn't have any backing field and can't use `@Transient`. 

One solution I thought about is doing a custom JsonAdapter that would skip the deserialization of this property. It is easy to skip serialization using `skipValue()`, but the `fromJson` function needs to return something, which will set the property.

How can I correct that?

Should properties without backing fields be skipped when serializing/deserializing the class?

alexbchr:
I didn't think about doing this at first! It works well! However, I think it is cleaner to define a `val` property along with a custom setter function, like so:
```
val vegetable: String get() = "Potato"
fun setVegetable(vegetable: String) {
    this.vegetable = vegetable
}
```
The 4 issues that I found with using an extension property are:

1. With the extension property, you can't set private or protected class variables or call protected and private functions in the class.
2. No way to declare the extension property in the class. You can declare it in the same file, before or after declaring the class but not in it.
3. Need to add a new import where the property is used.
4. Interoperability with Java is not that great, because you don't see the getter/setter functions directly at the class level, you need to call the static function getting or setting the field. In the project I work on, Java interop is important as not all code was migrated to Kotlin yet.

I still think it would be great to be able to exclude properties without backing field from serialization/deserialization. If not automatically, maybe with a custom annotation.

swankjesse:
Yeah, you're right. Need to find the right API.

thsaravana:
Is there a plan to fix this? We couldn't migrate to Moshi because of this.

tmm1:
I'm running into this as well. Cannot apply `@Transient` to getters without backing fields.

Gson seems to ignore the non-backed kotlin fields without issue. Maybe there's a clue there on how to detect and ignore these?

ZacSweers:
Gson has basically zero understanding of kotlin, anything "working" there is coincidental

goa:
This is a huge issue for our team as well. A class-wide option for the codegen to ignore properties without backing fields would be life-saving for now.

goa:
So, until this enhancement is implemented, here's an ugly hack you could use:

You can just reference a backing field in either the getter or the setter, without actually using it. That way Kotlin will allow you to use `Transient` and Moshi will ignore your computed property. 

[...]

PR #1153, opened 2020-06-04
Codegen: Ignore properties without backing fields

### LINKED ISSUE #1151: Codegen Adapters trying to set properties without backing fields
I have a model that looks something like this:

```kotlin
@Parcelize
data class TimelineEntry internal constructor(
  val path: String,
  internal var durationLong: Long
) : Parcelable {


  var duration: Millis
    get() = Millis(durationLong)
    set(value) {
      durationLong = value.msValue
    }

}
```

where `Millis` is an `inline class`, and we have this weird extra `Long` field to work around a bug in `@Parcelize` where it doesn't handle inline classes correctly.

Even though the `Millis` property is not backed by a field, the moshi codegen tries to create an adapter for the `Millis` class, has an extra step in `fromJson` where it tries to set this property on `result`, and also writes it out in `toJson`.

Since `duration` does not have a backing field, the compiler won't let us mark it with `@Transient` to instruct Moshi to ignore it.

dhleong:
Fixes #1151 

Currently, if a type has a property that is acting as a getter-setter pair (and, thus, does not have a backing field) Moshi still tries to create a JsonAdapter for it, read it from inputs, and write it to outputs. Since Moshi normally operates on fields rather than getters/setters (I think?) this behavior is counter to intuition, and can lead to extra work that is at best unnecessary and at worst crashing, when the property type is not serializable.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -385,9 +385,10 @@ private val AnnotationSpec.isTransient: Boolean
       val mirror = requireNotNull(tag<AnnotationMirror>()) {
         "Could not get the annotation mirror from the annotation spec"
       }
-      return mirror.elementValues.entries.single {
+      val isTransient = mirror.elementValues.entries.singleOrNull {
         it.key.simpleName.contentEquals("isTransient")
-      }.value.value as Boolean
+      } ?: return false
+      return isTransient.value.value as Boolean
     }
 
     return false
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -25,6 +25,7 @@ import com.squareup.kotlinpoet.TypeVariableName
 import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.asTypeName
 import com.squareup.kotlinpoet.metadata.ImmutableKmConstructor
+import com.squareup.kotlinpoet.metadata.ImmutableKmProperty
 import com.squareup.kotlinpoet.metadata.KotlinPoetMetadataPreview
 import com.squareup.kotlinpoet.metadata.isAbstract
 import com.squareup.kotlinpoet.metadata.isClass
@@ -401,6 +402,11 @@ internal fun TargetProperty.generator(
 
   if (!isSettable) {
     return null // This property is not settable. Ignore it.
+  }
+
+  val propertyInfo = propertySpec.tag<ImmutableKmProperty>()
+  if (propertyInfo != null && propertyInfo.fieldSignature == null) {
+    return null // This property does not have a backing field. Ignore it.
   }
 
   // Merge parameter and property annotations
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -404,11 +404,6 @@ internal fun TargetProperty.generator(
     return null // This property is not settable. Ignore it.
   }
 
-  val propertyInfo = propertySpec.tag<ImmutableKmProperty>()
-  if (propertyInfo != null && propertyInfo.fieldSignature == null) {
-    return null // This property does not have a backing field. Ignore it.
-  }
-
   // Merge parameter and property annotations
   val qualifiers = parameter?.qualifiers.orEmpty() + propertySpec.annotations.qualifiers(elements)
   for (jsonQualifier in qualifiers) {
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -366,13 +366,31 @@ private fun declaredProperties(
   return result
 }
 
-private val TargetProperty.isTransient get() = propertySpec.annotations.any { it.className == Transient::class.asClassName() }
+private val TargetProperty.isTransient: Boolean get() = propertySpec.annotations.any { it.isTransient }
 private val TargetProperty.isSettable get() = propertySpec.mutable || parameter != null
 private val TargetProperty.isVisible: Boolean
   get() {
     return visibility == KModifier.INTERNAL
         || visibility == KModifier.PROTECTED
         || visibility == KModifier.PUBLIC
+  }
+
+private val AnnotationSpec.isTransient: Boolean
+  get() {
+    if (className == Transient::class.asClassName()) {
+      return true
+    }
+
+    if (className == JSON) {
+      val mirror = requireNotNull(tag<AnnotationMirror>()) {
+        "Could not get the annotation mirror from the annotation spec"
+      }
+      return mirror.elementValues.entries.single {
+        it.key.simpleName.contentEquals("isTransient")
+      }.value.value as Boolean
+    }
+
+    return false
   }
 
 /**
@@ -447,9 +465,10 @@ private fun List<AnnotationSpec>?.jsonName(): String? {
     val mirror = requireNotNull(annotation.tag<AnnotationMirror>()) {
       "Could not get the annotation mirror from the annotation spec"
     }
-    mirror.elementValues.entries.single {
+    val nameValue = mirror.elementValues.entries.singleOrNull {
       it.key.simpleName.contentEquals("name")
-    }.value.value as String
+    }?.value?.value as? String
+    nameValue?.takeIf { it.isNotEmpty() }
   }
 }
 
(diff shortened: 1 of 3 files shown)


ZacSweers:
This was by design, moshi kotlin just speaks properties and doesn't think about fields vs getter/setters. You should use `@Transient` if a property should be excluded

dhleong:
As mentioned in the original issue, it is not possible to mark a property
without a backing field as Transient

ZacSweers:
I see. I think there needs to be a different solution then, as this regresses anyone intentionally using this behavior today. Maybe something like Json(transient = true)?

swankjesse:
I think it’s unfortunate to use mutable models with Moshi. Values you encode & decode are a lot simpler if they never mutate after you create ’em. 

Is it possible to make this work through the reflection adapter? I’d really like the two to be symmetric so people can do codegen in production and reflection in development.

dhleong:
> I think it’s unfortunate to use mutable models with Moshi. Values you encode & decode are a lot simpler if they never mutate after you create ’em.

In general I agree! When rapidly iterating/prototyping, however, Moshi is an easy crutch to lean on. 

> I see. I think there needs to be a different solution then, as this regresses anyone intentionally using this behavior today. Maybe something like Json(transient = true)?

This would certainly make it straightforward to implement for the reflection adapter. I *think* I remember seeing somewhere that KProperty can indicate whether it has a backing field as well, but that's probably not the correct approach anyway, for the reasons you mentioned.

I wonder if perhaps it would be better to do something like `@JsonClass(generateAdapter = true, constructorOnly = true)` to indicate that properties should be ignored? This feels in line with `data class` semantics, wouldn't break existing clients since it's opt-in, and feels better to me than creating a duplicate of `@Transient`. The downside, of course, is that it doesn't solve the case of wanting to serialize *some* properties but not all....

swankjesse:
Lets do `@Json(transient = true)`.

ZacSweers:
We'd have to do something like isTransient because I forgot it's a hard keyword in java 😅

dhleong:
Finally sat down and hacked something together. The downside of this version is that `name` is now necessarily optional on `@Json`; didn't thoroughly test the wider implications.

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -374,6 +375,25 @@ private val TargetProperty.isVisible: Boolean
...
+      return true
+    }
+
+    if (className == JSON) {
+      val mirror = requireNotNull(tag<AnnotationMirror>()) {
+        "Could not get the annotation mirror from the annotation spec"
+      }
+      val isTransient = mirror.elementValues.entries.singleOrNull {
+        it.key.simpleName.contentEquals("isTransient")
+      } ?: return false
```
ZacSweers, line 390: Let's make a reusable function here rather than duplicate the logic from the `jsonName()` function you pulled this from and consolidate

```kotlin
private operator fun Map<out ExecutableElement, AnnotationValue>.get(member: String): Any? {
  return entries.firstOrNull {
    it.key.simpleName.contentEquals(member)
  }?.value?.value
}

// Then you can do this!
val isTransient = mirror.elementValues["isTransient"] as? Boolean ?: false
```

kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessorTest.kt
```diff
@@ -398,6 +398,37 @@ class JsonClassCodegenProcessorTest {
     )
   }
 
+  @Test
+  fun `Properties with @Json(isTransient = true) should be ignored`() {
```
ZacSweers, line 402: This should just be a functional test, that way we can make sure it actually works too.

Are there any other cases we should check with this?

moshi/src/main/java/com/squareup/moshi/Json.java
```diff
@@ -39,5 +39,6 @@
 @Retention(RUNTIME)
 @Documented
 public @interface Json {
-  String name();
+  String name() default "";
+  boolean isTransient() default false;
```
ZacSweers, line 43: nit: let's add a small doc on each of these explaining them

dhleong:
> I don't know if defaulting to "" for name will work, because I think empty strings are technically valid keys? Quick SO check agrees https://stackoverflow.com/a/58917082

Oof, that's exactly the sort of thing I was afraid would come up from changing the semantics of `@Json(name)`.  Given that the json name can be literally any value, what do you think about instead introducing a new annotation, ex `@JsonTransient` or maybe `@TransientProperty`? 

I do like the simplicity of namespacing things into the one `@Json` annotation, but to enable the simple `@Json(isTransient=true)` usage all of its properties have to be optional, which means things like `@Json var meaninglesslyAnnotated: Boolean` will compile. Using a separate annotation would also mean:

1. We don't have to contrive a sentinel value like `"com.squareup.moshi.Json.name.DEFAULT"` or whatever and hope that nobody decides to return that in their API for some reason
1. We don't have to worry about changing all parts of the code that interact with `@Json.name`

ZacSweers:
Maybe requiring a redundant name is ok since this is sort of a "I'm being super explicit" mode? Either that or a default that isn't a legal key in json and sufficiently identifiable/unique as moshi's

dhleong:
I'm not sure I agree that it's really a "super explicit" mode; my
perspective is that it's sort of just bridging the gap for places where the
built-in `@Transient` isn't allowed to be. The other side of this is that
having `isTransient` in the `@Json` API that is available to Java users
feels odd, since it is meaningless for them. Not a big deal really, but
might be a point in favor of a separate annotation.

The problem is, I believe, there's no such thing as an illegal JSON key. We
could do that `com.squareup...` thing I proposed since it's pretty
thoroughly namespaced to Moshi, but it's definitely legal in JSON.

That said, my vote would probably be for requiring users to be explicit
with the name over using a sentinel value, since it doesn't change the
semantics of the `name` property.

tmm1:
I would still prefer a class-level annotation to ignore non-constructor fields. I use data classes with properties defined in the constructor for json fields, and then add tons of helper properties with custom accessors. None of these have backing fields and I would like to ignore them all during serialization/deserialization (without having to annotate every single one). For example:

```kotlin
data class Pet(
  var name: String,
  var ageInDays: Int,
) {
  val ageInYears get() = ageInDays / 365
  val ageInYearsFriendly get() = "$ageInYears years old"
  val ageInYearsShort get() = "$ageInYears yrs"
  val nameWithAge get() = "$name ($ageInYearsShort)"
  ...
}
```

cc #609

dhleong:
I would be very happy for a class-wide option to ignore backing field-less properties. We could add it to the `@JsonClass` annotation, since that's where we're generating the adapter anyway. Something like `@JsonClass(generateAdapter = true, ignorePropertiesWithoutBackingFields = true)` is perhaps a bit verbose but clear?

ZacSweers:
Superseded by #1417

tmm1:
I had been using the original iteration of this PR (https://github.com/square/moshi/pull/1153/commits/69418a4646c31922790a84860a15a1be53f03f6e), to ignore all computed properties (i.e, without a backing field)

Today, `com.squareup.kotlinpoet.metadata.ImmutableKmProperty` is no longer present. I see `kotlinx.metadata.KmProperty` is likely the replacement, but it doesn't have a `fieldSignature`. Anyone know how I can check for properties without backing fields with modern kotlinpoet?

tmm1:
> I see `kotlinx.metadata.KmProperty` is likely the replacement, but it doesn't have a `fieldSignature`.

I got this working as follows:

```diff
--- a/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
+++ b/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
@@ -46,6 +46,8 @@ import com.squareup.moshi.kotlin.codegen.api.TargetType
 import com.squareup.moshi.kotlin.codegen.api.rawType
 import com.squareup.moshi.kotlin.codegen.api.unwrapTypeAlias
 import kotlinx.metadata.KmConstructor
+import kotlinx.metadata.KmProperty
+import kotlinx.metadata.jvm.fieldSignature
 import kotlinx.metadata.jvm.signature
 import java.lang.annotation.Retention
 import java.lang.annotation.RetentionPolicy
@@ -427,6 +429,11 @@ internal fun TargetProperty.generator(
     return PropertyGenerator(this, DelegateKey(type, emptyList()), true)
   }

+  val propertyInfo = propertySpec.tag<KmProperty>()
+  if (propertyInfo != null && propertyInfo.fieldSignature == null) {
+    return null // This property does not have a backing field. Ignore it.
+  }
+
   if (!isVisible) {
     messager.printMessage(
       ERROR,
```

I'm still interested in making this a class-wide option, as discussed above:

> We could add it to the `@JsonClass` annotation, since that's where we're generating the adapter anyway. Something like `@JsonClass(generateAdapter = true, ignorePropertiesWithoutBackingFields = true)` is perhaps a bit verbose but clear?

but I could use some help understanding how that would be wired up.

ZacSweers:
Any reason Json.ignore doesn't resolve this? A little confused why you're pursuing alternative solutions when this has one

tmm1:
I have thousands of unbacked fields all over my app. Finding which definitions are backed vs unbacked and adding annotations to each of them is cumbersome, and also bloats my classes. I gave an example above:

> ```kotlin
> @JsonClass(generateAdapter = true)
> data class Pet(
>   var name: String,
>   var ageInDays: Int,
> ) {
>   val ageInYears get() = ageInDays / 365
>   val ageInYearsFriendly get() = "$ageInYears years old"
>   val ageInYearsShort get() = "$ageInYears yrs"
>   val nameWithAge get() = "$name ($ageInYearsShort)"
>   ...
> }
> ```

ZacSweers:
Those seem like good candidates for extension functions or [separating your domain models](https://blog.danlew.net/2022/08/15/domain-specific-models/)

PR #1155, opened 2020-06-09
Add Maven wrapper

JakeWharton:
Homebrew forced Maven to depend on OpenJDK 13 but I already have JDKs installed and I don't want another. So, I'm updating all the Maven projects to use Maven wrapper. It's Gradle wrapper, but for Maven.


JakeWharton:
Travis knows about it: https://docs.travis-ci.com/user/languages/java/#projects-using-maven

PR #1159, opened 2020-06-12
Migrate to Gradle

ZacSweers:
We haven't been running checkstyle prior to this, so there's a looooot of errors right now and I've had to disable it. Wanna save that for a followup PR with spotless + an autoformat pass.


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+  `kotlin-dsl`
+}
```
jrodbx, line 19: I think Square projects use (prefer?) the `apply plugin` syntax.
Same with Groovy build script style with kts style.  IIRC, there's still some perf issue related to using buildSrc.

How about mimicking the Gradle setup style of square/retrofit, square/wire, or cashapp/sqldelight?

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
```diff
@@ -114,7 +114,7 @@
       fail();
     } catch (JsonDataException expected) {
       assertThat(expected).hasMessage("Unexpected null at $[1]");
-      assertThat(reader.nextNull()).isNull();
+      assertThat(reader.<Object>nextNull()).isNull();
```
jrodbx, line 117: interesting: why do these tests require updating when moving to Gradle?

ZacSweers: whoops, meant to revert this. Was my early indicator that I needed to set the language version to 7 😅

buildSrc/build.gradle.kts
ZacSweers: That's fair in those projects since they don't use kotlin gradle script and there's no difference to the user. Here they do matter because the modern `plugins {}` syntax will actually make kotlin gradle script generate the type-safe accessors for plugin extensions, whereas the `apply()` syntax won't. We do the same in kotlinpoet for this reason: https://github.com/square/kotlinpoet/blob/master/kotlinpoet/build.gradle.kts.

In general for kotlin gradle dsl you should only use the `apply(plugin = ...)` syntax when conditionally applying it (like the root `build.gradle.kts` file here does).

moshi/src/test/java/com/squareup/moshi/JsonAdapterTest.java
ZacSweers: 62bf821

buildSrc/build.gradle.kts
jrodbx: Yep, but if we use the Groovy build script style instead (my second point), the apply vs plugins syntax becomes moot, I guess.  The referenced projects also use the `apply` syntax.  I think KotlinPoet is the only repo to use Kotlin DSL atm, fwiw.

JakeWharton:
This should very much by a post-1.10.0 thing, or we risk delaying an already-risky 1.10.0.

Thus far my opinion of Kotlin Gradle DSL is the same of kotlinx.html's Kotlin HTML DSL. Like sure I get it, neat, you did a thing. But also like yeesh, really? It's like buying a Tesla. Sure it's a nice electric car saving the environment but it'll occasionally drive you into an overturned truck at 65mph and blame you. And not that Hans or Andrey is an idiot like Elon Musk but maybe this somehow implicit bastard love-child is? Truly the _Ecce Homo_ of build systems.

ZacSweers:
Does this actually impact 1.10 or is the issue just opinions on kotlin gradle dsl? It’s easy enough to for me to change it to groovy if it’s holding up this PR.

JakeWharton:
Nah don't change it.

Unless there's some specific reason to land it before a 1.10, I don't see any need to. It only increases risk. 

When Retrofit was forced to migrate to Gradle for adding Android tests we found two critical problems while it baked in snapshots that would have required a patch release. I would simply rather this soak in snapshots for a bit.

ZacSweers:
Rebased. Now that 1.10's out, good to proceed with this?

build.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+          setUrl("https://jcenter.bintray.com/")
+        }
+      }
+      filter {
+        includeModule("org.jetbrains.dokka", "dokka-fatjar")
+      }
+    }
+  }
+
+//  apply(plugin = "checkstyle")
```
swankjesse, line 47: add a note for this?

swankjesse: (or delete it)

ZacSweers: I vote delete and replace with spotless and a bug formatting pass with GJF and ktlint. What do you think?

swankjesse: Works for me.

PR #1163, opened 2020-06-20
add Dependabot

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


swankjesse:
I don’t anticipate this will be able to handle our custom gradle artifact locations.

PR #1180, opened 2020-07-29
Expose getStrings() option on JsonReader.Options

### LINKED ISSUE #1173: Expose string array in JsonReader.Options
When using `JsonReader.selectName(Options)` it would be useful to be able to access the list of strings stored by this class. For example to give a more descriptive deserialisation error message by listing the valid values. Unfortunately the field `String[] strings` in `JsonReader.Options` is package private and so cannot be accessed.

While there are a bunch of workarounds to not being able to access this field, they all have various drawbacks. For example:
- Use `nextString` instead of `selectName`/`selectString`. This offers the greatest flexibility, but means giving up the performance benefit.
- Hardcode the values in the logic. This requires strings to be duplicated and both need to be maintained.
- Store values in another array/list and pass into `Options.of`. This works equivalently but requires two effectively identical variables which isn't great.
- Use reflection to temporarily get access to the internal variable. Not actually hugely horrible as my use cases are mostly for exception handling which will be the slow path anyway.

So while I'm not quite clear what the reasoning was for making this field package private, I do see a use case for implementing an accessor. Either to directly expose the string array, or as a list instead.

ZacSweers:
Resolves #1173

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -528,6 +528,11 @@ public abstract class JsonReader implements Closeable {
       this.doubleQuoteSuffix = doubleQuoteSuffix;
     }
 
+    /** Returns a copy of this {@link Options Option's} strings. */
+    public List<String> getStrings() {
+      return Arrays.asList(strings);
+    }
+
     @CheckReturnValue public static Options of(String... strings) {
       try {
         ByteString[] result = new ByteString[strings.length];
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -533,7 +533,7 @@ public abstract class JsonReader implements Closeable {
     }
 
     /** Returns a copy of this {@link Options Option's} strings. */
-    public List<String> getStrings() {
+    public List<String> strings() {
       return unmodifiableList(Arrays.asList(strings.clone()));
     }
 
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -19,10 +19,8 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import okio.Buffer;
@@ -534,7 +532,7 @@ public abstract class JsonReader implements Closeable {
 
     /** Returns a copy of this {@link Options Option's} strings. */
     public List<String> strings() {
-      return unmodifiableList(Arrays.asList(strings.clone()));
+      return unmodifiableList(Arrays.asList(strings));
     }
 
     @CheckReturnValue public static Options of(String... strings) {
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -530,7 +530,7 @@ public abstract class JsonReader implements Closeable {
 
     /** Returns a copy of this {@link Options Option's} strings. */
     public List<String> getStrings() {
-      return Arrays.asList(strings);
+      return Arrays.asList(strings.clone());
     }
 
     @CheckReturnValue public static Options of(String... strings) {
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -29,7 +29,7 @@ import okio.Buffer;
 import okio.BufferedSource;
 import okio.ByteString;
 
-import static java.util.Collections.unmodifiableSet;
+import static java.util.Collections.unmodifiableList;
 
 /**
  * Reads a JSON (<a href="http://www.ietf.org/rfc/rfc7159.txt">RFC 7159</a>)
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -528,6 +528,11 @@ private Options(String[] strings, okio.Options doubleQuoteSuffix) {
       this.doubleQuoteSuffix = doubleQuoteSuffix;
     }
 
+    /** Returns a copy of this {@link Options Option's} strings. */
+    public List<String> getStrings() {
+      return Arrays.asList(strings);
```
JakeWharton, line 533: This would allow mutation of items in the backing array.
```suggestion
      return Arrays.asList(strings.clone());
```

ZacSweers: Oh wow, I actually didn't realize it didn't make a copy of the input array. TIL

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -1157,6 +1157,13 @@ private void readPeek12Steps(JsonReader reader, int from, int until) throws IOEx
     reader.endObject();
   }
 
+  @Test public void optionsGetStrings() {
+    String[] options = new String[] { "a", "b", "c" };
+    JsonReader.Options abc = JsonReader.Options.of("a", "b", "c");
+    List<String> strings = abc.getStrings();
+    assertThat(options).containsExactlyElementsOf(strings);
```
ZacSweers, line 1164:
```suggestion
    assertThat(options).containsExactlyElementsOf(strings);
    // Add a value to ensure it doesn't mutate the underlying array if we read it again
    strings.add("d");
    assertThat(abc.getStrings()).containsOnly(options);
```

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -1157,6 +1157,16 @@ private void readPeek12Steps(JsonReader reader, int from, int until) throws IOEx
...
+    // Add a value to ensure it doesn't mutate the underlying array if we read it again
+    strings.add("d");
```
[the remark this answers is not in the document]
JakeWharton: Or switch to `set`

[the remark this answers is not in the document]
JakeWharton: Or switch to wrapping the returned result in `unmodifiableList`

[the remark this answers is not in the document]
ZacSweers: 6fb900d

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -528,6 +532,11 @@ private Options(String[] strings, okio.Options doubleQuoteSuffix) {
       this.doubleQuoteSuffix = doubleQuoteSuffix;
     }
 
+    /** Returns a copy of this {@link Options Option's} strings. */
+    public Set<String> getStrings() {
+      return unmodifiableSet(new LinkedHashSet<>(Arrays.asList(strings.clone())));
```
JakeWharton, line 537: The LinkedHashSet is not needed here. It creates an extra copy. We don't mutate the backing array so there's no need for the copy.

swankjesse: This should be a List, not a Set. Though the elements are unique, the ordering is very important because the rest of the API is focused on indexes.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -528,6 +532,11 @@ private Options(String[] strings, okio.Options doubleQuoteSuffix) {
       this.doubleQuoteSuffix = doubleQuoteSuffix;
     }
 
+    /** Returns a copy of this {@link Options Option's} strings. */
+    public Set<String> getStrings() {
```
swankjesse, line 536: should this be just `strings`? I don’t think we have a `get` prefix in Moshi anywhere else.

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -1157,6 +1158,20 @@ private void readPeek12Steps(JsonReader reader, int from, int until) throws IOEx
...
   }
 
+  @Test public void optionsGetStrings() {
+    String[] options = new String[] { "a", "b", "c" };
+    JsonReader.Options abc = JsonReader.Options.of("a", "b", "c");
+    Set<String> strings = abc.getStrings();
+    assertThat(options).containsExactlyElementsOf(strings);
+    try {
+      // Confirm it's unmodifiable and we can't mutate the original underlying array
+      strings.add("d");
```
swankjesse, line 1168: Note that `Arrays.asList(...).add()` fails even if it isn’t unmodifiable.

moshi/src/main/java/com/squareup/moshi/JsonReader.java
ZacSweers: my head's stuck in "make this nice for kotlin property syntax" land now, can change

ZacSweers: f6bdaec

moshi/src/main/java/com/squareup/moshi/JsonReader.java
ZacSweers: Cleaned up in dfcd74f

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -528,6 +532,11 @@ private Options(String[] strings, okio.Options doubleQuoteSuffix) {
       this.doubleQuoteSuffix = doubleQuoteSuffix;
     }
 
+    /** Returns a copy of this {@link Options Option's} strings. */
+    public List<String> strings() {
+      return unmodifiableList(Arrays.asList(strings.clone()));
```
JakeWharton, line 537: Sorry there's a ton of churn on this line but if you're wrapping in unmodifiable you also don't need the clone 😶

ZacSweers: De-cloned! 31cf8d1

JakeWharton:
**L**ets **G**et **T**his **M**erged

PR #1182, opened 2020-07-31
Simplify the internals of PolymorphicJsonAdapterFactory

swankjesse:
We don't need a custom JsonAdapter subclass - we can just call through
unconditionally and fail if the type is unexpected.

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
@@ -100,30 +100,22 @@ import javax.annotation.Nullable;
  *
  * <p>If an unknown subtype is encountered when decoding:
  *  <ul>
- *    <li> if {@link #withDefaultValue(Object)} is used, then {@code defaultValue} will be returned
- *    <li> if {@link #withFallbackJsonAdapter(JsonAdapter)} is used, then the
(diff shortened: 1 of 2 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java
```diff
@@ -205,19 +197,15 @@
   }
 
   private JsonAdapter<Object> buildFallbackJsonAdapter(final T defaultValue) {
-    return new DefaultJsonAdapter<Object>() {
-      @Nullable
-      @Override
-      public Object fromJson(JsonReader reader) throws IOException {
+    return new JsonAdapter<Object>() {
+      @Override public @Nullable Object fromJson(JsonReader reader) throws IOException {
```
ZacSweers, line 201:
```suggestion
      @Nullable @Override public Object fromJson(JsonReader reader) throws IOException {
```

swankjesse: `@Nullable` is a type annotation and annotates the return type`Object` and not the method. Like `?` in Kotlin!

ZacSweers: I know, it was a style comment. Looking through the rest of the codebase, seems moshi uses a mix of both annotating before and after the modifiers 😅

PR #1183, opened 2020-07-31
Remove kotlin-reflect, replace with kotlinx-metadata-jvm

### LINKED ISSUE #307: Can't moshi-kotlin use kotlinx.reflect.lite instead of heavyweight kotlin-reflect?

ZacSweers:
_Continuation of #1076_ 

Kotlin-reflect has been a long-standing thorn for `KotlinJsonAdapter` for a number of reasons:

- It has a huge impact on binary size, both in terms of the library and the compiler impact. When removing it from the slack code base recently, it saved us nearly 10k methods just by removing it (even though only 2 were used!)
- It's super slow and heavyweight
- It has a number of tricky proguard considerations

This PR replaces kotlin-reflect entirely with kotlinx-metadata-reflect, the low-level metadata library we use in code gen (via kotlinpoet-metadata). It's also what the `kotlinx.reflect.lite` project is in the process of being [migrated to](https://github.com/Kotlin/kotlinx.reflect.lite/pull/12), and used in a number of other major toolchains now (Dagger, R8, etc).

Effectively, this skips past `kotlinx.reflect.lite` and replaces `kotlin-reflect` with something closer to an "ultralight" setup, set up specific to Moshi's needs. It's a significant change to the adapter, but the core logic remains the same with one notable exception (documented at the end).

This supersedes #307, but effectively resolves #307 in spirit.

### Stats

Quick androidx-benchmark on my serialization benchmarks project shows this is ~21.5% faster for buffer reads as well! https://github.com/ZacSweers/json-serialization-benchmarking/pull/10

```
AndroidBenchmark.moshi_kotlin_codegen_buffer_toJson[minified=true]                   5,052,032  ns
AndroidBenchmark.moshi_kotlin_codegen_string_toJson[minified=true]                   7,072,449  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_toJson[minified=true]       8,096,823  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_toJson[minified=true]                9,461,355  ns
AndroidBenchmark.moshi_kotlin_reflective_string_toJson[minified=true]               11,308,542  ns

AndroidBenchmark.moshi_kotlin_codegen_buffer_fromJson[minified=true]                 7,553,282  ns
AndroidBenchmark.moshi_kotlin_codegen_string_fromJson[minified=true]                 8,746,668  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_fromJson[minified=true]     9,167,501  ns
AndroidBenchmark.moshi_kotlin_codegen_buffer_fromJson[minified=false]               10,039,950  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_fromJson[minified=true]             10,454,168  ns
AndroidBenchmark.moshi_kotlin_reflective_metadata_buffer_fromJson[minified=false]   11,691,721  ns
AndroidBenchmark.moshi_kotlin_codegen_string_fromJson[minified=false]               11,716,408  ns
AndroidBenchmark.moshi_kotlin_reflective_string_fromJson[minified=true]             11,734,532  ns
AndroidBenchmark.moshi_kotlin_reflective_buffer_fromJson[minified=false]            12,930,158  ns
AndroidBenchmark.moshi_kotlin_reflective_string_fromJson[minified=false]            14,747,397  ns
```

One open question - The behavior change is that moshi-kotlin still allowed extending from and encoding non-kotlin superclasses and their properties. I think this was an oversight when we stopped supporting mixing them in Moshi 1.8, as this doesn't match the behavior of code gen.


README.md
```diff
@@ -613,7 +613,31 @@ Snapshots of the development version are available in [Sonatype's `snapshots` re
 R8 / ProGuard
 --------
 
-Moshi contains minimally required rules for its own internals to work without need for consumers to embed their own. However if you are using reflective serialization and R8 or ProGuard, you must add keep rules in your proguard configuration file for your reflectively serialized classes.
+If using Android, the Android Gradle Plugin should automatically extract embedded rules included in
```
ZacSweers, line 616: One open bit from the previous PR: https://github.com/square/moshi/pull/1076#commitcomment-37485365

kotlin/reflect/pom.xml
```diff
@@ -101,6 +102,36 @@
...
+              <artifactSet>
+                <includes>
+                  <include>org.jetbrains.kotlinx:kotlinx-metadata-jvm</include>
+                </includes>
+              </artifactSet>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
+              </transformers>
+              <relocations>
+                <!-- Repackage kotlinx-metadata until their APIs are stable. -->
```
swankjesse, line 125: Perfect.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
```diff
@@ -0,0 +1,60 @@
+package com.squareup.moshi.kotlin.reflect
```
swankjesse, line 1: nit: copyright headers

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Returns the JVM signature in the form "$Name$MethodDescriptor", for example: `equals(Ljava/lang/Object;)Z`.
+ *
+ * Useful for comparing with [JvmMethodSignature].
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Method.jvmMethodSignature: String get() = "$name$descriptor"
+
+internal val Constructor<*>.descriptor: String
```
swankjesse, line 41: do we use both `descriptor` and `jvmMethodSignature` forms? Otherwise would be nice to only have one!

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
```diff
@@ -0,0 +1,60 @@
...
+internal val Constructor<*>.jvmMethodSignature: String get() = "<init>$descriptor"
+
+/**
+ * Returns the JVM signature in the form "$Name:$FieldDescriptor", for example: `"value:Ljava/lang/String;"`.
+ *
+ * Useful for comparing with [JvmFieldSignature].
+ *
+ * For reference, see the [JVM specification, section 4.3](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3).
+ */
+internal val Field.jvmFieldSignature: String get() = "$name:${type.descriptor}"
```
swankjesse, line 60: Is this a real thing? I’ve never heard of a field signature before. (method signatures yes!)

kotlin/reflect/src/main/resources/META-INF/com.android.tools/r8-upto-1.6.0/moshi-kotlin.pro
```diff
@@ -0,0 +1,18 @@
...
+-keepnames class kotlin.jvm.internal.DefaultConstructorMarker
+
+# Keep implementations of service loaded interfaces
+# R8 will automatically handle these these in 1.6+
+-keep interface com.squareup.moshi.kotlin.kotlinx.metadata.impl.extensions.MetadataExtensions
+-keep class * implements com.squareup.moshi.kotlin.kotlinx.metadata.impl.extensions.MetadataExtensions { public protected *; }
+
+# Keep generic signatures and annotations at runtime.
+# R8 requires InnerClasses and EnclosingMethod if you keepattributes Signature.
+-keepattributes InnerClasses,Signature,RuntimeVisible*Annotations,EnclosingMethod
```
swankjesse, line 18: It’d be pretty rad to run R8 against our code, then run the tests, and see what happens

swankjesse: (something to fantasize about for CI)

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -315,9 +316,7 @@ class KotlinJsonAdapterTest {
       moshi.adapter<RequiredTransientConstructorParameter>()
       fail()
     } catch (expected: IllegalArgumentException) {
-      assertThat(expected).hasMessage("No default value for transient constructor parameter #0 " +
-          "a of fun <init>(kotlin.Int): " +
-          "com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest.RequiredTransientConstructorParameter")
+      assertThat(expected).hasMessage("No default value for transient constructor parameter 'a' on type '${RequiredTransientConstructorParameter::class.qualifiedName}'")
```
swankjesse, line 319: dig

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -418,7 +417,9 @@ class KotlinJsonAdapterTest {
 
   internal class ExtendsPlatformClassWithPrivateField(var a: Int) : SimpleTimeZone(0, "C")
 
-  @Test fun extendsPlatformClassWithProtectedField() {
+  @Ignore("Pending review")
```
swankjesse, line 420: ?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -95,8 +108,8 @@ internal class KotlinJsonAdapter<T>(
 
     // Confirm all parameters are present, optional, or nullable.
     for (i in 0 until constructorSize) {
-      if (values[i] === ABSENT_VALUE && !constructor.parameters[i].isOptional) {
-        if (!constructor.parameters[i].type.isMarkedNullable) {
+      if (values[i] === ABSENT_VALUE && !constructor.parameters[i].declaresDefaultValue) {
+        if (!constructor.parameters[i].km.type!!.isNullable) {
```
swankjesse, line 112: (off topic: `!!.isNullable` is the cutest thing I’ve ever seen)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -133,46 +146,59 @@ internal class KotlinJsonAdapter<T>(
...
-    val name: String,
-    val jsonName: String?,
-    val adapter: JsonAdapter<P>,
-    val property: KProperty1<K, P>,
-    val parameter: KParameter?,
-    val propertyIndex: Int
+      val name: String,
+      val jsonName: String?,
+      val adapter: JsonAdapter<P>,
+      val property: PropertyData,
```
swankjesse, line 155: (I wonder if there’s a more descriptive suffix than Data...)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -133,46 +146,59 @@ internal class KotlinJsonAdapter<T>(
...
+
+    fun get(value: K): Any? {
+      property.jvmGetter?.let { getter ->
+        return getter.invoke(value)
+      }
+      property.jvmField?.let {
+        return it.get(value)
+      }
+
+      return null
```
swankjesse, line 167: is this reachable?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
-    require(!rawTypeKotlin.isInner) {
+    require(!Flag.Class.IS_INNER(kmClass.flags)) {
       "Cannot serialize inner class ${rawType.name}"
     }
-    require(rawTypeKotlin.objectInstance == null) {
+    require(!Flag.Class.IS_OBJECT(kmClass.flags)) {
       "Cannot serialize object declaration ${rawType.name}"
     }
-    require(!rawTypeKotlin.isSealed) {
+    require(!Flag.Class.IS_COMPANION_OBJECT(kmClass.flags)) {
```
swankjesse, line 245: Good!

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+    require(!Flag.IS_SEALED(kmClass.flags)) {
       "Cannot reflectively serialize sealed class ${rawType.name}. Please register an adapter."
     }
 
-    val constructor = rawTypeKotlin.primaryConstructor ?: return null
-    val parametersByName = constructor.parameters.associateBy { it.name }
-    constructor.isAccessible = true
+    val kmConstructor = kmClass.constructors.find { Flag.Constructor.IS_PRIMARY(it.flags) }
+        ?: return null
+    val kmConstructorSignature = kmConstructor.signature?.asString() ?: return null
```
swankjesse, line 254: is the `return null` reachable?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+        }
+        append(Util.DEFAULT_CONSTRUCTOR_MARKER!!.descriptor)
+        append(")V")
+      }
+      constructorsBySignature[defaultConstructorSignature] ?: return null
+    } else {
+      jvmConstructor
+    }
+
+    val constructorData = ConstructorData(
```
swankjesse, line 286: would it be useful to have a helper method to do the work to produce this?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+        parameters,
+        anyOptional
+    )
+    val parametersByName = parameters.associateBy { it.name }
 
     val bindingsByName = LinkedHashMap<String, KotlinJsonAdapter.Binding<Any, Any?>>()
 
-    for (property in rawTypeKotlin.memberProperties) {
-      val parameter = parametersByName[property.name]
+    // TODO this doesn't cover platform types
```
swankjesse, line 297: what does this mean we have to do?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+        }
       }
 
-      if (property !is KMutableProperty1 && parameter == null) continue
+      if (!Flag.Property.IS_VAR(property.flags) && parameterData == null) continue
 
-      property.isAccessible = true
-      val allAnnotations = property.annotations.toMutableList()
-      var jsonAnnotation = property.findAnnotation<Json>()
+      val getterMethod = property.getterSignature?.let(signatureSearcher::findMethod)
```
swankjesse, line 329: I like this pattern

swankjesse: though extensions could do nicer? 

```
val getterMethod = signatureSearcher.getter(property)
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+        Int::class.java -> 0
+        Float::class.java -> 0f
+        Long::class.java -> 0L
+        Double::class.java -> 0.0
+        Void.TYPE -> throw IllegalStateException("Parameter with void type is illegal")
+        else -> throw UnsupportedOperationException("Unknown primitive: $type")
+      }
+    } else null
+
+private fun Class<*>.toKmClass(throwOnNotClass: Boolean): KmClass? {
```
swankjesse, line 519: this wants to be 2 different symbols:

```
  val Class<*>.header: KotlinClassHeader?
  val KotlinClassHeader.kmClass KmClass?
```

That way callers can use `?:` rather than the awkward `throwOnNotClass` thing

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      throw IllegalStateException("Cannot serialize class with metadata $classMetadata")
+    } else {
+      return null
+    }
+  }
+
+  return classMetadata.toKmClass()
+}
+
+private class JvmSignatureSearcher(clazz: Class<*>) {
```
swankjesse, line 545: Is this just `KtClass` ?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      if (nextSignature == signatureString) {
+        return next
+      }
+    }
+
+    return null
+  }
+}
+
+internal data class ParameterData(
```
swankjesse, line 607: Is this just `KtParameter` ?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+    val km: KmValueParameter,
+    val index: Int,
+    val rawType: Class<*>,
+    val annotations: List<Annotation>
+) {
+  val name get() = km.name
+  val declaresDefaultValue get() = Flag.ValueParameter.DECLARES_DEFAULT_VALUE(km.flags)
+}
+
+internal data class ConstructorData(
```
swankjesse, line 617: Is this just `KtConstructor` ?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+
+    // DefaultConstructorMarker
+    arguments.add(null)
+
+    @Suppress("UNCHECKED_CAST")
+    return jvm.newInstance(*arguments.toTypedArray()) as R
+  }
+}
+
+internal data class PropertyData(
```
swankjesse, line 674: Is this just `KtProperty` ?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
swankjesse: These types should be in another file

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+private class JvmSignatureSearcher(clazz: Class<*>) {
+
+  private val methodSignatureCache = mutableMapOf<String, Method>()
+  private val fieldSignatureCache = mutableMapOf<String, Field>()
+  private val declaredMethodsIterator by lazy(NONE) { clazz.declaredMethods.iterator() }
+  private val declaredFieldsIterator by lazy(NONE) { clazz.declaredFields.iterator() }
+  private val methodIterator by lazy(NONE) { clazz.methods.iterator() }
+  private val fieldIterator by lazy(NONE) { clazz.fields.iterator() }
+
+  fun findMethod(signature: JvmMethodSignature): Method? {
```
swankjesse, line 554: FWIW, there’s another strategy here . . . decode `signature` into a `Class[]` and then call `getDeclaredMethod`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      methodSignatureCache[nextSignature] = next
+      if (nextSignature == signatureString) {
+        return next
+      }
+    }
+
+    while (methodIterator.hasNext()) {
+      val next = methodIterator.next()
+      val nextSignature = next.jvmMethodSignature
+      methodSignatureCache[nextSignature] = next
```
swankjesse, line 571: weird thing . . . declared methods and methods overlap.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      if (nextSignature == signatureString) {
+        return next
+      }
+    }
+
+    return null
+  }
+
+  fun findField(signature: JvmFieldSignature): Field? {
```
swankjesse, line 580: This definitely feels like it should be a straightup lookup rather than a loop

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+  }
+
+  val name get() = km.name
+
+  val javaType = jvmField?.genericType
+      ?: jvmGetter?.genericReturnType
+      ?: jvmSetter?.genericReturnType
+      ?: error("No type information available for property '${km.name}' with type '${km.returnType.canonicalName}'.")
+
+  val annotations: Set<Annotation> by lazy {
```
swankjesse, line 695: a findAnnotation() method would be much more efficient

README.md
```diff
@@ -613,7 +613,31 @@ Snapshots of the development version are available in [Sonatype's `snapshots` re
...
 
-Moshi contains minimally required rules for its own internals to work without need for consumers to embed their own. However if you are using reflective serialization and R8 or ProGuard, you must add keep rules in your proguard configuration file for your reflectively serialized classes.
+If using Android, the Android Gradle Plugin should automatically extract embedded rules included in 
+Moshi's jar artifacts.
+
+If you need to add them manually, you can add them from the following files:
+ 
+- [`moshi`](https://github.com/square/moshi/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro). 
+- [`moshi-kotlin`](https://github.com/square/moshi/blob/master/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro)
+- You might also need rules for [Okio](https://github.com/square/okio), which is a dependency of this library.
```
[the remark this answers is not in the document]
ZacSweers: I was thinking of its file https://github.com/square/okio/blob/master/okio/src/jvmMain/resources/META-INF/proguard/okio.pro

Still not worth mentioning?

README.md
```diff
@@ -613,7 +613,31 @@ Snapshots of the development version are available in [Sonatype's `snapshots` re
...
+
+**Important note:** These rules only cover Moshi's core machinery. You will still need to add appropriate
+rules for your own code based on your usage of reflection. `moshi-kotlin` relies on introspection of Kotlin's
+`@Metadata` classes, so you must keep appropriate JVM members that they read. In practice, such a set of
+rules could look like this:
+
+```proguard
+# Keep names of Kotlin classes but allow shrinking if unused
+-keepnames @kotlin.Metadata class your.class.Here
```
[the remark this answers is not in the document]
ZacSweers:
```suggestion
-keepnames @kotlin.Metadata class com.example.YourClassHere
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
ZacSweers: Yup. From the linked spec

> A descriptor is a string representing the type of a field or method. Descriptors are represented in the class file format using modified UTF-8 strings (§4.4.7) and thus may be drawn, where not further constrained, from the entire Unicode codespace.
>
> A signature is a string representing the generic type of a field or method, or generic type information for a class declaration.

kotlin/reflect/src/main/resources/META-INF/com.android.tools/r8-upto-1.6.0/moshi-kotlin.pro
ZacSweers: I think @JakeWharton's done this in one of this projects iirc, something to look into 👍

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
ZacSweers: From the PR description

> The behavior change is that moshi-kotlin still allowed extending from and encoding non-kotlin superclasses and their properties. I think this was an oversight when we stopped supporting mixing them in Moshi 1.8, as this doesn't match the behavior of code gen.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: this is what we called this in KotlinPoet-metadata, which I mentally modeled some of this off of. Open to suggestions!

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: yeah, signature may not be present

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Also from the PR description, we should just make a call here

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      val propertyData = PropertyData(
+          km = property,
+          jvmField = propertyField,
+          jvmGetter = getterMethod,
+          jvmSetter = setterMethod,
+          jvmAnnotationsMethod = annotationsMethod,
+          parameter = parameterData
+      )
+      val allAnnotations = propertyData.annotations.toMutableList()
+      val jsonAnnotation = allAnnotations.filterIsInstance<Json>().firstOrNull()
```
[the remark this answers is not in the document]
ZacSweers: Yeah a cheaper approach would be to do a manual search on all the places that might have the annotation in order of likelihood and check isAnnotationPresent() (or whatever the reflection variation is)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+          .map { (fromThis, fromOther) -> fromThis isEqualTo fromOther }
+          // searching for first negative answer
+          .contains(false)
+      if (areNotEqual) return false
+    }
+    // collections are same or they are contains same elements with same order
+    return true
+  }
+
+  private infix fun KmTypeProjection?.isEqualTo(other: KmTypeProjection?): Boolean {
```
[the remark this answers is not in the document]
ZacSweers: yeah they explicitly won't use data classes or implement these because they wanted the classes to be as lightweight as possible for embedded envs ¯\_(ツ)_/¯

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+
+      // TODO not sure if we care about expressing the other type information here
+    }
+  }
+
+/**
+ * Creates a canonical class name as represented in Metadata's [kotlinx.metadata.ClassName], where
+ * package names in this name are separated by '/' and class names are separated by '.'.
+ *
+ * For example: `"org/foo/bar/Baz.Nested"`.
```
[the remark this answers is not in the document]
ZacSweers: input!

README.md
```diff
@@ -613,7 +613,31 @@ Snapshots of the development version are available in [Sonatype's `snapshots` re
...
+- [`moshi`](https://github.com/square/moshi/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro). 
+- [`moshi-kotlin`](https://github.com/square/moshi/blob/master/kotlin/reflect/src/main/resources/META-INF/proguard/moshi-kotlin.pro)
+- You might also need rules for [Okio](https://github.com/square/okio), which is a dependency of this library.
+
+**Important note:** These rules only cover Moshi's core machinery. You will still need to add appropriate
+rules for your own code based on your usage of reflection. `moshi-kotlin` relies on introspection of Kotlin's
+`@Metadata` classes, so you must keep appropriate JVM members that they read. In practice, such a set of
+rules could look like this:
+
+```proguard
```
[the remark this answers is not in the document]
ZacSweers: That wouldn't really work for people that use reflect in debug and code gen for release I think

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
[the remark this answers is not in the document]
ZacSweers: I think this actually isn't avoidable, because we have to greedily check all annotations to see if they're a JsonQualifier too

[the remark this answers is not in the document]
swankjesse: Okay, we can start with this and potentially optimize later if there’s something the Java API offers.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      other: Collection<KmTypeProjection>): Boolean {
+    // check collections aren't same
+    if (this !== other) {
+      // fast check of sizes
+      if (this.size != other.size) return false
+      val areNotEqual = this.asSequence()
+          .zip(other.asSequence())
+          // check this and other contains same elements at position
+          .map { (fromThis, fromOther) -> fromThis isEqualTo fromOther }
+          // searching for first negative answer
```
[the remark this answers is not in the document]
ZacSweers: considering they're sequences, I don't think an indexed loop would work here?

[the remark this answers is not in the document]
ZacSweers: I guess if we make it speak lists it works

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+}
+
+private val KmType.isNullable: Boolean get() = Flag.Type.IS_NULLABLE(flags)
+private val KmType.canonicalName: String
+  get() {
+    return buildString {
+      val classifierString = when (val cl = classifier) {
+        is KmClassifier.Class -> createClassName(cl.name)
+        is TypeAlias -> createClassName(cl.name)
+        is TypeParameter -> arguments[cl.id].type?.canonicalName ?: "TypeVar(${cl.id})"
```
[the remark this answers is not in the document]
ZacSweers: can you elaborate?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
ZacSweers: ed9d6e7

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: 0d679b1

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
ZacSweers: 5af57b6

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: I suppose not, will make it an error call instead 20f762f

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
 
     val nonTransientBindings = bindings.filterNotNull()
     val options = JsonReader.Options.of(*nonTransientBindings.map { it.name }.toTypedArray())
-    return KotlinJsonAdapter(constructor, bindings, nonTransientBindings, options).nullSafe()
+    return KotlinJsonAdapter(constructorData, bindings, nonTransientBindings, options).nullSafe()
+  }
+
+  private infix fun KmType?.isEqualTo(other: KmType?): Boolean {
```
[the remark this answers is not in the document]
ZacSweers: bb51ea1

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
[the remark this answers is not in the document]
ZacSweers: 5555cee

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+      if (this.size != other.size) return false
+      val areNotEqual = this.asSequence()
+          .zip(other.asSequence())
+          // check this and other contains same elements at position
+          .map { (fromThis, fromOther) -> fromThis isEqualTo fromOther }
+          // searching for first negative answer
+          .contains(false)
+      if (areNotEqual) return false
+    }
+    // collections are same or they are contains same elements with same order
```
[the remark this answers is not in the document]
ZacSweers: 5555cee

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+    return buildString {
+      val classifierString = when (val cl = classifier) {
+        is KmClassifier.Class -> createClassName(cl.name)
+        is TypeAlias -> createClassName(cl.name)
+        is TypeParameter -> arguments[cl.id].type?.canonicalName ?: "TypeVar(${cl.id})"
+      }
+      append(classifierString)
+
+      val args = arguments.joinToString(", ") {
+        "${it.variance?.name} ${it.type?.canonicalName ?: "*"}".trim()
```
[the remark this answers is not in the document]
ZacSweers: Accident from something before I think 64013a4

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+/**
+ * Creates a canonical class name as represented in Metadata's [kotlinx.metadata.ClassName], where
+ * package names in this name are separated by '/' and class names are separated by '.'.
+ *
+ * For example: `"org/foo/bar/Baz.Nested"`.
+ *
+ * Local classes are prefixed with ".", but for KotlinPoetMetadataSpecs' use case we don't deal
+ * with those.
+ */
+private fun createClassName(kotlinMetadataName: String): String {
```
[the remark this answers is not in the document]
ZacSweers: it's not in this case, leftover from the kotlinpoet implementation where we do want to know the package name separately. 

324f6b7

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -284,6 +372,333 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
+    "Local/anonymous classes are not supported!"
+  }
+  // Top-level: package/of/class/MyClass
+  // Nested A:  package/of/class/MyClass.NestedClass
+  val simpleName = kotlinMetadataName.substringAfterLast(
+      '/', // Drop the package name, e.g. "package/of/class/"
+      '.' // Drop any enclosing classes, e.g. "MyClass."
+  )
+  val packageName = kotlinMetadataName.substringBeforeLast("/", missingDelimiterValue = "")
+  val simpleNames = kotlinMetadataName.removeSuffix(simpleName)
```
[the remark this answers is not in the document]
ZacSweers: Removed! 
324f6b7

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: 8b42ed1

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: c727eec

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: c727eec

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: c727eec

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: 5db8757

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Cleaned up a bit in 5db8757

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: 5db8757

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: Per above - needed due to greedy lookup of all annotations to look for json qualifiers

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: f3e3003

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -204,55 +230,119 @@ class KotlinJsonAdapterFactory : JsonAdapter.Factory {
     require(!rawType.isLocalClass) {
       "Cannot serialize local class or object expression ${rawType.name}"
     }
-    val rawTypeKotlin = rawType.kotlin
-    require(!rawTypeKotlin.isAbstract) {
+
+    val kmClass = rawType.toKmClass(throwOnNotClass = true) ?: return null
```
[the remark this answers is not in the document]
ZacSweers: What would that look like API-wise?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: f3e3003

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
ZacSweers: 09dd16c

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
ZacSweers: Resolving as it's been removed anyway

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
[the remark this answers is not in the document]
ZacSweers: Doc updated in 324f6b7

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+    return buildString {
+      val classifierString = when (val cl = classifier) {
+        is KmClassifier.Class -> createClassName(cl.name)
+        is TypeAlias -> createClassName(cl.name)
+        is TypeParameter -> arguments[cl.id].type?.canonicalName ?: "TypeVar(${cl.id})"
+      }
+      append(classifierString)
+
+      val args = arguments.joinToString(", ") {
+        "${it.variance?.name} ${it.type?.canonicalName ?: "*"}"
```
swankjesse, line 64: if there’s no variance?

swankjesse: The `*` is surprising on this line. I understood `*` to mean `out Any`, not unknown

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+      }
+
+      if (args.isNotBlank()) {
+        append('<')
+        append(args)
+        append('>')
+      }
+
+      // TODO not sure if we care about expressing the other type information here
```
swankjesse, line 73: anything in particular? I’d rather just wait for tests to fail

swankjesse: (the TODO seems needlessly ominous)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+  }
+
+/**
+ * Creates a canonical class name as represented in Metadata's [kotlinx.metadata.ClassName], where
+ * package names in this name are separated by '/' and class names are separated by '.'.
+ *
+ * Example ClassName that we want to canonicalize: `"org/foo/bar/Baz.Nested"`.
```
swankjesse, line 81: wanna use Map.Entry? It’s a well known nested class so it’s intuitive what the package part is and the nested class part is

swankjesse: (with this it’s not _that_ obvious whether the package is org.foo.bar or org.foo.bar.Baz)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * Local classes are prefixed with ".", but for Moshi's use case we don't deal with those.
+ */
+private fun createClassName(kotlinMetadataName: String): String {
+  require(!kotlinMetadataName.isLocal) {
+    "Local/anonymous classes are not supported!"
```
swankjesse, line 87: wanna include the local class? That way if this trips you aren’t debugging a massive codebase trying to figure out what changed

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+  }
+  return kotlinMetadataName.replace("/", ".")
+}
+
+internal data class KtParameter(
```
swankjesse, line 92: nice

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+  val index: Int,
+  val rawType: Class<*>,
+  val annotations: List<Annotation>
+) {
+  val name get() = km.name
+  val declaresDefaultValue get() = Flag.ValueParameter.DECLARES_DEFAULT_VALUE(km.flags)
+  val isNullable get() = km.type!!.isNullable
+}
+
+internal data class KtConstructor(
```
swankjesse, line 103: nice

swankjesse: I like the kt/km split

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+        masks.add(mask)
+        mask = 0
+      }
+
+      when {
+        args.containsKey(parameter) -> {
+          arguments.add(args[parameter])
+        }
+        parameter.declaresDefaultValue -> {
+          arguments += defaultPrimitiveValue(parameter.rawType)
```
swankjesse, line 132: the use of `+=` here and `.add()` above hurts me

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+    val arguments = ArrayList<Any?>(parameters.size)
+    var mask = 0
+    val masks = ArrayList<Int>(1)
+    var index = 0
+
+    @Suppress("UseWithIndex")
+    for (parameter in parameters) {
+      if (index != 0 && index % Integer.SIZE == 0) {
+        masks.add(mask)
+        mask = 0
```
swankjesse, line 124: you could do `index = 0` here too and save some modding later

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+  val km: KmConstructor,
+  val jvm: Constructor<*>,
+  val parameters: List<KtParameter>,
+  val isDefault: Boolean
+) {
+  init {
+    jvm.isAccessible = true
+  }
+
+  fun <R> callBy(args: Map<KtParameter, Any?>): R {
```
swankjesse, line 114: args –> argumentsMap

using args and arguments as different things is gross. Too easy to mix up.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+
+    @Suppress("UseWithIndex")
+    for (parameter in parameters) {
+      if (index != 0 && index % Integer.SIZE == 0) {
+        masks.add(mask)
+        mask = 0
+      }
+
+      when {
+        args.containsKey(parameter) -> {
```
swankjesse, line 128: any way you can save a map lookup? You’re doing containsKey and get, but you probably only need to look inside the map 1x not 2x. (worst case the map contains null, and you need to check containsKey() if get() returned null)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+
+    // DefaultConstructorMarker
+    arguments.add(null)
+
+    @Suppress("UNCHECKED_CAST")
+    return jvm.newInstance(*arguments.toTypedArray()) as R
+  }
+
+  companion object {
+    fun create(rawType: Class<*>, kmClass: KmClass): KtConstructor? {
```
swankjesse, line 161: should this method be named primary?

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+      val anyOptional = parameters.any { it.declaresDefaultValue }
+      val actualConstructor = if (anyOptional) {
+        val prefix = jvmConstructor.jvmMethodSignature.removeSuffix(")V")
+        val parameterCount = jvmConstructor.parameterTypes.size
+        val maskParamsToAdd = if (parameterCount == 0) {
+          0
+        } else {
+          (parameterCount + 31) / 32
```
swankjesse, line 181: why is this a special case?

swankjesse: ie. `0 + 31 / 32 == 0`

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+    jvmGetter?.isAccessible = true
+    jvmSetter?.isAccessible = true
+  }
+
+  val name get() = km.name
+
+  val javaType = jvmField?.genericType
+    ?: jvmGetter?.genericReturnType
+    ?: jvmSetter?.genericReturnType
+    ?: error(
```
swankjesse, line 226: should this be a condition on init that the property has at least one of these non-null?

kotlin/reflect/src/test/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtilTest.kt
```diff
@@ -0,0 +1,51 @@
+package com.squareup.moshi.kotlin.reflect
```
swankjesse, line 1: supernit: copyright

kotlin/reflect/src/test/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtilTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Test
+
+class JvmDescriptorUtilTest {
+
+  private fun getTypesOf(desc: String): List<Class<*>> {
+    return JvmMethodSignature("test", desc).decodeParameterTypes()
+  }
+
+  @Test
+  fun readTypes_primitives() {
```
swankjesse, line 14: cute

kotlin/reflect/src/test/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtilTest.kt
```diff
@@ -0,0 +1,51 @@
+package com.squareup.moshi.kotlin.reflect
+
+import com.google.common.truth.Truth.assertThat
+import kotlinx.metadata.jvm.JvmMethodSignature
+import org.junit.Test
+
+class JvmDescriptorUtilTest {
```
swankjesse, line 7: Nice test

swankjesse (approved):
LGTM on everything but KotlinJsonAdapter.kt , which I still need to read!

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
swankjesse: ie. add the name to the failure message

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
```diff
@@ -0,0 +1,146 @@
+/*
```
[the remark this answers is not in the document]
ZacSweers: yeah that's fair, this was originally based on what we have in KotlinPoet

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+  } else null
+
+internal val KmType.canonicalName: String
+  get() {
+    return buildString {
+      val classifierString = when (val cl = classifier) {
+        is KmClassifier.Class -> createClassName(cl.name)
+        is TypeAlias -> createClassName(cl.name)
+        is TypeParameter -> arguments[cl.id].type?.canonicalName ?: "TypeVar(${cl.id})"
```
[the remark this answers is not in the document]
ZacSweers: yeah just trying to do something reasonable here, I don't think this would ever come up at runtime

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: From the doc

> @property type the projected type, or `null` if this is a star projection

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: I don't remember 😅. Easy to fix

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: not sure I follow?

kotlin/reflect/src/test/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtilTest.kt
ZacSweers: Merge #1231 so CI can supernit instead

README.md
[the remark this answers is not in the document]
ZacSweers: 895b0e4

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtil.kt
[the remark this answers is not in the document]
ZacSweers: d746fed

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: f12a084

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: 1d1de9a

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: Good idea f8569b8

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: Cleaned up some b63a727

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: 02cae9a

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: ea6f67a

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: b0950d9

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: 264cdb5

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: f54a5b0

kotlin/reflect/src/test/java/com/squareup/moshi/kotlin/reflect/JvmDescriptorUtilTest.kt
ZacSweers: e341005

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
ZacSweers: and 8f369d062c3fc0fc369fb677ce321d58681d5d1d (committed halfway)

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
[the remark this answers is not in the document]
ZacSweers: I remembered why, this is missing a .orEmpty() on the nullable name and the trim() was to handle if it was blank 3ac5c1fcd38a7d654864b45645af2742df8c5baf

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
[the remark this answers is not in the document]
ZacSweers: Actually after looking at the doc for the other comment, I realize this case means star f813e5d62b6cbee4cc498d4ec822ca1416d94844

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
swankjesse: Umm I think this happens where it should. Disregard me!

ZacSweers:
For those wanting to try this early, I've published this in a separate project for now https://github.com/ZacSweers/MoshiX/tree/main/moshi-metadata-reflect

oharaandrew314:
Thanks for the library @ZacSweers; it brings my shaded jar down to 8 MB, which is perfect for AWS Lambda.  I'm very disappointed this PR was approved and then left to rot.

ZacSweers:
Closing for now. This is available in github.com/zacsweers/moshix

unoexperto:
@swankjesse Why wasn't this merged if you don't mind me asking ?

ZacSweers:
@unoexperto He didn't make the decision, I did. I closed it because it's fine living in that repo in the meantime and was relying on unstable APIs that didn't have a definite stability point

PR #1185, opened 2020-08-02
Make JsonReader.promoteNameToValue() a public API

swankjesse:


JakeWharton:
What about `promoteValueToName`? Seems like we should do both or none.

swankjesse:
We’ve got one use case for `promoteValueToName()`: it lets you reuse existing `JsonAdapters` for map keys without their knowledge. Moshi uses it internally for things like maps with enum keys.

I can’t think of a use case for `promoteValueToName`.

swankjesse:
We’ve got one use case for promoteNameToValue(): it lets you reuse existing JsonAdapters for map keys without their knowledge. Moshi uses it internally for things like maps with enum keys.

I can’t think of a use case for promoteValueToName().

JakeWharton:
Isn't it the symmetric same use case? Using an adapter to write a value as a name.

swankjesse:
Yeah but because we need the value case much much more we can solve both cases with only this one.

JakeWharton:
I don't follow. How do I write an adapter for a `Map<T, String>` given only an opaque adapter for `T`?

swankjesse:
Ahhhhhhhh, yep I totally missed this. I was worrying all about reading these things that I forgot all about writing ’em. That is now fixed, please take another look!

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -514,8 +514,19 @@ public final void setFailOnUnknown(boolean failOnUnknown) {
   /**
    * Changes the reader to treat the next name as a string value. This is useful for map adapters so
    * that arbitrary type adapters can use {@link #nextString} to read a name value.
+   *
```
JakeWharton, line 517: suuuuuper nit: this is Java and you need to take a `<p>` 😢

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -443,8 +443,20 @@ public final JsonWriter jsonValue(@Nullable Object value) throws IOException {
   /**
    * Changes the writer to treat the next value as a string name. This is useful for map adapters so
    * that arbitrary type adapters can use {@link #value} to write a name value.
+   *
```
JakeWharton, line 446: Also `<p>`

moshi/src/main/java/com/squareup/moshi/JsonReader.java
swankjesse: OMG how quickly I forget!

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
swankjesse: ℙ

PR #1195, opened 2020-08-27
Add missing 1.10 bits to changelog

ZacSweers:
Some were cherrypicked to 1.9.3, this adds (I think) everything else that's fixed or new in 1.10


swankjesse:
Thanks!

PR #1196, opened 2020-08-28
Switch to spotless and format code

ZacSweers:
After merging, I can make a second PR setting up the format commit as an ignored revision to better protect history

Note that I haven't enabled licenses yet because there seems to be a bug in Spotless where it always overwrites the copyright year even if we had one set. https://github.com/diffplug/spotless/issues/677


nedtwigg:
I might pause on merging this for a moment to evaluate `ratchetFrom 'origin/main'`, I'll explain in the issue you raised on Spotless.  Moshi is great btw, thanks!

adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.java
```diff
@@ -26,11 +26,12 @@
...
  * https://github.com/FasterXML/jackson-databind/blob/67ebf7305f492285a8f9f4de31545f5f16fc7c3a/src/main/java/com/fasterxml/jackson/databind/util/ISO8601Utils.java
  *
- * Utilities methods for manipulating dates in iso8601 format. This is much much faster and GC
+ * <p>Utilities methods for manipulating dates in iso8601 format. This is much much faster and GC
  * friendly than using SimpleDateFormat so highly suitable if you (un)serialize lots of date
  * objects.
  *
- * Supported parse format: [yyyy-MM-dd|yyyyMMdd][T(hh:mm[:ss[.sss]]|hhmm[ss[.sss]])]?[Z|[+-]hh[:]mm]]
+ * <p>Supported parse format:
+ * [yyyy-MM-dd|yyyyMMdd][T(hh:mm[:ss[.sss]]|hhmm[ss[.sss]])]?[Z|[+-]hh[:]mm]]
```
swankjesse, line 34: this one’s weird

adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
```diff
@@ -27,17 +30,17 @@
...
-    Moshi moshi = new Moshi.Builder()
-        .add(PolymorphicJsonAdapterFactory.of(Message.class, "type")
-            .withSubtype(Success.class, "success")
-            .withSubtype(Error.class, "error"))
-        .build();
+  @Test
+  public void fromJson() throws IOException {
+    Moshi moshi =
+        new Moshi.Builder()
+            .add(
```
swankjesse, line 39: ew

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
```diff
@@ -159,17 +170,17 @@ internal class AdapterGenerator(
...
-          prop.type.rawType() == JsonAdapter::class.asClassName()
-        }
-        .filter { prop -> prop.annotations.isNotEmpty() }
-        .mapTo(mutableSetOf()) { prop ->
-          QualifierAdapterProperty(
-              name = prop.name,
-              qualifiers = prop.annotations.mapTo(mutableSetOf()) { it.className }
-          )
-        }
+      .asSequence()
```
swankjesse, line 173: still should be +4 on wrapped lines, right?

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
```diff
@@ -84,136 +86,157 @@
...
-      @FromJson @Uppercase String fromJson(@Uppercase String string) {
-        return string.substring(1, string.length() - 1);
-      }
-
-      @ToJson @Uppercase String toJson(@Uppercase String value) {
-        return "|" + value + "|";
-      }
-    }).add(new UppercaseAdapterFactory()).build();
+  @Test
+  public void toAndFromWithIntermediateWithQualifier() throws Exception {
```
swankjesse, line 169: I think this is strictly worse

swankjesse:
Also please don’t address my feedback above; I’m just ranting at bad choices made by the formatter

moshi/src/test/java/com/squareup/moshi/AdapterMethodsTest.java
JakeWharton: https://github.com/google/google-java-format/issues/342

adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
JakeWharton: you get used to it

https://github.com/google/google-java-format/wiki/The-Rectangle-Rule

adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.java
JakeWharton:
```
jshell> " * Supported parse format: [yyyy-MM-dd|yyyyMMdd][T(hh:mm[:ss[.sss]]|hhmm[ss[.sss]])]?[Z|[+-]hh[:]mm]]".length()
$1 ==> 101
```

JakeWharton:
:shipit:. Switched over Retrofit a while ago and I don't care. Same for using ktlint on a bunch of projects. Whevs.

swankjesse:
SGTM. Thanks Zach!

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
ZacSweers: Ktlint doesn't support continuation indent anymore, and the compose and kotlin people sold me recently on it being a "you get used to it" thing. The kotlin plugin is iffy on its support of it too

JakeWharton: Yep, you won't notice in a week.

PR #1198, opened 2020-08-28
Update Okio and kotlin-compile-testing dependencies

ZacSweers:
Okio is opportunistic, kotlin-compile-testing adds kotlin 1.4.0 support and allows us to re-enable the previously ignored tests


ZacSweers:
Rebased and added an updated ktlint version that I forgot to push before merging the formatting update

PR #1202, opened 2020-08-29
Promote Kotlin type-inferring APIs to the main Moshi package (round 2!)

ZacSweers:
Continuation of #969, but with Kotlin 1.4 and formal availability of reflect-less `typeOf()`

* This adds extensions in a few places. We probably don't want all of them, but putting them here for consideration based on personal experience and ones I've found myself writing in multiple projects.
* Makes the `moshi` artifact a kotlin project but with the stdlib only as a `compileOnly` dependency. Only kotlin consumers would be able to use these APIs
* Fixed a couple of gradle/java version configuration I noticed along the way
* Some APIs seem inviting, like `Class<*>.parameterizedBy(...)`, but I avoided adding these because the whole point is that `typeOf()` allows us to avoid these.

### DIFF
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DefaultConstructorTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DefaultConstructorTest.kt
@@ -17,6 +17,7 @@ package com.squareup.moshi.kotlin
 
 import com.squareup.moshi.JsonClass
 import com.squareup.moshi.Moshi
+import com.squareup.moshi.adapter
 import org.junit.Test
 
 class DefaultConstructorTest {
@@ -25,7 +26,7 @@ class DefaultConstructorTest {
     val expected = TestClass("requiredClass")
     val json =
       """{"required":"requiredClass"}"""
-    val instance = Moshi.Builder().build().adapter<TestClass>(TestClass::class.java)
+    val instance = Moshi.Builder().build().adapter<TestClass>()
       .fromJson(json)!!
     check(instance == expected) {
       "No match:\nActual  : $instance\nExpected: $expected"
@@ -36,7 +37,7 @@ class DefaultConstructorTest {
     val expected = TestClass("requiredClass", "customOptional", 4, "setDynamic", 5, 6)
     val json =
       """{"required":"requiredClass","optional":"customOptional","optional2":4,"dynamicSelfReferenceOptional":"setDynamic","dynamicOptional":5,"dynamicInlineOptional":6}"""
-    val instance = Moshi.Builder().build().adapter<TestClass>(TestClass::class.java)
+    val instance = Moshi.Builder().build().adapter<TestClass>()
       .fromJson(json)!!
     check(instance == expected) {
       "No match:\nActual  : $instance\nExpected: $expected"
@@ -47,7 +48,7 @@ class DefaultConstructorTest {
     val expected = TestClass("requiredClass", "customOptional")
     val json =
       """{"required":"requiredClass","optional":"customOptional"}"""
-    val instance = Moshi.Builder().build().adapter<TestClass>(TestClass::class.java)
+    val instance = Moshi.Builder().build().adapter<TestClass>()
       .fromJson(json)!!
     check(instance == expected) {
       "No match:\nActual  : $instance\nExpected: $expected"
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -448,7 +448,7 @@ class KotlinJsonAdapterTest {
       fail()
     } catch (e: IllegalArgumentException) {
       assertThat(e).hasMessageThat().isEqualTo(
-        "Platform class kotlin.Triple in kotlin.Triple<java.lang.Object, java.lang.Object, java.lang.Object> requires explicit JsonAdapter to be registered"
+        "Platform class kotlin.Triple in kotlin.Triple<?, ?, ?> requires explicit JsonAdapter to be registered"
       )
     }
   }
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -858,13 +858,7 @@ class KotlinJsonAdapterTest {
 
   @Test fun genericTypes() {
     val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
-    val stringBoxAdapter = moshi.adapter<Box<String>>(
-      Types.newParameterizedTypeWithOwner(
-        KotlinJsonAdapterTest::class.java,
-        Box::class.java,
-        String::class.java
-      )
-    )
+    val stringBoxAdapter = moshi.adapter<Box<String>>()
     assertThat(stringBoxAdapter.fromJson("""{"data":"hello"}""")).isEqualTo(Box("hello"))
     assertThat(stringBoxAdapter.toJson(Box("hello"))).isEqualTo("""{"data":"hello"}""")
   }
@@ -873,18 +867,7 @@ class KotlinJsonAdapterTest {
 
   @Test fun nestedGenericTypes() {
     val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
-    val type = Types.newParameterizedTypeWithOwner(
-      KotlinJsonAdapterTest::class.java,
-      NestedGenerics::class.java,
-      String::class.java,
-      Int::class.javaObjectType,
-      Types.newParameterizedTypeWithOwner(
-        KotlinJsonAdapterTest::class.java,
-        Box::class.java,
-        String::class.java
-      )
-    )
-    val adapter = moshi.adapter<NestedGenerics<String, Int, Box<String>>>(type).indent("  ")
+    val adapter = moshi.adapter<NestedGenerics<String, Int, Box<String>>>().indent("  ")
     val json =
       """
       |{
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -26,9 +26,8 @@ import com.squareup.moshi.JsonReader
 import com.squareup.moshi.JsonWriter
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.ToJson
-import com.squareup.moshi.Types
-import org.assertj.core.api.Assertions
 import com.squareup.moshi.adapter
+import org.assertj.core.api.Assertions
 import org.junit.Assert.fail
 import org.junit.Test
 import java.io.ByteArrayOutputStream
--- a/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -28,6 +28,7 @@ import com.squareup.moshi.Moshi
 import com.squareup.moshi.ToJson
 import com.squareup.moshi.Types
 import org.assertj.core.api.Assertions
+import com.squareup.moshi.adapter
 import org.junit.Assert.fail
 import org.junit.Test
 import java.io.ByteArrayOutputStream
--- a/moshi/gradle.properties
+++ b/moshi/gradle.properties
@@ -18,3 +18,7 @@ POM_NAME=Moshi
 POM_ARTIFACT_ID=moshi
 POM_PACKAGING=jar
 AUTOMATIC_MODULE_NAME=com.squareup.moshi
+
+# Kotlin adds the stdlib dep by default in 1.4.0+, but we want to effectively make it compileOnly
+# for our case to avoid imposing it on consumers.
+kotlin.stdlib.default.dependency=false
--- a/moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
+++ b/moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Square, Inc.
+ * Copyright (C) 2020 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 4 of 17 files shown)


ZacSweers:
One open comment from the previous PR around `adapter(KType)`'s avoidance of re-wrapping nonNull/nullSafe adapters from https://github.com/square/moshi/pull/969/files#r339377867

> Yep, I think we add a TypeAdapter.Factory that takes a KType and makes the adapters responsible for this decision.

@swankjesse can you elaborate on this? I didn't quite see how we'd implement that

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
```diff
@@ -32,11 +29,14 @@ import kotlin.reflect.typeOf
...
-}
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.adapter(): JsonAdapter<T> = adapter(typeOf<T>())
+
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.Builder.addAdapter(adapter: JsonAdapter<T>): Moshi.Builder = add(typeOf<T>().toType(), adapter)
 
-inline fun <reified T> Moshi.Builder.addAdapter(adapter: JsonAdapter<T>) = add(typeOf<T>().toType(), adapter)
+@ExperimentalStdlibApi
+inline fun <reified T> Moshi.nextAdapter(skipPast: JsonAdapter.Factory, annotations: Set<Annotation>): JsonAdapter<T> = nextAdapter(skipPast, typeOf<T>().toType(), annotations)
```
ZacSweers, line 39: I actually wonder if this is possibly a footgun and should be left out. Often this is used in more abstract cases you don't actually know the type parameter and are just deferring to the `type` param. I spent a minute scratching my head at some test failures before realizing that it was because it was using a reified `Any` as the adapter to look up rather than the type.

JakeWharton: I vote to leave it out. We can always add, we can't remove.

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
```diff
@@ -0,0 +1,74 @@
...
+import kotlin.reflect.typeOf
+
+/** Returns the raw [Class] type of this type. */
+val Type.rawType: Class<*> get() = Types.getRawType(this)
+
+/**
+ * Checks if [this] contains [T]. Returns the subset of [this] without [T], or null if
+ * [this] does not contain [T].
+ */
+inline fun <reified T : Annotation> Set<Annotation>.nextAnnotations(): Set<Annotation>? = Types.nextAnnotations(this, T::class.java)
```
JakeWharton, line 34: I would probably name this "filter"-something

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+    val moshi = Moshi.Builder()
+      .addAdapter(customIntdapter)
+      .build()
+
+    check(moshi.adapter<Int>().fromJson("5") == -1)
+  }
+
+  private inline fun <reified T> obtainType(): KType {
```
JakeWharton, line 64: Why is this indirection needed?

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,67 @@
...
+@TestAnnotation2
+class KotlinExtensionsTest {
+
+  @Test
+  fun nextAnnotationsShouldWork() {
+    val annotations = KotlinExtensionsTest::class.java.annotations
+      .filterTo(mutableSetOf()) {
+        it.annotationClass.java.isAnnotationPresent(JsonQualifier::class.java)
+      }
+    check(annotations.size == 2)
```
JakeWharton, line 26: Use assertEquals and the like in this class

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -643,4 +662,12 @@ public static JsonDataException unexpectedNull(
     }
     return new JsonDataException(message);
   }
+
+  // Public due to inline access in MoshiKotlinTypesExtensions
+  public static <T> Class<T> wrap(Class<T> type) {
```
JakeWharton, line 667: Maybe `boxIfNeeded`? or `boxIfPrimitive`? Without looking at the implementation I had no idea what this was going to do.

ZacSweers: I was just using the name that guava's Primitives.java used, but can rename

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: Accidentally left over from debugging something!

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
ZacSweers: I feel like that might be semantically weird since this returns null if empty rather than an actual empty list, also matches the Types method

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
ZacSweers: 146780a

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: 8feb66e

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: d7a0c8f

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: 02848bd

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DefaultConstructorTest.kt
```diff
@@ -32,7 +33,7 @@ class DefaultConstructorTest {
     val expected = TestClass("requiredClass", "customOptional")
     val json =
       """{"required":"requiredClass","optional":"customOptional"}"""
-    val instance = Moshi.Builder().build().adapter<TestClass>(TestClass::class.java)
+    val instance = Moshi.Builder().build().adapter<TestClass>()
```
swankjesse, line 36: this is rad

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -849,13 +849,7 @@ class KotlinJsonAdapterTest {
...
   @Test fun genericTypes() {
     val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
-    val stringBoxAdapter = moshi.adapter<Box<String>>(
-      Types.newParameterizedTypeWithOwner(
-        KotlinJsonAdapterTest::class.java,
-        Box::class.java,
-        String::class.java
-      )
-    )
+    val stringBoxAdapter = moshi.adapter<Box<String>>()
```
swankjesse, line 852: so much better

moshi/gradle.properties
```diff
@@ -18,3 +18,7 @@ POM_NAME=Moshi
 POM_ARTIFACT_ID=moshi
 POM_PACKAGING=jar
 AUTOMATIC_MODULE_NAME=com.squareup.moshi
+
+# Kotlin adds the stdlib dep by default in 1.4.0+, but we want to effectively make it compileOnly
```
swankjesse, line 22: Thanks for adding this comment

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Returns a type that represents an unknown type that extends [T]. For example, if
+ * [T] is [CharSequence], this returns `? extends CharSequence`. If
+ * [T] is [Any], this returns `?`, which is shorthand for `? extends Object`.
```
swankjesse, line 39: nit: weird to have `?` in the docs which is a Java concept, whereas Kotlin uses `*`

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
```diff
@@ -0,0 +1,74 @@
+/*
```
swankjesse, line 1: Does the name of this class show up in the bytecode?

swankjesse: I’m tempted to request something else.

For some of the methods here . . . do the signatures interop with Moshi’s `Types` class? Can we rewrite that class in Kotlin and do the Kotlin conversion? That way we share the API for Java and Kotlin, rather than having one for each?

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
```diff
@@ -0,0 +1,74 @@
...
+inline fun <reified T> supertypeOf(): WildcardType {
+  var type = typeOf<T>().javaType
+  if (type is Class<*>) {
+    type = Util.boxIfPrimitive(type)
+  }
+  return Types.supertypeOf(type)
+}
+
+/** Returns true if [this] and [other] are equal. */
+fun Type?.isEqualTo(other: Type?): Boolean = Types.equals(this, other)
```
swankjesse, line 64: this is the biggest challenge of the lot. I almost prefer forcing Kotlin users to use the static function which is ugly, because this is also ugly

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,64 @@
+package com.squareup.moshi
```
swankjesse, line 1: supernit: copyright

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,64 @@
...
+@JsonQualifier
+@Retention(RUNTIME)
+annotation class TestAnnotation2
+
+@TestAnnotation1
+@TestAnnotation2
+class KotlinExtensionsTest {
+
+  @Test
+  fun nextAnnotationsShouldWork() {
```
swankjesse, line 22: nice

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    check(stringArray.genericComponentType == String::class.java)
+
+    val stringListType = typeOf<List<String>>()
+    val stringListArray = stringListType.asArrayType()
+    val expected = Types.arrayOf(Types.newParameterizedType(List::class.java, String::class.java))
+    assertEquals(stringListArray, expected)
+  }
+
+  @Test
+  fun addAdapterInferred() {
```
swankjesse, line 46: Nice. Will it work if the type is parameterized?

swankjesse (commented):
Nice.

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
ZacSweers: If we rewrote it, we probably would need to make the stdlib a direct dependency rather than compileOnly. I'd be fine with just ditching this class entirely for now and save rewriting for a full rewrite later

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: It should, I'll add a test to be sure

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
ZacSweers: Good catch, missed this when I copied it 8681111

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinTypesExtensions.kt
ZacSweers: I kind of agree. Removed it 1590d84

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: 1590d84

moshi/src/test/java/com/squareup/moshi/KotlinExtensionsTest.kt
ZacSweers: Yup test worked ee1fe41

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Square, Inc.
+ * Copyright (C) 2020 Square, Inc.
```
swankjesse, line 2: FWIW we typically never bump these after creation

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
```diff
@@ -32,11 +29,11 @@ import kotlin.reflect.typeOf
  * @return a [JsonAdapter] for [T], creating it if necessary. Note that while nullability of [T]
  *         itself is handled, nested types (such as in generics) are not resolved.
  */
-inline fun <reified T> Moshi.adapter(): JsonAdapter<T> {
-  return adapter(typeOf<T>())
-}
+@ExperimentalStdlibApi
```
swankjesse, line 32: this is still experimental eh, too bad

swankjesse (approved):
I’d like to land this without the new MoshiKotlinExtensions utilities.

Then we ship a last Java version of Moshi, kick off the Kotlin conversion, and add this in that release. Moshi 2.0 will be the Kotlin release.

ZacSweers:
what else would we land if we disabled those? Should we just wait?

swankjesse:
Lets do this in Moshi 2, the Kotlin update. Biggest challenge is I think we might want to have support for either Java’s or Kotlin’s types in JsonAdapter.Factory. Somehow.

ZacSweers:
sounds good. I have an idea around that I'll send you

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
ZacSweers: 9cd0535

ZacSweers:
For those wanting to try this early, I've published this in a separate project for now https://github.com/ZacSweers/MoshiX/tree/main/moshi-ktx

swankjesse (approved):

swankjesse:
Any concerns with maintaining binary compatibility on these when we jump to Kotlin? I don’t anticipate anything other than the `-Moshi` names in the classes, and I think that’s okay.

ZacSweers:
should be fine as long as we keep the names and paths as-is. We are at some liberty to change the `typeOf()`-using ones as they're behind the kotlin stdlib experimental flag too

PR #1204, opened 2020-08-29
Switch to Github Actions for CI

ZacSweers:
Switches from Travis to GH Actions for CI

* Doesn't run if just markdown/readme files changed
* Builds a matrix of newer JDKs (something started in #681)
* Runs `uploadArchives` on master pushes for snapshots


ZacSweers:
hmm, not sure if nothing ran because I'm not a collaborator or something else 🤔

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+
+      - name: Generate cache key
+        run: ./.github/workflows/checksum.sh checksum.txt
```
JakeWharton, line 41: Skeptical of this complexity compared to its value.

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
+name: CI
+
+on:
+  # Only run push on master
+  push:
+    branches:
+      - master
```
JakeWharton, line 7: This is why it didn't run on your PR. I think we want this to be `on: [push, pull_request]`. I wouldn't bother with the markdown exclusion. PRs modifying only docs are rare and we don't care about wasting GitHub's money.

.github/workflows/build.yml
ZacSweers: we could manually implement it, the main issue is that their examples only cover .gradle files and not properties or wrapper changes

.github/workflows/build.yml
ZacSweers: No it doesn't run because I'm not a collaborator 😅. I use this configuration in another project and it works fine. Can remove the md bits though, it's more useful if/when you check in docs like mkdocs sites and have a longer CI time

ZacSweers: the master-only push is to reduce PR noise, as otherwise it'd run both for any collaborators pushing to branches directly

JakeWharton: That seems like what we want. I push branches to test things on CI all the time (i almost never run tests locally). Moreover, it means that people working in the repo will have their changes tested prior to merge. Plus it's how every other project is set up.

I don't see how this would build your branch even if you pushed your branch to the repo directly.

.github/workflows/build.yml
JakeWharton: I mean caching things probably isn't worth it. We either download 100MB from the cache or we download 100MB from Maven central and Gradle which is also just a cache. What's the difference except one requires no additional build configuration to maintain.

ZacSweers: The github cache is significantly faster in my experience, compared to waiting for gradle to figure it out and roundtrip to central. I don't have strong feelings on it, but think it's marginally worth it for the relatively low cost

.github/workflows/build.yml
ZacSweers: ah! In that case I'll change it. Still fairly certain it won't run regardless though until it's merged

ZacSweers:
Everything is green on my own push: https://github.com/ZacSweers/moshi/actions/runs/233782965

.github/workflows/build.yml
ZacSweers: A good comparison from the runs on my repo that I linked below: cached build took ~2:20, cache misses took ~4min

.github/workflows/build.yml
ZacSweers: 4212163

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Configure JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java-version }}
+
+      - name: Test
+        run: ./gradlew build check
+
+      - name: Publish (default branch only)
+        if: github.ref == 'refs/heads/master' && matrix.java-version == '1.7'
```
JakeWharton, line 48: There is no 1.7 shard. Was this meant to be 1.8?

JakeWharton: Ninja'd on master. Also removed snapshot script.

ZacSweers: Derp, yeah thanks for fixing

PR #1206, opened 2020-08-29
Add .git-blame-ignore-revs

ZacSweers:
Annoyingly this requires users to locally configure this once via `git config blame.ignoreRevsFile .git-blame-ignore-revs`

image


JakeWharton:
I'm not sure this is worth it. I'm used to following anywhere from 3 to 10 commits in repos to find the source of things since git already doesn't track anything but the most basic moves.

ZacSweers:
Yeah inclined to agree since it's not automatic locally

PR #1209, opened 2020-08-30
Moshi event listener proposal

LachlanMcKee:
Relates to issue #1126.

This is a potential counter proposal to https://github.com/square/moshi/pull/1131 in which events are listened to in a more generic way that does require extending/delegating to JsonReader.

This listener would allow the user to listen to events per serialization/deserialization call, which then grants them more context to the events (as they can potentially store state in the listener if they choose to).

I did attempt to also create 'global' listeners that could be added to the `Moshi` builder itself, however it seemed very difficult to pass the listeners to the adapters without creating a breaking change to the API, as the adapters would either need to be passed via the fromJson/toJson methods, or the constructor.

If the adapters had a reference to Moshi itself, it would be trivial to fetch the listeners, however this would also likely be a breaking change.

Any thoughts?

This change is actually quite useful when writing a custom Retrofit Moshi converter as the developer can have context of the API request that has potentially caused an error, and then log the event and any contextual data to their analytics tools.

(Note: there are likely more unit tests that should be added, but I thought I would hold off until the approach was agreed)

### DIFF
--- a/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
@@ -15,7 +15,12 @@
  */
 package com.squareup.moshi.adapters;
 
-import com.squareup.moshi.*;
+import com.squareup.moshi.Json;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonDataException;
+import com.squareup.moshi.JsonEventListener;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
--- a/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
+++ b/adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
@@ -15,13 +15,10 @@
  */
 package com.squareup.moshi.adapters;
 
-import com.squareup.moshi.Json;
-import com.squareup.moshi.JsonAdapter;
-import com.squareup.moshi.JsonDataException;
-import com.squareup.moshi.JsonReader;
-import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.*;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.List;
 import javax.annotation.Nullable;
 
 /**
@@ -97,7 +94,18 @@ public final class EnumJsonAdapter<T extends Enum<T>> extends JsonAdapter<T> {
       throw new JsonDataException(
           "Expected a string but was " + reader.peek() + " at path " + path);
     }
-    reader.skipValue();
+
+    List<JsonEventListener> eventListeners = reader.getEventListeners();
+    if (eventListeners.size() > 0) {
+      UnknownEnumValueEvent unknownValueEvent =
+          new UnknownEnumValueEvent(path, reader.nextString());
+      for (JsonEventListener listener : eventListeners) {
+        listener.onReadEvent(unknownValueEvent);
+      }
+    } else {
+      reader.skipValue();
+    }
+
     return fallbackValue;
   }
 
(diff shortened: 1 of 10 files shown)


adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
```diff
@@ -15,13 +15,10 @@
  */
 package com.squareup.moshi.adapters;
 
-import com.squareup.moshi.Json;
-import com.squareup.moshi.JsonAdapter;
-import com.squareup.moshi.JsonDataException;
-import com.squareup.moshi.JsonReader;
-import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.*;
```
ZacSweers, line 18: While waiting for review, you should fix these star imports

swankjesse:
This is interesting!

Unfortunately there's a few things that'll make it difficult to land:
 - The contents of the events. This needs lots of design, and I don't think it's necessary even feasible.
 - The performance. JSON decoding is inner-loop stuff, and putting arbitrary user code in there is going to wreck throughput.

swankjesse:
Tell me what you think about this: create a `JsonAdapter.Factory` that always calls `nextAdapter()` and wraps it. Your wrapper will forward to the `nextAdapter` delegate and will observe all inputs and outputs at every level of the parse.

LachlanMcKee:
Thanks for your quick reply! I understand your points, and have some thoughts:

Contents of the events
> I'm not quite sure what you mean regarding this one. The core library itself doesn't necessarily need to build ship with any concrete event implementations, and would be up to the user or 3rd party libraries to define.

Performance
> Surely this would be up to the developer (as they would be the ones writing the arbitrary user code)? If they wanted to have a trade-off of performance for potentially finding issues, they can do so. If a developer does not register any listeners, nothing changes in the adapters, as the size of the listeners would be 0.
> Would it be any better if there was only one listener rather than a list? I presume no, as the user could still trigger long running logic.

With regards to the `JsonAdapter.Factory`, I will certainly investigate this. However I am concerned that something like the `EnumAdaper` in the Moshi codebase does not currently output that a default value was used.

Are you suggesting that I tweak the `EnumAdapter` to output this information, but wrap it in a different adapter that logs the information and changes the output to the previous `EnumAdapter` behaviour? Why would this be any more efficient if it still happening at every level of the parse?

adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
LachlanMcKee: done. sorry about that!

swankjesse:
Why not just write your own enum adapter? It's way simpler than a framework to describe what an adapter decided.

LachlanMcKee:
The problem with writing my own enum adapter is that I wouldn't be able to retrieve the data from the JsonReader when the JSON has been deserialized. If I have to register the listener, or data store against the custom enum adapter, it would be very tricky to retrieve the content without using static variables or making the enum adapter a singleton. If I was deserializing multiple JSON streams in parallel it would become tricky to differentiate them.

Without making some form of change to the core library (the listener, or potentially a simpler event data store per JsonReader instance), then I don't see how it is feasible to retrieve events specific to a particular deserialization like you can see in the testcase I wrote for the enum adapter.

Would it be possible to introduce something (similar to what I have proposed) into the core library without modifying the existing enum adapter so that devs/3rd parties can extend the library without negatively affecting the core library performance?

swankjesse:
Would it be useful to have a tags map on the JsonReader and JsonWriter? A place where you could access side-channel data in your custom encoder or decoder?

```
    Map<Class<?>, Object> tags;
```

LachlanMcKee:
If that's something that could be added into the library, I'm sure I could make it work =D.

so it could look something like (kotlin): 

```
data class UnknownEnumValue(val path: String, val unknownValue: String)
```

Customer Adapter
```
...
val existingUnknownValues = reader.getTag(CustomEnumJsonAdapter::class) as? MutableList<UnknownEnumValue>
if (existingUnknownValues != null) {
     existingUnknownValues.add(UnknownEnumValue("path", "value"))
} else {
     val unknownValuesList = mutableListOf(UnknownEnumValue("path", "value"))
     reader.setTag(CustomEnumJsonAdapter::class, unknownValuesList)
}
...
```

```
val reader = JsonReader.of(...)
EnumJsonAdapter.create(Roshambo.class).fromJson(reader)
val existingUnknownValues : List<UnknownEnumValue> = reader.getTag(CustomEnumJsonAdapter::class)

// Log events
```

It would be nice for the reader to have (but obviously in Java):

```
fun getTag(clazz: Class<*>): Any
fun getTags(): Map<Class<*>, Any>
fun setTag(clazz: Class<*>, value: Any)
```

LachlanMcKee:
@swankjesse I have raised a new PR that uses the tag approach that you suggested: https://github.com/square/moshi/pull/1227. Happy to make any adjustments you think are required =]

swankjesse:
Obsoleted by the work on tags.

PR #1210, opened 2020-08-30
Gradle wrapper validation

sullis:
https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action


ZacSweers:
There is already a PR for this #1024

JakeWharton:
Also it needs to run before we invoke the binary, but this will run concurrently meaning a vulnerable version will race.

PR #1213, opened 2020-09-04
Fix codegen compilation.

Tolriq:
- Fix ShadowJar configuration to properly update source code to point to relocated package
- Update Shadow plugin to 6.0.0

### DIFF
--- a/kotlin/codegen/build.gradle.kts
+++ b/kotlin/codegen/build.gradle.kts
@@ -22,7 +22,7 @@ plugins {
   kotlin("jvm")
   kotlin("kapt")
   id("com.vanniktech.maven.publish")
-  id("com.github.johnrengelman.shadow") version "5.2.0"
+  id("com.github.johnrengelman.shadow") version "6.0.0"
 }
 
 tasks.withType<KotlinCompile>().configureEach {
@@ -109,3 +109,5 @@ afterEvaluate {
     }
   }
 }
+
+tasks.named("assemble") { dependsOn(tasks.named("shadowJar")) }


kotlin/codegen/build.gradle.kts
```diff
@@ -44,6 +44,7 @@ dependencies {
     exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
   }
   implementation(Dependencies.KotlinPoet.kotlinPoet)
+  kapt(Dependencies.KotlinPoet.kotlinPoet)
```
JakeWharton, line 47: Why is this needed? The only processors on the processorpath are auto service and incap helper.

Tolriq: Actually seems it's not needed. As said in both issues this is not a part I usually use and master.
During first attempt adding this did embed kotlinpoet in the jar so I assumed it was necessary and part of the fix.

Just tested and just properly binding the shadowJar task at proper timing is enough. PR updated.

kotlin/codegen/build.gradle.kts
```diff
@@ -109,3 +109,5 @@ afterEvaluate {
     }
   }
 }
+
+tasks.named("assemble") { dependsOn(tasks.named("shadowJar")) }
```
ZacSweers, line 113: Why is this needed? shadowJar is already configured above this for publications

```
> Task :kotlin:codegen:jar
> Task :kotlin:codegen:shadowJar
> Task :kotlin:tests:kaptGenerateStubsTestKotlin
> Task :kotlin:tests:kaptTestKotlin
```

shadowed kotlinpoet/kotlinx-metadata classes are present too 

image

The problem here appears to be that kotlinpoet itself is getting shaded

image

I tried with your changes, and the same issue is still present

image

The proper fix here is to figure out why this block is excluding kotlinpoet's core artifact from the pom and shading it. This PR just updates the shadow plugin dependency but doesn't fix anything. We probably missed this in the switch to gradle because the kotlin gradle plugin often just uses compiled class files from their local project's build dir directly rather than packaged jars (sometimes annoying quirk of how their incremental compilation works).

image

One quick fix is to add `exclude(group = "com.squareup", module = "kotlinpoet")` to the shaded kotlinpoet deps to ensure it doesn't shade the core artifact, but still leaves the `runtime` issue. This appears to be a long-standing issue in the shadow plugin: https://github.com/johnrengelman/shadow/issues/321. We can fix this with a `withXml` adjustment. I've opened #1214 to do these

Tolriq: This is needed because it ran too late and the source code was not updated.
So the code still referenced the non shadowed one and failed.

Issue #1147, opened 2020-05-26
Question: Enum, codegen and proguard


glureau-betclic:
Trying to use an enum with Moshi-kotlin-codegen (1.9.2).

    enum class LevelDto {
        @Json(name = "low") low,
        @Json(name = "high") high
    }

This version is not safe for proguard, but I'm not sure if I should add a `@Keep` (or a proguard rule to preserve all enums) or a `@JsonAdapter(generateAdapter = false)`. The only "documentation" that I found points to use the latter, but it's 2 years old so I'd have a confirmation if possible.

https://github.com/square/moshi/blob/0c85eae34af00ecbee46beaa5b25fb4af00fb9f2/moshi/src/main/resources/META-INF/proguard/moshi.pro#L12

If it's still right, maybe editing the README.md could help future users?

ZacSweers:
Feel free to send a PR to the README!

PR #1216, opened 2020-09-09
Update README to define Codegen enum best practice

glureau-betclic:
https://github.com/square/moshi/issues/1147

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,16 @@ data class BlackjackHand(
 )
 ```
 
+or for the enum classes:
+
+```kotlin
+@JsonClass(generateAdapter = false)
+enum class Suit {
+  CLUBS, DIAMONDS, HEARTS, SPADES;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -557,6 +557,16 @@ data class BlackjackHand(
 )
 ```
 
+or for the enum classes:
```
ZacSweers, line 560: Can you add more here? This is vague on its own and should actually go on the proguard rules section

glureau-betclic: I'm afraid I don't have enough knowledge to be sure of what it changes compared to a simple @Keep, I had the feeling it was more than just proguard.

ZacSweers: I mean just make this a full sentence and put it in the correct proguard section of the readme

ZacSweers:
Closing for now as this has gone stale

PR #1219, opened 2020-09-11
Implement JsonReader#valueSource()

### LINKED ISSUE #675: JsonReader.readJsonString

ZacSweers:
This is a reader analog for #943 and resolves #675. The tests need a bit more fleshing out, but want to start with this for review.

At a high level
- JsonUtf8Reader works by basically replacing `skipValue()` with one that optionally accepts a `Sink`. `skipValue()` calls this method with a black hole, and `valueSource()` calls this with a `Buffer` that the returned `BufferedSource` ultimately consumes. A lot of `skip()`/`readByte()` calls are reworked to defer to this sink, so blackhole behavior just skips under the hood like before and valueSource() keeps the data in its buffer. I don't know if there's a lazier way to push this along though, as right now it eagerly reads the value like skipValue does.
- JsonValueReader works by just emitting utf8 values directly to a buffer that the returned source consumes. One interesting caveat here is that when creating value maps, it always uses doubles for numbers, which could create asymmetric inputs/outputs.

There's probably a couple places these need to check scope in other methods, I'm in unfamiliar territory with this PR so feedback welcome


moshi/src/test/java/com/squareup/moshi/JsonStringQualifierRecipe.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.reflect.Type;
+import java.util.Set;
+import okio.BufferedSink;
+import okio.BufferedSource;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public final class JsonStringQualifierRecipe {
```
ZacSweers, line 15: I initially wrote this before deciding to move it to the recipe subproject instead. Weirdly though, _this_ version works but the recipe version in Kotlin does not, despite appearing to have identical implementations. The recipe version always omits the leading curly brace 🤔

moshi/src/test/java/com/squareup/moshi/ValueSourceEscapesTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.Okio;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public final class ValueSourceEscapesTest {
```
ZacSweers, line 13: This is adapted from https://github.com/infotech-group/moshi/commit/1d052e71e163b99a6f4b233f951edf9b71f646f8, but I wasn't able to get `escapedSingleQuote()` and `backslash1` working. Feedback welcome

moshi/src/test/java/com/squareup/moshi/ValueSourceEscapesTest.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private void testSpecialCharacter(String value) throws IOException {
+    String input = "{\"a\":\"" + value + "\"}";
+    JsonReader reader = JsonReader.of(Okio.buffer(Okio.source(new ByteArrayInputStream(input.getBytes()))));
+    Buffer out = new Buffer();
+    JsonWriter writer = JsonWriter.of(out);
+
+    reader.beginObject();
+    writer.beginObject();
```
ZacSweers: @swankjesse I wonder if this could ultimately be a solution to your previously mentioned desire to have a reader/writer symmetry where the write transforms the json going into the reader and can omit/modify certain keys/values

recipes/src/main/java/com/squareup/moshi/recipes/JsonString.kt
```diff
@@ -0,0 +1,56 @@
...
+  val moshi = Builder()
+    .add(JsonStringJsonAdapter.Factory)
+    .build()
+
+  val example: ExampleClass = moshi.adapter(ExampleClass::class.java).fromJson(json)!!
+
+  check(example.type == 1)
+
+  //language=JSON
+  check(example.rawJson == "{\"a\":2,\"b\":3,\"c\":[1,2,3]}")
```
ZacSweers, line 55: As mentioned above - this inexplicably fails due to it missing the leading curly brace 🤔

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -432,6 +440,15 @@ public final boolean failOnUnknown() {
    */
   public abstract void skipValue() throws IOException;
 
+  /**
+   * Reads the next value recursively into a {@link BufferedSource}. If it is an object or array,
+   * all nested elements are read into the source as well. This method is intended for use when the
+   * JSON token stream contains custom data that the caller wants to read bytes directly off of the
+   * underlying buffer. You <b>must</b> call {@link BufferedSource#close()} before interacting with
+   * this {@code JsonReader} instance again.
```
swankjesse, line 448: Nice docs!

swankjesse (commented):
I’d prefer a brace-matching implementation. Did you attempt that? What’s the drawback?

moshi/src/main/java/com/squareup/moshi/JsonValueReader.java
```diff
@@ -315,6 +321,84 @@ public int nextInt() throws IOException {
     return result;
   }
 
+  @Override
+  public BufferedSource valueSource() {
+    Object peeked = stackSize != 0 ? stack[stackSize - 1] : null;
+    pushScope(STREAMING_VALUE);
+    final Buffer buffer = new Buffer();
+    writeValueToBuffer(peeked, buffer);
```
swankjesse, line 329: This doesn’t stream; it buffers the entire contents

ZacSweers:
At the time no, the idea was just to make skipValue read the bytes it skips into a sink and return that. A simple braces check would work I suppose, but we'd have to peek for a number/Boolean/string too right?

ZacSweers:
Superseded by #1223

PR #1222, opened 2020-09-12
JsonValueSource: a brace-matching implementation of okio.Source

swankjesse:
This will read one JSON object and then stop.

This is a part of https://github.com/square/moshi/issues/675

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
@@ -1,0 +1,208 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi;
+
+import java.io.EOFException;
+import java.io.IOException;
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.ByteString;
+import okio.Source;
+import okio.Timeout;
+
+/**
+ * This source reads a prefix of another source as a JSON value and then terminates. It can read
+ * top-level arrays, objects, or strings only.
+ *
+ * <p>It implements {@linkplain JsonReader#setLenient lenient parsing} and has no mechanism to
+ * enforce strict parsing. If the input is not valid or lenient JSON the behavior of this source is
+ * unspecified.
+ */
+final class JsonValueSource implements Source {
+  private static final ByteString STATE_JSON = ByteString.encodeUtf8("[]{}\"'/#");
+  private static final ByteString STATE_SINGLE_QUOTED = ByteString.encodeUtf8("'\\");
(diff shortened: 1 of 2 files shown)


ZacSweers:
Dope. Would this be usable for a fast `skipValue` too?

swankjesse:
Yep.

ZacSweers (approved):
Implementation looks good to me, but I'm admittedly out of my depth in this space so definitely think another reviewer would be good

moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
```diff
@@ -0,0 +1,208 @@
...
+import okio.BufferedSource;
+import okio.ByteString;
+import okio.Source;
+import okio.Timeout;
+
+/**
+ * This source reads a prefix of another source as a JSON value and then terminates. It can read
+ * top-level arrays, objects, or strings only.
+ *
+ * <p>It implements {@link JsonReader#setLenient lenient parsing} and has no mechanism to enforce
```
JakeWharton, line 30:
```suggestion
 * <p>It implements {@linkplain JsonReader#setLenient lenient parsing} and has no mechanism to enforce
```

moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /**
+   * Discards any remaining JSON data in this source that was left behind after it was closed. It is
+   * an error to call {@link #read} after calling this method.
+   */
+  public void discard() throws IOException {
+    closed = true;
+    while (state != STATE_END_OF_JSON) {
+      advanceLimit(8192);
```
JakeWharton, line 175: Unfortunate that we have this constant in multiple codebases downstream of Okio. I'm not sure how it would even be exposed if we wanted... `Buffer.defaultSegmentSize()`?

moshi/src/test/java/com/squareup/moshi/JsonValueSourceTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+import java.io.EOFException;
+import java.io.IOException;
+import okio.Buffer;
+import org.junit.Test;
+
+public final class JsonValueSourceTest {
+  @Test
+  public void simpleValues() throws IOException {
+    assertThat(jsonPrefix("{\"hello\": \"world\"}, 1, 2, 3")).isEqualTo("{\"hello\": \"world\"}");
```
JakeWharton, line 29: What do you think about making this function `splitJson` which returns `List<String>` so that we can assert the code does not consume too many bytes. This specific example would change to
```java
assertThat(splitJson("..")).containsExactly("json", ", 1, 2, 3");
```

moshi/src/test/java/com/squareup/moshi/JsonValueSourceTest.java
```diff
@@ -0,0 +1,204 @@
...
+  private String jsonPrefix(String string) throws IOException {
+    Buffer allData = new Buffer();
+    allData.writeUtf8(string);
+
+    Buffer jsonPrefixBuffer = new Buffer();
+    jsonPrefixBuffer.writeAll(new JsonValueSource(allData));
+
+    String result = jsonPrefixBuffer.readUtf8();
+    String remainder = allData.readUtf8();
+    assertThat(result + remainder).isEqualTo(string);
```
JakeWharton, line 200: Ah I see you're implicitly testing that behavior here. I'm satisfied.

JakeWharton (approved):

moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
swankjesse: Fixed!

moshi/src/main/java/com/squareup/moshi/JsonValueSource.java
swankjesse: Yeah, agreed.

PR #1223, opened 2020-09-13
WIP: Use JsonValueSource

swankjesse:


moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -432,6 +432,14 @@ public final boolean failOnUnknown() {
    */
   public abstract void skipValue() throws IOException;
 
+  /**
+   * Reads the next value recursively into a {@link BufferedSource}. If it is an object or array,
```
JakeWharton, line 436: I think I would drop "recursively" here. The next sentence already explains the behavior in a way that is more clear.

tmm1:
How come this was closed?

ZacSweers:
Because it was implemented in another PR.

PR #1224, opened 2020-09-13
New public API, JsonReader.nextSource()

swankjesse:
The streaming side of https://github.com/square/moshi/issues/675

We can still add a nicer wrapper on top to read the JSON value as
a string.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -22,6 +22,7 @@ import javax.annotation.Nullable;
 import okio.Buffer;
 import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
 
 final class JsonUtf8Reader extends JsonReader {
   private static final long MIN_INCOMPLETE_INTEGER = Long.MIN_VALUE / 10;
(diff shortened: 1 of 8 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -1013,6 +1025,55 @@ public void skipValue() throws IOException {
...
+  public BufferedSource nextSource() throws IOException {
+    int p = peeked;
+    if (p == PEEKED_NONE) {
+      p = doPeek();
+    }
+
+    int valueSourceStackSize = 0;
+    Buffer prefix = new Buffer();
+    ByteString state = JsonValueSource.STATE_END_OF_JSON;
+    if (p == PEEKED_BEGIN_ARRAY) {
```
JakeWharton, line 1038: Why is this logic needed here rather than being part of the value source? I would have thought it would start in a START_DOCUMENT-like state and handle all these cases just like it would if it was parsing a nested value.

JakeWharton (approved):

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
swankjesse: Unfortunately Moshi reads too much!

Suppose you do this:
```
if (reader.peek() != JsonReader.Token.NULL) {
 ...
}
```

Moshi will look at the next character. If it’s `[` it’ll consume it and set the peeked state to `PEEKED_BEGIN_ARRAY`. There’s no putting that character back! Or at least, not without a lot of rework of other Moshi internals.

Another example – if you do `reader.select(options)`, Moshi will first attempt to do an efficient Okio select. But if that doesn’t match one of the expected strings, it’ll consume the string! This is necessary so that "p\u0065psi" is matched on a select for ["pepsi", "coke"]. And when we wanna read a value, we put the `"pepsi"` string back.

PR #1225, opened 2020-09-15
Small improvements to JsonReader.nextSource

swankjesse:
Defer doing I/O until strictly necessary.

Remove some unnecessary branching in doPeek(), which is in the
inner loop of all JSON parsing.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
@@ -241,10 +241,6 @@ final class JsonUtf8Reader extends JsonReader {
   }
 
   private int doPeek() throws IOException {
-    if (valueSource != null) {
-      valueSource.discard();
-      valueSource = null;
-    }
     int peekStack = scopes[stackSize - 1];
     if (peekStack == JsonScope.EMPTY_ARRAY) {
       scopes[stackSize - 1] = JsonScope.NONEMPTY_ARRAY;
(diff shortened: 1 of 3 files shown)


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -241,10 +241,6 @@ public Token peek() throws IOException {
   }
 
   private int doPeek() throws IOException {
-    if (valueSource != null) {
-      valueSource.discard();
-      valueSource = null;
-    }
```
swankjesse, line 247: I didn’t like how we were evaluating the state of valueSource on like, EVERY read.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -329,6 +325,13 @@ private int doPeek() throws IOException {
       } else {
         checkLenient();
       }
+    } else if (peekStack == JsonScope.STREAMING_VALUE) {
+      valueSource.discard();
+      valueSource = null;
+      stackSize--;
+      pathIndices[stackSize - 1]++;
+      pathNames[stackSize - 1] = "null";
+      return doPeek();
```
swankjesse, line 334: ... now it should happen less frequently. I accomplish this by using the stack to hint that a value source is non-null.

JakeWharton (approved):

PR #1227, opened 2020-09-15
Introduced tags to reader/writers

LachlanMcKee:
Relates to issue #1126.

This is an implementation of a suggestion from @swankjesse (https://github.com/square/moshi/pull/1209#issuecomment-683903925)

This will give developers the ability to append custom data to JsonReader/JsonWriter classes. This will be very useful for custom adapters that wish to track supplemental data.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -573,16 +573,26 @@ public abstract class JsonReader implements Closeable {
   }
 
   /** Returns the tag value for the given class key. */
+  @SuppressWarnings("unchecked")
   @CheckReturnValue
-  public final @Nullable Object getTag(Class<?> clazz) {
+  public final @Nullable <T> T tag(Class<? extends T> clazz) {
     if (tags == null) {
       return null;
     }
-    return tags.get(clazz);
+    return (T) tags.get(clazz);
   }
 
   /** Assigns the tag value using the given class key and value. */
-  public final void setTag(Class<?> clazz, Object value) {
+  public final <T> void setTag(Class<? extends T> clazz, T value) {
+    if (!clazz.isInstance(value)) {
+      throw new IllegalArgumentException(
+          "Tag value '"
+              + value
+              + "' ("
+              + value.getClass().getName()
+              + ") is not of type "
+              + clazz.getName());
+    }
     if (tags == null) {
       tags = new HashMap<>();
     }
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -584,14 +584,8 @@ public abstract class JsonReader implements Closeable {
 
   /** Assigns the tag value using the given class key and value. */
   public final <T> void setTag(Class<? extends T> clazz, T value) {
-    if (!clazz.isInstance(value)) {
-      throw new IllegalArgumentException(
-          "Tag value '"
-              + value
-              + "' ("
-              + value.getClass().getName()
-              + ") is not of type "
-              + clazz.getName());
+    if (clazz != value.getClass()) {
+      throw new IllegalArgumentException("Tag value must be of type " + clazz.getName());
     }
     if (tags == null) {
       tags = new HashMap<>();
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -584,7 +584,7 @@ public abstract class JsonReader implements Closeable {
 
   /** Assigns the tag value using the given class key and value. */
   public final <T> void setTag(Class<T> clazz, T value) {
-    if (clazz != value.getClass()) {
+    if (!clazz.isAssignableFrom(value.getClass())) {
       throw new IllegalArgumentException("Tag value must be of type " + clazz.getName());
     }
     if (tags == null) {
--- a/moshi/src/main/java/com/squareup/moshi/JsonReader.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonReader.java
@@ -21,6 +21,8 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.CheckReturnValue;
@@ -196,6 +198,8 @@ public abstract class JsonReader implements Closeable {
 
   /** True to throw a {@link JsonDataException} on any attempt to call {@link #skipValue()}. */
   boolean failOnUnknown;
+
+  private Map<Class<?>, Object> tags;
 
   /** Returns a new instance that reads UTF-8 encoded JSON from {@code source}. */
   @CheckReturnValue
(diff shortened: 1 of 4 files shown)


LachlanMcKee:
An example usecase for this can be found here: https://github.com/square/moshi/pull/1228. (Note this example is not intended to be merged, it's just to demonstrate)

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -569,6 +573,32 @@ public final String getPath() {
...
+  public final @Nullable Object getTag(Class<?> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return tags.get(clazz);
+  }
+
+  /** Returns all tag values as an unmodifiable list. */
+  @CheckReturnValue
+  public final Map<Class<?>, Object> getTags() {
```
swankjesse, line 587: Drop this method please, if we expose it then adding a tag becomes an implicit contract that other code can rely upon.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -561,4 +565,30 @@ public final void endFlatten(int token) {
...
+  public final @Nullable Object getTag(Class<?> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return tags.get(clazz);
+  }
+
+  /** Returns all tag values as an unmodifiable list. */
+  @CheckReturnValue
+  public final Map<Class<?>, Object> getTags() {
```
swankjesse, line 580: same!

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
```diff
@@ -1433,6 +1434,24 @@ public void nextSourceNotConsumed() throws IOException {
     assertThat(reader.nextString()).isEqualTo("d");
   }
 
+  @Test
+  public void tags() throws IOException {
+    JsonReader reader = newReader("{}");
+    assertThat(reader.getTag(String.class)).isNull();
+    assertThat(reader.getTag(Integer.class)).isNull();
```
swankjesse, line 1441: One extra constraint – we should require that the tag value is an instance of the tag’s type. In practice you’d create a type for ever tag you set.

swankjesse (commented):
I like where you’re going with this. I think the signature of the method to set a tag should be this:
```
 <T> void setTag(Class<? extends T> type, T tag)
```

And the function to get it should omit the get prefix:
```
 <T> @Nullable T tag(Class<? extends T> type);
```

moshi/src/main/java/com/squareup/moshi/JsonReader.java
LachlanMcKee: removed.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
LachlanMcKee: removed.

moshi/src/test/java/com/squareup/moshi/JsonReaderTest.java
LachlanMcKee: I've added the constraint in the test, as well as adding an exception in the `setTag` method in case someone attempts to assign a wrong value.

LachlanMcKee:
@swankjesse I have pushed changed with the suggestions that you made.

I wasn't sure if you wanted me to create tag classes within the tests rather than using String/Integer, so I have left that as is for now.

LachlanMcKee:
I've made a slight tweak to avoid checking instanceOf and check that the tag class and the value class are the same.

I also added some simple tag classes (one using inheritance). I miss data classes haha

LachlanMcKee:
I've updated the example PR as well to demonstrate

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -569,6 +572,27 @@ public final String getPath() {
...
+    return (T) tags.get(clazz);
+  }
+
+  /** Assigns the tag value using the given class key and value. */
+  public final <T> void setTag(Class<T> clazz, T value) {
+    if (clazz != value.getClass()) {
+      throw new IllegalArgumentException("Tag value must be of type " + clazz.getName());
+    }
+    if (tags == null) {
+      tags = new HashMap<>();
```
JakeWharton, line 591: LinkedHashMap please

https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -569,6 +572,27 @@ public final String getPath() {
...
+  public final @Nullable <T> T tag(Class<T> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return (T) tags.get(clazz);
+  }
+
+  /** Assigns the tag value using the given class key and value. */
+  public final <T> void setTag(Class<T> clazz, T value) {
+    if (clazz != value.getClass()) {
```
JakeWharton, line 587: This means I can't use polymorphism. For example, I couldn't assign a String to the CharSequence tag. Should be an isAssignable check.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -561,4 +564,25 @@ public final void endFlatten(int token) {
...
+    return (T) tags.get(clazz);
+  }
+
+  /** Assigns the tag value using the given class key and value. */
+  public final <T> void setTag(Class<T> clazz, T value) {
+    if (clazz != value.getClass()) {
+      throw new IllegalArgumentException("Tag value must be of type " + clazz.getName());
+    }
+    if (tags == null) {
+      tags = new HashMap<>();
```
JakeWharton, line 584: LinkedHashMap

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -561,4 +564,25 @@ public final void endFlatten(int token) {
...
+  public final @Nullable <T> T tag(Class<T> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return (T) tags.get(clazz);
+  }
+
+  /** Assigns the tag value using the given class key and value. */
+  public final <T> void setTag(Class<T> clazz, T value) {
+    if (clazz != value.getClass()) {
```
JakeWharton, line 580: Same as above

JakeWharton (approved):
You can use erasure to test assignability by casting String.class to Class<Object> and then using an integer or object for the value.

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
LachlanMcKee: done

moshi/src/main/java/com/squareup/moshi/JsonReader.java
LachlanMcKee: done

moshi/src/main/java/com/squareup/moshi/JsonReader.java
LachlanMcKee: seems reasonable. I had been on the fence about this, but I have made the change =]

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
LachlanMcKee: done

LachlanMcKee:
@JakeWharton I have made your proposed changes. Thanks for the feedback =]

moshi/src/main/java/com/squareup/moshi/JsonReader.java
```diff
@@ -569,6 +572,27 @@ public final String getPath() {
...
+  public final @Nullable <T> T tag(Class<T> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return (T) tags.get(clazz);
+  }
+
+  /** Assigns the tag value using the given class key and value. */
+  public final <T> void setTag(Class<T> clazz, T value) {
+    if (!clazz.isAssignableFrom(value.getClass())) {
```
swankjesse, line 587: nit: I think we want `clazz.isInstance(value)`

swankjesse (approved):
Awesome. Thanks for driving this.

LachlanMcKee:
Thanks both =)

PR #1228, opened 2020-09-15
EXAMPLE - Not for review - Moshi tags proposal example

LachlanMcKee:
This is an example implementation of the tags change proposed here: https://github.com/square/moshi/pull/1227

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/JsonWriter.java
+++ b/moshi/src/main/java/com/squareup/moshi/JsonWriter.java
@@ -577,7 +577,7 @@ public abstract class JsonWriter implements Closeable, Flushable {
 
   /** Assigns the tag value using the given class key and value. */
   public final <T> void setTag(Class<T> clazz, T value) {
-    if (clazz != value.getClass()) {
+    if (!clazz.isAssignableFrom(value.getClass())) {
       throw new IllegalArgumentException("Tag value must be of type " + clazz.getName());
     }
     if (tags == null) {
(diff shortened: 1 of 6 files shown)


moshi/src/main/java/com/squareup/moshi/JsonWriter.java
```diff
@@ -561,4 +565,30 @@ public final void endFlatten(int token) {
...
+  /** Returns the tag value for the given class key. */
+  @CheckReturnValue
+  public final @Nullable Object getTag(Class<?> clazz) {
+    if (tags == null) {
+      return null;
+    }
+    return tags.get(clazz);
+  }
+
+  /** Returns all tag values as an unmodifiable list. */
```
swankjesse, line 578: I don’t think we want this method. You should need to know about a tag to see it! Otherwise tags become externally observable behavior

swankjesse (commented):
I like the features on JsonReader and JsonWriter. I argue that for UnknownEnumValue we’d make that something application-layer developers would need to do. But the tag mechanism would give them a way to pass information learned during parsing back to the caller.

LachlanMcKee:
@swankjesse I completely agree that this shouldn't be added to the library, I just wanted to demonstrate an example in a separate PR to avoid preventing the other pr being merged

moshi/src/main/java/com/squareup/moshi/JsonWriter.java
LachlanMcKee: sounds good. I'll remove it shortly.

LachlanMcKee:
Closing this as the PR this is related to has been merged.

PR #1234, opened 2020-09-20
Use nextSource().readAll(blackhole()) for fast skip

ZacSweers:


ZacSweers:
Currently fails a number of tests for me locally, which seems surprising

cketti:
The following snippet uses the existing skip behavior for simple values and only defers to `nextSource()` for objects and arrays. Passes all tests on my machine.

```java
  @Override
  public void skipValue() throws IOException {
    if (failOnUnknown) {
      throw new JsonDataException("Cannot skip unexpected " + peek() + " at " + getPath());
    }
    int p = peeked;
    if (p == PEEKED_NONE) {
      p = doPeek();
    }

    if (p == PEEKED_BEGIN_ARRAY || p == PEEKED_BEGIN_OBJECT) {
      nextSource().readAll(Okio.blackhole());
      valueSource.discard();
      valueSource = null;
      stackSize--;
    } else if (p == PEEKED_UNQUOTED_NAME || p == PEEKED_UNQUOTED) {
      skipUnquotedValue();
    } else if (p == PEEKED_DOUBLE_QUOTED || p == PEEKED_DOUBLE_QUOTED_NAME) {
      skipQuotedValue(DOUBLE_QUOTE_OR_SLASH);
    } else if (p == PEEKED_SINGLE_QUOTED || p == PEEKED_SINGLE_QUOTED_NAME) {
      skipQuotedValue(SINGLE_QUOTE_OR_SLASH);
    } else if (p == PEEKED_NUMBER) {
      buffer.skip(peekedNumberLength);
    } else if (p == PEEKED_END_ARRAY || p == PEEKED_END_OBJECT || p == PEEKED_EOF) {
      throw new JsonDataException("Expected a value but was " + peek() + " at path " + getPath());
    }

    peeked = PEEKED_NONE;
    pathIndices[stackSize - 1]++;
    pathNames[stackSize - 1] = "null";
  }
```

Theschorpioen (approved):

ZacSweers:
@cketti I wasn't able to get all tests to pass with that but maybe in the future I'll look again

PR #1236, opened 2020-09-22
Fix problems discovered using nextSource() to implement skipValue

swankjesse:


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -564,6 +562,7 @@ public String nextName() throws IOException {
       result = nextQuotedValue(SINGLE_QUOTE_OR_SLASH);
     } else if (p == PEEKED_BUFFERED_NAME) {
       result = peekedString;
+      peekedString = null;
```
swankjesse, line 565: unrelated bug fix. No visible effects, but it was wrong to retain this value when it wasn’t reachable.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java
```diff
@@ -1069,9 +1068,14 @@ public BufferedSource nextSource() throws IOException {
       throw new JsonDataException("Expected a value but was " + peek() + " at path " + getPath());
     }
 
+    // Advance the path and clear peeked if we haven't already.
+    if (peeked != PEEKED_NONE) {
```
swankjesse, line 1072: this check is necessary to avoid double-incrementing the position in cases like unquoted strings.

swankjesse (commented):
This passes tests with an alternate `skip()` method.

```
  @Override
  public void skipValue() throws IOException {
    if (failOnUnknown) {
      throw new JsonDataException("Cannot skip unexpected " + peek() + " at " + getPath());
    }
    if (peek() == Token.NAME) {
      skipName();
    } else {
      nextSource();

      // If this element is in an object clear out the key.
      if (stackSize > 1) {
        pathNames[stackSize - 2] = "null";
      }
    }
  }
```

ZacSweers (approved):

PR #1238, opened 2020-09-25
Prepare dokka

ZacSweers:
This updates to Dokka 1.4.10 and prepares for a future with Kotlin rewrite and mkdocs

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,6 +15,8 @@
  */
 
 import org.gradle.jvm.tasks.Jar
+import org.jetbrains.dokka.gradle.DokkaTask
+import java.net.URL
 
 buildscript {
   dependencies {
@@ -23,8 +25,8 @@ buildscript {
 }
 
 plugins {
-  id("com.vanniktech.maven.publish") version "0.12.0" apply false
-  id("org.jetbrains.dokka") version "0.10.1" apply false
+  id("com.vanniktech.maven.publish") version "0.13.0" apply false
+  id("org.jetbrains.dokka") version "1.4.10" apply false
   id("com.diffplug.spotless") version "5.2.0"
 }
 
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -99,5 +95,18 @@ subprojects {
...
     }
+
+    if (name != "codegen" && pluginManager.hasPlugin("org.jetbrains.kotlin.jvm")) {
+      apply(plugin = "org.jetbrains.dokka")
+      tasks.named<DokkaTask>("dokkaHtml") {
+        outputDirectory.set(rootDir.resolve("docs/1.x"))
+        dokkaSourceSets.configureEach {
+          skipDeprecated.set(true)
+          externalDocumentationLink {
+            url.set(URL("https://square.github.io/okio/2.x/okio/"))
```
ZacSweers, line 106: Note that 1.x isn't an option because the javadoc is broken https://square.github.io/okio/1.x/okio/

swankjesse: Yikes. https://github.com/square/okio/issues/775#issuecomment-699266366

swankjesse (approved):
Presumably we won’t run Dokka on Moshi 1.x?

ZacSweers:
We do already today, just with an older version and only for the moshi-kotlin and code gen artifacts. This HTML option is more useful for a project site

build.gradle.kts
```diff
@@ -24,9 +26,9 @@ buildscript {
 }
 
 plugins {
-  id("com.vanniktech.maven.publish") version "0.12.0" apply false
-  id("org.jetbrains.dokka") version "0.10.1" apply false
-  id("com.diffplug.spotless") version "5.5.0"
+  id("com.vanniktech.maven.publish") version "0.13.0" apply false
+  id("org.jetbrains.dokka") version "1.4.10" apply false
+  id("com.diffplug.spotless") version "5.6.0"
```
ZacSweers, line 31: Opportunistically upped this

swankjesse (approved):

PR #1286, opened 2021-01-09
Improve KotlinJsonAdapter performance by invoke KFunction by "call".

k163377:
This is my first time posting pull request and I am not good at English, so I'm sorry if there is something wrong with it.

## Overview
If the arguments are fully specified (= don't have to use `Default arguments`), can speed up by making a invoke with `KFunction.call`.

## Description
Currently `KotlinJsonAdapter` does mapping by calling` KFunction.callBy`.
On the other hand, [`KFunction.callBy` has about 1/5 the performance of` KFunction.call`](https://github.com/ProjectMapK/FastKFunction#how-fast), so you can expect speedup by doing `KFunction.call` as much as possible.

I tried to achieve this by making `IndexedParameterMap` manage the initialization state.

### Base branch
Which `branch` is the appropriate destination for a `pull request`?
I'm expecting `moshi_1.11.x`, but I'm tentatively specifying `master`.

### Benchmarking
I'm thinking of making a benchmark to show the performance improvement with this method.
If you don't mind, could you point me to a suitable existing benchmark code?

I was thinking of using [this pull request](https://github.com/square/moshi/pull/773) or [this repository](https://github.com/ZacSweers/json-serialization-benchmarking) as a basis.

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -94,21 +94,27 @@ internal class KotlinJsonAdapter<T>(
     reader.endObject()
 
     // Confirm all parameters are present, optional, or nullable.
+    var isFullInitialized = true
     for (i in 0 until constructorSize) {
-      if (values[i] === ABSENT_VALUE && !constructor.parameters[i].isOptional) {
-        if (!constructor.parameters[i].type.isMarkedNullable) {
-          throw Util.missingProperty(
+      if (values[i] === ABSENT_VALUE) {
+        when {
+          constructor.parameters[i].isOptional -> isFullInitialized = false
+          constructor.parameters[i].type.isMarkedNullable -> values[i] = null // Replace absent with null.
+          else -> throw Util.missingProperty(
             constructor.parameters[i].name,
             allBindings[i]?.jsonName,
             reader
           )
         }
-        values[i] = null // Replace absent with null.
       }
     }
 
     // Call the constructor using a Map so that absent optionals get defaults.
-    val result = constructor.callBy(IndexedParameterMap(constructor.parameters, values))
+    val result = if (isFullInitialized) {
+      constructor.call(*values)
+    } else {
+      constructor.callBy(IndexedParameterMap(constructor.parameters, values))
+    }
 
     // Set remaining properties.
     for (i in constructorSize until allBindings.size) {
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -94,7 +94,7 @@ internal class KotlinJsonAdapter<T>(
     reader.endObject()
 
     // Confirm all parameters are present, optional, or nullable.
-    var isFullInitialized = true
+    var isFullInitialized = allBindings.size == constructorSize
     for (i in 0 until constructorSize) {
       if (values[i] === ABSENT_VALUE) {
         when {


ZacSweers:
While neat, I'm not sure if this is worth the added complexity. Speed improvement is nice ultimately if someone cares about speed they should use code generation

k163377:
@ZacSweers 
I agree that if the speed improvement in this code is not important, the value of incorporating complex changes is small.

If you want to use `KFunction.call` without compromising simplicity as much as possible, you could modify only about 20 lines from line 97 of `KotlinJsonAdapter` to look like this.

```kotlin
    var isFullInitialized = true
      if (values[i] === ABSENT_VALUE) {
        if (constructor.parameters[i].isOptional) {
          isFullInitialized = false
        } else {
          if (constructor.parameters[i].type.isMarkedNullable) {
            throw Util.missingProperty(
              constructor.parameters[i].name,
              allBindings[i]?.jsonName,
              reader
            )
          }
          values[i] = null // Replace absent with null.
        }
      }
    }

    // Call the constructor using a Map so that absent optionals get defaults.
    val result = if (isFullInitialized)
      constructor.call(*values)
    else
      constructor.callBy(IndexedParameterMap(constructor.parameters, values))
```

However, I think the value of incorporating it is small even with this content.
If you reply that this change is unnecessary, I will close this pull request.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/SpreadWrapper.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.moshi.kotlin.reflect;
+
+import kotlin.reflect.KFunction;
+
+class SpreadWrapper {
```
ZacSweers, line 20:
```suggestion
internal class SpreadWrapper {
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -108,7 +107,10 @@ internal class KotlinJsonAdapter<T>(
     }
 
     // Call the constructor using a Map so that absent optionals get defaults.
-    val result = constructor.callBy(IndexedParameterMap(constructor.parameters, values))
+    val result = if (values.isFullInitialized())
+      SpreadWrapper.call(constructor, values.parameterValues)
+    else
+      constructor.callBy(values)
```
ZacSweers, line 113: Please match our style

```suggestion
    val result = if (values.isFullInitialized()) {
      SpreadWrapper.call(constructor, values.parameterValues)
    } else {
      constructor.callBy(values)
    }
```

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -95,7 +94,7 @@ internal class KotlinJsonAdapter<T>(
 
     // Confirm all parameters are present, optional, or nullable.
     for (i in 0 until constructorSize) {
-      if (values[i] === ABSENT_VALUE && !constructor.parameters[i].isOptional) {
+      if (values.containsKey(i) && !constructor.parameters[i].isOptional) {
```
ZacSweers, line 97: Let's save off the parameter and wrap it in a null-check instead, rather than checking that it contains the key and then looking it up two more times

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
```diff
@@ -154,28 +156,47 @@ internal class KotlinJsonAdapter<T>(
...
+      if (!initializationStatuses[index]) {
+        initializationStatuses[index] = true
+        count++
+      }
     }
+
+    operator fun get(index: Int) = parameterValues[index]
+
+    override val entries: MutableSet<MutableMap.MutableEntry<KParameter, Any?>>
+      get() = parameterKeys.mapIndexedNotNull { index, value ->
```
ZacSweers, line 188: is there a `mapIndexedNotNullTo` we could use to avoid the second set creation?

ZacSweers (commented):
Let's go with the simplified format, also left some other comments

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/SpreadWrapper.java
k163377: `SpreadWrapper` is a `Java` class.  

The reason it is written in `Java` is to avoid using `spread operator`.
I felt that this was a hack that should not be included in this module, so I removed it.

k163377:
@ZacSweers 
I fixed to the simplest form, removing hacks and extra changes.

I apologize for wasting your review.

kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
k163377: I have fixed it.

https://github.com/square/moshi/pull/1286/files#diff-129f467727ea6286b698c1c3be4e4844742f064fc0aeacae5af5dc338d7fe3e4R113-R117

ZacSweers (approved):

ZacSweers:
Apologies for the delay on this, this looks good to me and we ended up adopting something similar in codegen too. Thanks!

PR #1293, opened 2021-01-31
DO NOT MERGE: Add example failing test

alexvanyo:
This PR adds a failing test, which should cause CI to fail. However, it currently isn't due to #1294


ZacSweers:
Resolved by #1295

PR #1299, opened 2021-02-06
fix: removed an extra blank line.

apigeoneer:
* My pull request has a descriptive title.
* It targets the master branch of Moshi.
* All files I've changed are in the same language i.e. English.


ZacSweers:
This is a generated file, if you want to suggest a fix then you should PR this on Gradle itself

apigeoneer:
sure.

PR #1303, opened 2021-02-14
Suppress RedundantVisibilityModifier in generated adapters

ZacSweers:
KotlinPoet always generates explicit `PUBLIC` modifiers for public members now, so this avoid another point of compiler warnings in generated code


Egorand (approved):
I think it's weird that the compiler flags this in generated code. But I guess compiler doesn't really know it's generated code? Could be something to raise to JB, but might not really be actionable.

ZacSweers:
Yeah we suppress the others for the same reason, there's no way to disable certain checks on only certain files (I.e. generated code)

PR #1306, opened 2021-02-15
Suppress some poet deprecations

ZacSweers:
These are unlikely to actually go away in 2.0 since they make sense in this case

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -231,7 +231,7 @@ internal fun targetType(
   val resolvedTypes = mutableListOf<ResolvedTypeMapping>()
   val superTypes = appliedType.supertypes(types)
     .filterNot { supertype ->
-      @Suppress("DEPRECATION")
+      @Suppress("DEPRECATION") // Appropriate in this case
       supertype.element.asClassName() == OBJECT_CLASS || // Don't load properties for java.lang.Object.
         supertype.element.kind != ElementKind.CLASS // Don't load properties for interface types.
     }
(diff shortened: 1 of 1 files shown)


Egorand (approved):
Would it be helpful to add a short comment to each suppress to explain why it's being suppressed?

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -231,6 +231,7 @@ internal fun targetType(
   val resolvedTypes = mutableListOf<ResolvedTypeMapping>()
   val superTypes = appliedType.supertypes(types)
     .filterNot { supertype ->
+      @Suppress("DEPRECATION")
```
ZacSweers, line 234:
```suggestion
      @Suppress("DEPRECATION") // Appropriate in this case
```

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -270,6 +271,7 @@ internal fun targetType(
         val superSuperClass = supertype.element.superclass as DeclaredType
 
         // Convert to an element and back to wipe the typed generics off of this
+        @Suppress("DEPRECATION")
```
ZacSweers, line 274:
```suggestion
        @Suppress("DEPRECATION") // Appropriate in this case
```

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -285,6 +287,7 @@ internal fun targetType(
     }
 
   for ((localAppliedType, supertypeApi) in superTypes.entries) {
+    @Suppress("DEPRECATION")
```
ZacSweers, line 290:
```suggestion
    @Suppress("DEPRECATION") // Appropriate in this case
```

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -311,6 +314,8 @@ internal fun targetType(
       .any { it.isInternal }
     if (forceInternal) KModifier.INTERNAL else visibility
   }
+
+  @Suppress("DEPRECATION")
```
ZacSweers, line 318:
```suggestion
  @Suppress("DEPRECATION") // Appropriate in this case
```

PR #1287, opened 2021-01-13
Fix a crash when processing a parameter annotated with an annotation that placed in annotation package

kubode:
This problem happens with KotlinPoet 1.7.x.

From KotlinPoet 1.7.x, `TypeName.toString()` is escaped with backquotes if the package name contains keywords.

e.g. `ClassNames.get(java.lang.annotation.Annotation::class).toString()`

KotlinPoet 1.6.x | KotlinPoet 1.7.x
--|--
`java.lang.annotation.Annotation` | ``java.lang.`annotation`.Annotation``

So NPE will be thrown if an annotation is placed in `annotation` package because `elements.getTypeElement(it.typeName.toString())` returns `null`.

Edit: Also resolves #1203

PR #1307, opened 2021-02-15
Update dependencies

ZacSweers:
Gradle, Kotlin, ~KotlinPoet~, ~Truth~


Egorand (approved):

ZacSweers:
This is a real head scratcher - kotlinpoet 1.7.x appears to be what breaks this and results in no generated adapters. It does appear to work in kotlin-compile-testing tests though. Need to dig in more

ZacSweers:
KotlinPoet issue was fixed in #1287

PR #1308, opened 2021-02-15
Fix private computed properties not being ignored

### LINKED ISSUE #1259: Kotlin codegen causes compile error for private computed val
If I try to use the code generator for Kotlin class with a private computed val, it fails with an error.

```Kotlin
data class SampleClass(
        val fieldA: String,
        val fieldB: Int
) {

    private val computed: String
        get() = if (fieldB == 0) {
            "zero"
        } else {
            fieldA
        }

    // Other code here that uses the computed property
}
```
gives the error "SampleClass.java:7: error: property computed is not visible"

But `computed` is a virtual property with no actual backing field, so Moshi should be ignoring it.

The usual advice I found on Stack Overflow of having Moshi ignore a field with `@Transient` also doesn't work, giving a Kotlin compile error reporting that the annotation is not applicable to a property without a backing field or delegate.

I originally saw this with version 1.9.2, and it's still present in version 1.11.0.

The workaround is to just remove the `private` keyword and make the property visible. Doing so appears to be harmless, as the generated JsonAdapter class isn't trying to reference `computed` anyways.

ZacSweers:
Resolves #1259

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -430,7 +430,7 @@ internal fun TargetProperty.generator(
     return PropertyGenerator(this, DelegateKey(type, emptyList()), true)
   }
 
-  if (!isVisible) {
+  if (!isVisible && parameter != null) {
     messager.printMessage(
       ERROR,
       "property $name is not visible",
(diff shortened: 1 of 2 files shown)


ZacSweers:
Looking at this more, I'm actually not sure what the original intended behavior was here. It looks like it wants private properties to also be annotated with `@Transient`, but the ask in the issue is sort of asking for them to be treated as implicitly transient properties. We could do that, but takes a bit more work. Thoughts?

ZacSweers:
Superseded by Json.ignore

PR #1323, opened 2021-03-24
Warn on missing TypeElement rather than error

ZacSweers:
One example I found for this is when a property is annotated with android's `@SuppressLint`, which is only available from the android jar and not visible during apt. This should just be a warning (if even?) and not a hard error.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
@@ -62,6 +62,7 @@ import javax.lang.model.type.DeclaredType
 import javax.lang.model.util.Elements
 import javax.lang.model.util.Types
 import javax.tools.Diagnostic.Kind.ERROR
+import javax.tools.Diagnostic.Kind.WARNING
 
 private val JSON_QUALIFIER = JsonQualifier::class.java
 private val JSON = Json::class.asClassName()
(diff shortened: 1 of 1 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
```diff
@@ -486,13 +487,17 @@ internal fun TargetProperty.generator(
...
-private fun List<AnnotationSpec>?.qualifiers(elements: Elements): Set<AnnotationSpec> {
+private fun List<AnnotationSpec>?.qualifiers(
+  messager: Messager,
+  elements: Elements
+): Set<AnnotationSpec> {
   if (this == null) return setOf()
   return filterTo(mutableSetOf()) {
-    val typeElement = checkNotNull(elements.getTypeElement(it.typeName.rawType().canonicalName)) {
-      "Could not get the type element of $it"
+    val typeElement: TypeElement? = elements.getTypeElement(it.typeName.rawType().canonicalName)
```
Egorand, line 496: Wouldn't `typeElement` here represent the enclosing type, and not an annotation on the element we're looking at (as in the example with `@SuppressLint`)? I think a test would help visualize the problem that's being solved.

Egorand (approved):

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/metadata.kt
ZacSweers: I can't really reproduce it as it's specific to how AGP and laptop wire their compilation classpaths. This is iterating over all the annotations on a given property, so the element would be the annotation type

PR #1325, opened 2021-03-29
Bump the Kotlin JVM metadata library to the latest release. 

vRallev:
This will avoid problems with Kotlin 1.5.0-M2.


vRallev:
I verified locally that it works in our project and unblocks `1.5.0-M2`.

PR #1326, opened 2021-03-30
Invoke constructor directly if all params with default are set

### LINKED ISSUE #1302: Code gen optimization: check if all params with defaults are set
Currently, we always look up and reflectively invoke the defaults constructor. We should, in theory, be able to check if all the masks have their default parameters fulfilled. If so, we can happy-path that and invoke the normal constructor. It'd look a little meaty in source but pretty low overhead in bytecode. Only concern I'd have is that, until contracts are stable and we can (mis)use them to remove intrinsics, we are technically _adding_ the cost of an intrinsic nullcheck for non-nullable vars that we don't currently pay with the reflection approach.

```kotlin
// During code gen

// New - track what an "all required params set" looks like
var allRequiredSetMask = -1

// later in code gen
val inverted = (1 shl maskIndex).inv()
result.addComment("\$mask = \$mask and (1 shl %L).inv()", maskIndex)
result.addStatement(
  "%1L = %1L and 0x%2L.toInt()",
  maskNames[maskNameIndex],
  Integer.toHexString(inverted)
)

// Set index in 
if (hasDefault) {
  allRequiredSet = allRequiredSetMask and inverted
}

// new in generated code = compare against computed "all set" value
if (mask0 == 0xffffffff) {
  // Invoke constructor directly with named args
}
```

ZacSweers:
Resolves #1302

Ported from my KSP impl [KSP impl](https://github.com/ZacSweers/MoshiX/pull/86). Now this class would result in in the associated generated adapter:

```kotlin
data class Example(val param1: String, val param2: String = "")
```

```kotlin
public class ExampleJsonAdapter(
  moshi: Moshi
) : JsonAdapter<Example>() {
  // Properties...

  public override fun toString(): String = buildString(29) {
      append("GeneratedJsonAdapter(").append("Example").append(')') }

  public override fun fromJson(reader: JsonReader): Example {
    var param1: String? = null
    var param2: String? = null
    var mask0 = -1
    reader.beginObject()
    while (reader.hasNext()) {
        0 -> param1 = stringAdapter.fromJson(reader) ?: throw Util.unexpectedNull("param1",
            "param1", reader)
        1 -> {
          param2 = stringAdapter.fromJson(reader) ?: throw Util.unexpectedNull("param2", "param2",
              reader)
          // $mask = $mask and (1 shl 1).inv()
          mask0 = mask0 and 0xfffffffd.toInt()
        }
        -1 -> {
          // Unknown name, skip it.
          reader.skipName()
          reader.skipValue()
        }
      }
    }
    reader.endObject()
    if (mask0 == 0xfffffffd.toInt()) {
      // All parameters with defaults are set, invoke the constructor directly
      return  Example(
          param1 = param1 ?: throw Util.missingProperty("param1", "param1", reader),
          param2 = param2 as String
      )
    } else {
      @Suppress("UNCHECKED_CAST")
      val localConstructor: Constructor<Example> = this.constructorRef ?:
          Example::class.java.getDeclaredConstructor(String::class.java, String::class.java,
          Int::class.javaPrimitiveType, Util.DEFAULT_CONSTRUCTOR_MARKER).also {
          this.constructorRef = it }
      return localConstructor.newInstance(
          param1 ?: throw Util.missingProperty("param1", "param1", reader),
          param2,
          mask0,
          null
      )
    }
  }

  public override fun toJson(writer: JsonWriter, value: Example?): Unit {
    // ...
  }
}
```

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -366,6 +366,7 @@ internal class AdapterGenerator(
     val maskNames = Array(maskCount) { index ->
       nameAllocator.newName("mask$index")
     }
+    val maskAllSetValues = Array(maskCount) { -1 }
     val useDefaultsConstructor = components.filterIsInstance<ParameterComponent>()
       .any { it.parameter.hasDefault }
     if (useDefaultsConstructor) {
@@ -440,6 +441,9 @@ internal class AdapterGenerator(
         }
         if (property.hasConstructorDefault) {
           val inverted = (1 shl maskIndex).inv()
+          if (input is ParameterComponent && input.parameter.hasDefault) {
+            maskAllSetValues[maskNameIndex] = maskAllSetValues[maskNameIndex] and inverted
+          }
           result.addComment("\$mask = \$mask and (1 shl %L).inv()", maskIndex)
           result.addStatement(
             "%1L = %1L and 0x%2L.toInt()",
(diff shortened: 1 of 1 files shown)


Egorand (approved):
Nice! Needs tests?

ZacSweers:
Nope! No new tests needed at least. This is a nice case where existing tests cover it functionally and compilation is a good layer of redundancy

ZacSweers:
Some folks have also been using the moshi-ksp impl for awhile and haven't gotten any issues reported, for added peace of mind

PR #1333, opened 2021-04-07
Fix version on readme

k163377:
Fixed from `1.11.0` to `1.12.0`.

## Question
Still have 1.11.0 in the following part, should fix it?
https://github.com/square/moshi/blob/1654313feed48ab09e6edfe7ebbef23a57869020/moshi/japicmp/build.gradle.kts#L28
https://github.com/square/moshi/blob/1654313feed48ab09e6edfe7ebbef23a57869020/adapters/japicmp/build.gradle.kts#L28

I'm not good at English, so I'm sorry if there is any rudeness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,7 +664,7 @@ add the following to your build to enable the annotation processor:
 ```
 
 ```kotlin
-kapt("com.squareup.moshi:moshi-kotlin-codegen:1.11.0")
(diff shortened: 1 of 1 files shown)


ZacSweers (approved):
Thanks!

ZacSweers:
Don't worry about the build file for now, we can update that later

PR #1339, opened 2021-05-06
Update to Kotlin 1.5 (and associated deps)

ZacSweers:
This updates to Kotlin 1.5. Note that there's a larger KotlinPoet update in the works too, as there's a few changes in the kotlinx-metadata infrastructure.

This starts some preliminary support of value classes, but only just bumping our test example up.

Other bits:
* Auto-migrated some deprecated APIs
* Kept jvmTarget at "1.6" for backward compatibility, but we could revisit this
* ktlint update has to go with this too as newer version supports the `value class` keyword combo

### DIFF
--- a/adapters/build.gradle.kts
+++ b/adapters/build.gradle.kts
@@ -14,10 +14,18 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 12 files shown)


Egorand (approved):
Hype!

PR #1343, opened 2021-05-09
ci: add JDK 16

sullis:


ZacSweers:
You need to add custom jvm args for kapt to work with jdk 16

PR #1354, opened 2021-06-01
Update README to include enum section in ProGuard/R8 rules

iurysza:
I've updated the README to include a section to describe how to work with enums in builds using Proguard/R8.
Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,7 +708,7 @@ Moshi contains minimally required rules for its own internals to work without ne
 
 #### Enums
 
-Remember to annotate enums with `@JsonClass(generateAdapter = false)` to prevent them from being removed from your code by R8/ProGuard.
+Annotate enums with `@JsonClass(generateAdapter = false)` to prevent them from being removed/obfuscated from your code by R8/ProGuard.
 
 License
 --------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -706,6 +706,10 @@ R8 / ProGuard
 
 Moshi contains minimally required rules for its own internals to work without need for consumers to embed their own. However if you are using reflective serialization and R8 or ProGuard, you must add keep rules in your proguard configuration file for your reflectively serialized classes.
 
+#### Enums
+
+Remember to annotate enums with `@JsonClass(generateAdapter = false)` to prevent them from being removed from your code by R8/ProGuard.
```
ZacSweers, line 711:
```suggestion
Annotate enums with `@JsonClass(generateAdapter = false)` to prevent them from being removed/obfuscated from your code by R8/ProGuard.
```

ZacSweers (approved):
One wording nit!

adams-y-chen:
Thanks for the updating the documentation. I tested the annotation using the code like below. The enum class still gets obfuscated in the apk. On the other hand, Proguard rule "-keep class xxx" would stop the obfuscation.

Did I miss something? Is there anything else that is needed in addition to the `@JsonClass(generateAdapter = false)` annotation?

```
@JsonClass(generateAdapter = false)
enum class MyEnumClass {
    MyEnum1,
    MyEnum2
}
```

Issue #1345, opened 2021-05-16
Proposal: Add Kotlin examples to the README.md 


sangeetds:
The README.md file lacks any Kotlin example for anyone to begin with. It would be nice if it would include examples like it already has for Java. 
Will be happy to add the examples if this sounds good.

Egorand:
Sounds good, feel free to send us a PR! A good place to start might be converting the existing examples into Kotlin, similar to what we did for Okio: https://square.github.io/okio/

@ZacSweers any other Kotlin examples that you think we should add to README?

ZacSweers:
Not that I can think of, just a simple example should suffice

sangeetds:
I went through the examples on the Okia website. Java/Kotlin codes were written in tabsets there. How do you suggest I write it here. Would Java code followed by Kotlin work? Or will collapsible sections be better?

Egorand:
Both options sound good to me! We can try collapsible sections with the Java section being revealed by default while the Kotlin one would be collapsed by default.

sangeetds:
Hey @ZacSweers @Egorand! Bumping this thread for visibility cause I forgot to mention that I have raised a PR for the same.

PR #1355, opened 2021-06-02
Update README.md to include Kotlin Examples

sangeetds:
Added Kotlin Examples, as discussed in Issue #1345

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ class EventJsonAdapter {
   @FromJson fun eventFromJson(eventJson: EventJson): Event {
     val event = Event()
     event.title = eventJson.title
-    event.beginDateAndTime = "${eventJson.begin_date}  ${eventJson.begin_time}"
+    event.beginDateAndTime = "${eventJson.begin_date} ${eventJson.begin_time}"
     return event
   }
 
--- a/README.md
+++ b/README.md
@@ -787,11 +787,11 @@ public final class BlackjackHand {
     <summary>Kotlin</summary>
 
 ```kotlin
-class BlackjackHand(
-  @Transient private val total: Int,
+class BlackjackHand(...) {
+  @Transient var total: Int
   
   ...
-)
+}
 ```
 </details>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -197,21 +340,72 @@ class EventJsonAdapter {
...
+
+<details>
+    <summary>Kotlin</summary>
+
+```kotlin
+class EventJsonAdapter {
+  @FromJson fun eventFromJson(eventJson: EventJson): Event {
+    val event = Event()
+    event.title = eventJson.title
+    event.beginDateAndTime = "${eventJson.begin_date}  ${eventJson.begin_time}"
```
saket, line 353:
```suggestion
    event.beginDateAndTime = "${eventJson.begin_date} ${eventJson.begin_time}"
```

saket (approved):
LGTM!

@ZacSweers @swankjesse Thoughts on replacing all java blocks with kotlin instead of keeping them both?

README.md
```diff
@@ -341,9 +578,23 @@ class Player {
   ...
 }
 ```
+</details>
 
-Because JSON field names are always defined with their Java fields, Moshi makes it easy to find
-fields when navigating between Java and JSON.
+<details>
```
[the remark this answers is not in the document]
sangeetds: You're right. Should I include a message saying it is assuming the use of `KotlinJsonAdapterFactory` and code-gen?

README.md
```diff
@@ -787,11 +787,11 @@ public final class BlackjackHand {
     <summary>Kotlin</summary>
 
 ```kotlin
-class BlackjackHand {
-  @Transient private val total: Int
+class BlackjackHand(
```
ZacSweers, line 790: This case should have a default value assigned to it or be converted to a non-constructor var property. Probably something like this

```kotlin
class BlackjackHand(...) {
  @Transient
  private var total: Int = 0
}
```

ZacSweers (approved):
One more nit!

README.md
sangeetds: Done.

ZacSweers (approved):

saket (approved):

sangeetds:
@ZacSweers can you run the tests again? Thanks

ZacSweers:
Merged. Thanks a ton!

PR #1364, opened 2021-07-12
set Gitignore: ignore local.properties

solomaticydl:


JakeWharton:
This is not an Android project, though. This file is never needed so you can just delete it if you have one.

PR #1353, opened 2021-06-01
Add an argument to be disable proguard code generating

SeongUgJung:
This is to make an option to disable proguard rule generating.
I put an option argument : `moshi.enabledProguardGenerated`. 
Once it sets `false`, it will skip to generate proguard rule.

Issue : https://github.com/square/moshi/issues/1351

PR #1365, opened 2021-07-16
Small grammar and naming followups for proguard gen

ZacSweers:
Small followups from #1353


rharter (approved):
This looks clearer to me, but this is a breaking change, right? Need any sort of backwards compat support?

ZacSweers:
Not needed as this hasn't released yet

PR #1381, opened 2021-08-20
Implement reflective support for Java Records

### LINKED ISSUE #1278: Support Java record classes when they are stable
As of the latest release `2.11.0` (and probably past releases too) Moshi is not able to deserialize Java Records, it simply fails with an `AssertionError` when calling `Adapter#fromJson` with an appropriate JSON object.
See the attached very simple sample project for more details: [test_project.zip](https://github.com/square/moshi/files/5630567/test_project.zip)
Stacktrace:
```
Exception in thread "main" java.lang.AssertionError
	at com.squareup.moshi.ClassJsonAdapter.fromJson(ClassJsonAdapter.java:208)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:51)
	at Main.main(Main.java:8)
```

ZacSweers:
Porting from https://github.com/ZacSweers/MoshiX/tree/main/moshi-records-reflect

This makes a few changes along the way to make the build all happy with how we're doing this:
- Standardizing around Java 8 (which is the minimum supported by gradle toolchains and also soon to be the minimum supported by Kotlin)
- Create a separate Java 16 source set for hosting the actual adapter implementation. This is compiled with Java 16 but is then packaged into the main Moshi jar and denoted as a Multi-Release JAR.
- We gate access to this record adapter at runtime by checking if the requisite record APIs are available.
- We separately have a shim copy of the record adapter in the `main` source set for linking in `BUILT_IN_FACTORIES`, but this is booted out by the real one during packaging.
- Replace animal sniffer with `-release 8`, which functionally accomplishes the same
- Tests are in a dedicated separate `records-test` subproject, making it easier to both write new ones and also ensure our packaging rules for the jar is working correctly (we've had success with this in the past with the codegen artifact's shading being tested in a dedicated tests project)

Resolves #1278

Tests are a little skimpy right now, happy to take suggestions on more we should add! Can also look at fleshing them out more in a followup PR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,14 +10,8 @@ jobs:
       fail-fast: false
       matrix:
         java-version:
-          - 8
-          - 9
-          - 10
-          - 11
-          - 12
-          - 13
-          - 14
-          - 15
+          - 16
+          - 17-ea
 
     steps:
       - name: Checkout
@@ -46,7 +40,7 @@ jobs:
         run: ./gradlew build check --stacktrace
 
       - name: Publish (default branch only)
-        if: github.repository == 'square/moshi' && github.ref == 'refs/heads/master' && matrix.java-version == '8'
+        if: github.repository == 'square/moshi' && github.ref == 'refs/heads/master' && matrix.java-version == '16'
         run: ./gradlew uploadArchives
         env:
           ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,6 @@ jobs:
       matrix:
         java-version:
           - 16
-          - 17-ea
 
     steps:
       - name: Checkout
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,7 +69,7 @@ spotless {
       "**/TypesTest.java"
     )
     val configureCommonJavaFormat: JavaExtension.() -> Unit = {
-      googleJavaFormat("1.10")
+      googleJavaFormat("1.11")
     }
     java {
       configureCommonJavaFormat()
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -70,7 +70,7 @@ spotless {
       "**/TypesTest.java"
     )
     val configureCommonJavaFormat: JavaExtension.() -> Unit = {
-      googleJavaFormat("1.7")
+      googleJavaFormat("1.10")
     }
     java {
       configureCommonJavaFormat()
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -43,50 +43,47 @@ spotless {
     indentWithSpaces(2)
     endWithNewline()
   }
-  // GJF not compatible with JDK 15 yet
-  if (!JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_15)) {
-    val externalJavaFiles = arrayOf(
-      "**/ClassFactory.java",
-      "**/Iso8601Utils.java",
-      "**/JsonReader.java",
-      "**/JsonReaderPathTest.java",
-      "**/JsonReaderTest.java",
-      "**/JsonScope.java",
-      "**/JsonUtf8Reader.java",
-      "**/JsonUtf8ReaderPathTest.java",
-      "**/JsonUtf8ReaderTest.java",
-      "**/JsonUtf8ReaderTest.java",
-      "**/JsonUtf8Writer.java",
-      "**/JsonUtf8WriterTest.java",
-      "**/JsonWriter.java",
-      "**/JsonWriterPathTest.java",
-      "**/JsonWriterTest.java",
-      "**/LinkedHashTreeMap.java",
-      "**/LinkedHashTreeMapTest.java",
-      "**/PolymorphicJsonAdapterFactory.java",
-      "**/RecursiveTypesResolveTest.java",
-      "**/Types.java",
-      "**/TypesTest.java"
+  val externalJavaFiles = arrayOf(
+    "**/ClassFactory.java",
+    "**/Iso8601Utils.java",
+    "**/JsonReader.java",
+    "**/JsonReaderPathTest.java",
+    "**/JsonReaderTest.java",
+    "**/JsonScope.java",
+    "**/JsonUtf8Reader.java",
+    "**/JsonUtf8ReaderPathTest.java",
+    "**/JsonUtf8ReaderTest.java",
+    "**/JsonUtf8ReaderTest.java",
+    "**/JsonUtf8Writer.java",
+    "**/JsonUtf8WriterTest.java",
+    "**/JsonWriter.java",
+    "**/JsonWriterPathTest.java",
+    "**/JsonWriterTest.java",
+    "**/LinkedHashTreeMap.java",
+    "**/LinkedHashTreeMapTest.java",
+    "**/PolymorphicJsonAdapterFactory.java",
+    "**/RecursiveTypesResolveTest.java",
+    "**/Types.java",
+    "**/TypesTest.java"
+  )
+  val configureCommonJavaFormat: JavaExtension.() -> Unit = {
+    googleJavaFormat("1.11.0")
+  }
+  java {
+    configureCommonJavaFormat()
+    target("**/*.java")
+    targetExclude(
+      "**/spotless.java",
+      "**/build/**",
+      *externalJavaFiles
     )
-    val configureCommonJavaFormat: JavaExtension.() -> Unit = {
-      googleJavaFormat("1.11")
-    }
-    java {
-      configureCommonJavaFormat()
-      target("**/*.java")
-      targetExclude(
-        "**/spotless.java",
-        "**/build/**",
-        *externalJavaFiles
-      )
-      licenseHeaderFile("spotless/spotless.java")
-    }
-    format("externalJava", JavaExtension::class.java) {
-      // These don't use our spotless config for header files since we don't want to overwrite the
-      // existing copyright headers.
-      configureCommonJavaFormat()
-      target(*externalJavaFiles)
-    }
+    licenseHeaderFile("spotless/spotless.java")
+  }
+  format("externalJava", JavaExtension::class.java) {
+    // These don't use our spotless config for header files since we don't want to overwrite the
+    // existing copyright headers.
+    configureCommonJavaFormat()
+    target(*externalJavaFiles)
   }
   kotlin {
     ktlint(Dependencies.ktlintVersion).userData(mapOf("indent_size" to "2"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -127,10 +127,13 @@ subprojects {
   }
 
   // Apply with "java" instead of just "java-library" so kotlin projects get it too
-  pluginManager.withPlugin("java") {
-    configure<JavaPluginExtension> {
-      sourceCompatibility = JavaVersion.VERSION_1_7
-      targetCompatibility = JavaVersion.VERSION_1_7
+  if (project.name != "records-tests") {
+    pluginManager.withPlugin("java") {
+      configure<JavaPluginExtension> {
+        toolchain {
+          languageVersion.set(JavaLanguageVersion.of(8))
+        }
+      }
     }
   }
 
@@ -146,6 +149,7 @@ subprojects {
       kotlinOptions {
         @Suppress("SuspiciousCollectionReassignment")
         freeCompilerArgs += listOf("-progressive")
+        jvmTarget = "1.8"
       }
     }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,7 +33,6 @@ plugins {
   id("com.vanniktech.maven.publish") version "0.14.2" apply false
   id("org.jetbrains.dokka") version "1.4.32" apply false
   id("com.diffplug.spotless") version "5.12.4"
-  id("ru.vyarus.animalsniffer") version "1.5.3" apply false
   id("me.champeau.gradle.japicmp") version "0.2.9" apply false
 }
 
@@ -134,13 +133,6 @@ subprojects {
           languageVersion.set(JavaLanguageVersion.of(8))
         }
       }
-    }
-  }
-
-  pluginManager.withPlugin("ru.vyarus.animalsniffer") {
-    dependencies {
-      "compileOnly"(Dependencies.AnimalSniffer.annotations)
-      "signature"(Dependencies.AnimalSniffer.java7Signature)
     }
   }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -126,12 +126,15 @@ subprojects {
   }
 
   // Apply with "java" instead of just "java-library" so kotlin projects get it too
-  if (project.name != "records-tests") {
-    pluginManager.withPlugin("java") {
-      configure<JavaPluginExtension> {
-        toolchain {
-          languageVersion.set(JavaLanguageVersion.of(8))
-        }
+  pluginManager.withPlugin("java") {
+    configure<JavaPluginExtension> {
+      toolchain {
+        languageVersion.set(JavaLanguageVersion.of(16))
+      }
+    }
+    if (project.name != "records-tests") {
+      tasks.withType<JavaCompile>().configureEach {
+        options.release.set(8)
       }
     }
   }
--- a/gradle.properties
+++ b/gradle.properties
@@ -29,6 +29,20 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8 \
   --add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
   --add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
 
+kotlin.daemon.jvmargs=-Dfile.encoding=UTF-8 \
+  --add-opens=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED \
+  --add-opens=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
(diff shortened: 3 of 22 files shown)


ZacSweers:
It looks like animal sniffer ships a shaded old version of ASM and is also no longer being updated since the `-release` flag covers the same on JDK 9+. Thing is we want to support 8, so we can't use it yet. Seems like we'll need to drop animal sniffer if we want to proceed. Thoughts?

moshi/build.gradle.kts
```diff
@@ -22,11 +22,43 @@ plugins {
...
+  options.release.set(16)
+}
+
+// We need to include our actual RecordJsonAdapter from java16 sources and replace the shime
+tasks.named<Jar>("jar") {
+  from(java16.output) {
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+  }
+  manifest {
+    // Indicate that this is a multi release jar
```
melix, line 47: Is it? Just looks like a regular jar to me.

ZacSweers: I was sort of assuming this was needed if the jar contains classes targeting multiple java versions but can omit it if it doesn't really change anything

ZacSweers: 3eb768f

moshi/src/main/java/com/squareup/moshi/internal/Util.java
```diff
@@ -55,6 +56,8 @@
   /** A map from primitive types to their corresponding wrapper types. */
   private static final Map<Class<?>, Class<?>> PRIMITIVE_TO_WRAPPER_TYPE;
 
+  private static final AtomicBoolean RECORDS_AVAILABLE = new AtomicBoolean(false);
```
JakeWharton, line 59: Normal `boolean`? Static init blocks cannot run concurrently.

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+                + replaced.name
+                + "\n"
+                + "    "
+                + componentBinding.name);
+      }
+    }
+    Constructor<Object> constructor;
+    try {
+      //noinspection unchecked
+      constructor = (Constructor<Object>) rawType.getDeclaredConstructor(constructorParams);
```
JakeWharton, line 115: Is there no way to get a MethodHandle from the Class once we know it's a record? Can't believe they expect you to use ancient java.lang.reflect reflection to programmatically instantiate these...

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+    } catch (NoSuchMethodException e) {
+      throw new AssertionError(e);
+    }
+    return new RecordJsonAdapter<>(constructor, bindings).nullSafe();
+  };
+
+  private static record ComponentBinding<T>(String name, String jsonName, JsonAdapter<T> adapter, Method accessor) {
+  }
+
+    private final String targetClass;
```
JakeWharton, line 126: Wrong indent. GJF bug?

build.gradle.kts
```diff
@@ -70,7 +69,7 @@ spotless {
       "**/TypesTest.java"
     )
     val configureCommonJavaFormat: JavaExtension.() -> Unit = {
-      googleJavaFormat("1.7")
+      googleJavaFormat("1.10")
```
JakeWharton, line 72: 1.11 latest

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Override
+    public void toJson(JsonWriter writer, T value) throws IOException {
+      writer.beginObject();
+
+      for (var binding : componentBindingsArray) {
+        writer.name(binding.jsonName);
+        try {
+          binding.adapter.toJson(writer, binding.accessor.invoke(value));
+        } catch (IllegalAccessException | InvocationTargetException e) {
```
JakeWharton, line 181: `InvocationTargetException` should be unwrapped and re-thrown since it indicates an exception on the other side of the reflection.
```java
catch (InvocationTargetException e) {
  Throwable cause = e.getCause();
  if (cause instanceof RuntimeException) throw (RuntimeException) cause;
  if (cause instanceof Error) throw (Error) cause;
  throw new RuntimeException(cause);
}
```

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+      }
+      reader.endObject();
+
+      var initialAccess = constructor.canAccess(null);
+      try {
+        if (!initialAccess) {
+          constructor.setAccessible(true);
+        }
+        return constructor.newInstance(resultsArray);
+      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
```
JakeWharton, line 166: Same InvocationTargetException unwrapping

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+        throw new AssertionError(e);
+      } finally {
+        constructor.setAccessible(false);
```
JakeWharton, line 169: Is there a reason to reset this? I expect the `canAccess` check to not be cheap such that I'm wondering why we don't make it accessible once and never again.

JakeWharton (commented):
Where do you actually pack the MR jar? I don't see any references to `META-INF/versions/16/`

ZacSweers:
> Where do you actually pack the MR jar? I don't see any references to META-INF/versions/16/

Yup removed! I misunderstood the requirements for this

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: I might be missing something but java doesn't seem to like this
image

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: GJF bug, I'll file something over there

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Honestly never used that API, let me look

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Actually I think spotless is ignoring this? Let me look more

ZacSweers: Yeah spotless is ignoring this. Formatting it with GJF manually for now

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Fair enough, I meant to clean this up with the previous cleanup I did with Method

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Good news is we can use MethodHandle for this but bad news is `RecordComponent` itself still returns only `Method` for its accessors :/

build.gradle.kts
ZacSweers: 20d9715

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: d98ee90

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: 765a642 and also part of later changes

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Folded into 355422c

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Constructor bit is in 355422c

I also wrapped the accessor Method with one too, but not sure (lack of familiarity) if this is worthwhile 4700a63. Let me know

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: Cleaned up with the constructor conversion to MethodHandle

JakeWharton:
No MR jar probably means tools like Graal and D8 (and others) will choke processing the jar.

Instead of a gated use and stub you can make the stub return null from a factory method and make the 16 class return a real implementation. Then you unconditionally call the factory and gate behavior on non-null values.

ZacSweers:
ah good idea. Thanks for bearing with me, learning something new with these MR jars!

ZacSweers:
Ok think I've got it working now. This is neat!

image

ZacSweers:
Being reminded how much the java module system sucks when you have to configure stuff

moshi/src/main/java/com/squareup/moshi/internal/Util.java
ZacSweers: Ended up removing this

.github/workflows/build.yml
```diff
@@ -10,14 +10,7 @@ jobs:
...
         java-version:
-          - 8
-          - 9
-          - 10
-          - 11
-          - 12
-          - 13
-          - 14
-          - 15
+          - 16
```
swankjesse, line 13: Not testing on Java 8 is annoying. Something to look into separately from this change.

build.gradle.kts
```diff
@@ -129,15 +125,14 @@ subprojects {
...
+        languageVersion.set(JavaLanguageVersion.of(16))
+      }
     }
-  }
-
-  pluginManager.withPlugin("ru.vyarus.animalsniffer") {
-    dependencies {
-      "compileOnly"(Dependencies.AnimalSniffer.annotations)
-      "signature"(Dependencies.AnimalSniffer.java7Signature)
+    if (project.name != "records-tests") {
```
swankjesse, line 132: Very nice

.github/workflows/build.yml
ZacSweers: We can make the main tests run with java 8 via Gradle toolchains

gradle.properties
```diff
@@ -15,9 +15,36 @@
 #
 
 # For Dokka https://github.com/Kotlin/dokka/issues/1405
-org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
+org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8 \
+  --add-opens=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED \
```
swankjesse, line 19: These are ... unfortunate. The Dokka issue above seems unrelated. I think this is the tracking issue.
https://youtrack.jetbrains.com/issue/KT-45545

kotlin/codegen/build.gradle.kts
```diff
@@ -27,18 +27,27 @@ plugins {
...
     )
   }
 }
 
-// To make Gradle happy
-java {
-  sourceCompatibility = JavaVersion.VERSION_1_8
-  targetCompatibility = JavaVersion.VERSION_1_8
+tasks.withType<Test>().configureEach {
+  // For kapt to work with kotlin-compile-testing
```
swankjesse, line 38: Tracking here: https://youtrack.jetbrains.com/issue/KT-45545

moshi/build.gradle.kts
```diff
@@ -19,14 +19,49 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
   kotlin("jvm")
   id("com.vanniktech.maven.publish")
-  id("ru.vyarus.animalsniffer")
+}
+
+val mainSourceSet by sourceSets.named("main")
+val java16 by sourceSets.creating {
```
swankjesse, line 25: neat

moshi/build.gradle.kts
```diff
@@ -35,6 +70,8 @@ tasks.withType<KotlinCompile>()
   }
 
 dependencies {
+  // So the j16 source set can "see" main Moshi sources
+  "java16Implementation"(mainSourceSet.output)
```
swankjesse, line 74: off topic: this syntax offends me so much

moshi/records-tests/src/test/java/com/squareup/moshi/records/RecordsTest.java
```diff
@@ -0,0 +1,195 @@
...
+  public void genericBoundedRecord() throws IOException {
+    var adapter =
+        moshi.<GenericBoundedRecord<Integer>>adapter(
+            Types.newParameterizedTypeWithOwner(
+                RecordsTest.class, GenericBoundedRecord.class, Integer.class));
+    assertThat(adapter.fromJson("{\"value\":4}")).isEqualTo(new GenericBoundedRecord<>(4));
+  }
+
+  @Test
+  public void qualifiedValues() throws IOException {
```
swankjesse, line 160: awesome

moshi/src/main/java/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+package com.squareup.moshi;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * This is just a simple shim for linking in {@link StandardJsonAdapters} and swapped with a real
```
swankjesse, line 25: I really like how this plugs in

swankjesse: Feels very reminiscent of expect/actual

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,218 @@
...
+
+        var rawType = Types.getRawType(type);
+        if (!rawType.isRecord()) {
+          return null;
+        }
+
+        Map<String, Type> mappedTypeArgs = null;
+        if (type instanceof ParameterizedType parameterizedType) {
+          Type[] typeArgs = parameterizedType.getActualTypeArguments();
+          var typeVars = rawType.getTypeParameters();
```
swankjesse, line 60: var! (My first time seeing this in real life)

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,218 @@
...
+          return null;
+        }
+
+        var rawType = Types.getRawType(type);
+        if (!rawType.isRecord()) {
+          return null;
+        }
+
+        Map<String, Type> mappedTypeArgs = null;
+        if (type instanceof ParameterizedType parameterizedType) {
```
swankjesse, line 58: This reads to me as very similar to `Util.resolve()`. Presumably we can’t use that here because this needs more records stuff? Either way this behavior really wants to be extracted into a separately-testable thing

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,218 @@
...
+        MethodHandle constructor;
+        try {
+          constructor = lookup.findConstructor(rawType, methodType(void.class, constructorParams));
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+          throw new AssertionError(e);
+        }
+        return new RecordJsonAdapter<>(constructor, rawType.getSimpleName(), bindings).nullSafe();
+      };
+
+  private static record ComponentBinding<T>(
```
swankjesse, line 136: yo dog

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
```diff
@@ -0,0 +1,218 @@
...
+      String name, String jsonName, JsonAdapter<T> adapter, MethodHandle accessor) {}
+
+  private final String targetClass;
+  private final MethodHandle constructor;
+  private final ComponentBinding<Object>[] componentBindingsArray;
+  private final JsonReader.Options options;
```
swankjesse, line 142: Yay options

swankjesse (commented):
I think I’d like to do a follow-up to this to test its internal components, though that shouldn’t block release.

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
swankjesse: Fix: https://github.com/square/moshi/pull/1442/files

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.java
ZacSweers: yeah I think the static is implicit but the IDE didn't give me a warning about it being redundant

gradle.properties
ZacSweers: The kapt parts are no longer relevant or on `master` now at least with Kotlin 1.6

PR #1387, opened 2021-08-29
Configure spotless per subproject

Goooler:
Like [Configure spotless per subproject](https://github.com/square/kotlinpoet/pull/1129) did.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,11 +15,12 @@
  */
 
 import com.diffplug.gradle.spotless.JavaExtension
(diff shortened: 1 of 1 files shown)


ZacSweers:
After thinking on this more, I'm going to close this for the same reason as I did your [EitherNet PR](https://github.com/slackhq/EitherNet/pull/35#issuecomment-917353636). Unless you use Moshi and have a specific use case for these updates, please don't send these types of PRs just to fluff your contribution stats.

PR #1390, opened 2021-09-04
Support direct annotation instantiation in code gen on Kotlin 1.6

### LINKED ISSUE #1382: Directly instantiate qualifier annotations in Kotlin 1.6+
https://youtrack.jetbrains.com/issue/KT-45395 will allow for direct instantiation of annotations, which would save us the current workaround of storing them on a field and reflectively looking them up.

- [ ] Update to Kotlin 1.6
- [ ] Generate code that directly invoke qualifier annotations
- [ ] Remove generated proguard rules for keeping these annotations
- [ ] Remove requirement for qualifier annotations to support `FIELD` use site targets

ZacSweers:
Resolves #1382. This also implements it gated on the target type's language version, which allows this to seamlessly work in a backward-compatible way.

Generated adapter now looks like this

```kotlin
public class GeneratedAdaptersTest_PropertyWithQualifierJsonAdapter(
  moshi: Moshi
) : JsonAdapter<GeneratedAdaptersTest.PropertyWithQualifier>() {
  private val options: JsonReader.Options = JsonReader.Options.of("a", "b")

  private val stringAtUppercaseAdapter: JsonAdapter<String> = moshi.adapter(String::class.java,
      setOf(GeneratedAdaptersTest.Uppercase(inFrench = true)), "a")

  private val stringAdapter: JsonAdapter<String> = moshi.adapter(String::class.java, emptySet(),
      "b")
      
  // ...
}
```


ZacSweers:
The failure is interesting. Basically, the `AdapterMethodsFactory` quality check here fails, but I'm not sure if this is a bug on our end or Kotlin's. Trying to reproduce it in a simple case

https://github.com/square/moshi/blob/a1852a7315b46dc3ee75f391d4096f13d5a6c5ca/moshi/src/main/java/com/squareup/moshi/AdapterMethodsFactory.java#L318

ZacSweers:
It's a bug on their side with `hashCode()`, filed here https://youtrack.jetbrains.com/issue/KT-48606

ZacSweers:
Looks like it's not in the 1.6-M1 branch yet. Will try again when it is

image

ZacSweers:
Still broken in 1.6.0-M1, pushed to 1.6.0-RC now -_-

ZacSweers:
Small update after talking with @swankjesse 

* We'll add an option to optionally disable this behavior, in case there's lingering bugs with the Kotlin implementation
* It will be enabled by default for Kotlin 1.6+

Will wait for #1393 to merge in and Kotlin 1.6-RC (which should have this implemented)

build.gradle.kts
```diff
@@ -125,8 +130,8 @@ subprojects {
   pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
     tasks.withType<KotlinCompile>().configureEach {
       kotlinOptions {
-        @Suppress("SuspiciousCollectionReassignment")
-        freeCompilerArgs += listOf("-progressive")
+//        @Suppress("SuspiciousCollectionReassignment")
```
Egorand, line 133: Delete if not needed?

ZacSweers:
After talking with @Egorand, plan is option A. It's already merged in KotlinPoet and I'll update this after that release is out

swankjesse (commented):
LGTM

PR #1393, opened 2021-09-10
Upstream KSP implementation

ZacSweers:
This upstreams the `moshi-ksp` implementation from https://github.com/zacsweers/moshix.

**Highlights**
* KSP lives in a `codegen.ksp` subpackage alongside the existing apt version
* They reuse the same integration tests in `:moshi:kotlin:tests`. The KSP implementation is functionally compatible!
* Bringing along a few KSP-specific regression tests
  * ~One is presented but commented out until next KotlinPoet release as we ported the same fix there - https://github.com/square/kotlinpoet/pull/1137~
* CI runs a matrix of both kapt and KSP.
  * This also made me realize we could possibly use a CI matrix to truly achieve what #1064 was trying to do (but that's for another PR)
* Switches to auto-service-ksp for for AutoService support and removes kapt itself from the codegen project, as I was really tired of dealing with Kapt's strange shenanigans on JDK 16.

Some other things to do:

- [x] Consolidate options constants
- [x] ~Consolidate processor tests with a parameterized form?~ Will save this for a later PR after https://github.com/google/ksp/issues/621 is resolved
- [ ] ??

PR #1394, opened 2021-09-20
Update to JDK 17

ZacSweers:
Everything still targets JDK 8 (non-records) or JDK 16 (records only)


ZacSweers:
This won't work until #1393 is in because Kapt is a dumpster fire

Issue #1252, opened 2020-10-05
Upgrade Moshi to Kotlin and call it Moshi 2


swankjesse:
As with [Okio](https://github.com/square/okio/issues/370) and [OkHttp](https://github.com/square/okhttp/issues/4723) let’s get Moshi implemented in Kotlin.

## Goals

Implemented in Kotlin with a dependency on the Kotlin standard library. We want the option to support multiplatform including Kotlin/Native .

100% binary compatibility with Moshi 1.x. 

100% Java source compatibility with Moshi 1.x. 

Easy, automated upgrade for Kotlin source

No performance regressions. We must write careful Kotlin, paying attention to abstractions that have a runtime cost.

## Non-Goals

I’d like to do the Kotlin conversion first, and then attack the problem of deeply supporting Kotlin’s richer type system. I prefer to save that for a follow-up release, though I can be convinced otherwise!

## Tasks

 - [ ] Integrate binary-compatibility checks with japicmp.
 - [ ] Write a test case for Kotlin source compatibility
 - [ ] Migrate public API types
 - [ ] Adopt Kotlin idioms

ZacSweers:
Can you expand on what you mean by the type system bit? For instance, I think we may want to make Moshi run on KType under the hood rather than Type, not sure if that has implications on what you were thinking of

swankjesse:
I’d like to do that, but I think we can hold it until 2.1.

ZacSweers:
I'll take a pass at adding the japicmp tooling today or tomorrow. Should be good just pulling from what Okio/OkHttp do?

For benchmarks, we could reuse my [benchmarking project](https://github.com/ZacSweers/json-serialization-benchmarking) with one of 1.x or 2.x shaded for comparisons. Have any other benchmarks in mind?

swankjesse:
Borrowing liberally from OkHttp and Okio is good.

For benchmarks we just want to compare old vs. new with some repeatability.

ZacSweers:
Should we integrate kotlin' ABI validator on top of japicmp? This way we can prevent surprise ABI changes due to kotlin bytecode changes. Noticed leakcanary uses it too and seems fairly easy to integrate

https://github.com/Kotlin/binary-compatibility-validator

JakeWharton:
Why would you use it on top of japicmp? What does it provide that japicmp does not? Last I checked it was just a signature validator exactly like japicmp.

ZacSweers:
[...]

PR #1400, opened 2021-10-12
`JsonUtf8Reader` --> Kt

JvmName:
(I'm aware that this PR will probably block on the work to upstream KSP into Moshi; I figured I'd just get this up while I had the time)

Attempting to follow the guidelines in #1252, I've converted `JsonUtf8Reader` to Kotlin!

This was done the _Convert To Kotlin_ machine-translation, and then I went over each line, compared to the Java source, and fixed/updated as necessary. 

Some decisions that I made that are definitely open to discussion: 
- all code comments + javadocs have been retained, except when the conversion made them irrelevant (e.g. `//fallthrough`, which doesn't exist in `when`s)
- using string templates instead of concat
- created a private inline extension to convert `Byte` to `Char`, since that happens a lot
- I preferred `when` expressions to `if/else`, mostly for readability

I wasn't sure about a few things, so I'll call them out below. Let me know if there's something I've missed!

Oh yeah, fun fact, the Kotlin code is 6% shorter!


moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+  private fun doPeek(): Int {
+    val peekStack = scopes[stackSize - 1]
+    when (peekStack) {
+      JsonScope.EMPTY_ARRAY -> scopes[stackSize - 1] = JsonScope.NONEMPTY_ARRAY
+      JsonScope.NONEMPTY_ARRAY -> {
+        // Look for a comma before the next element.
+        val c = nextNonWhitespace(true).toChar()
+        buffer.readByte() // consume ']' or ','.
+        when (c) {
+          ']' -> return PEEKED_END_ARRAY.also { peeked = it }
```
JvmName, line 183: the original Java sources for this were: 

```java
return peeked = PEEKED_END_ARRAY
```

I figured this is the equivalent Kotlin, but I'm also happy to convert it to 

```kotlin
peeked = PEEKED_END_ARRAY
return peeked
```
if the `.also` construction doesn't feel readable/idiomatic/maintainable

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+          buffer.readByte() // Consume ']'.
+          return PEEKED_END_ARRAY.also { peeked = it }
+        }
+        // In lenient mode, a 0-length literal in an array means 'null'.
+        return when (peekStack) {
+          JsonScope.EMPTY_ARRAY, JsonScope.NONEMPTY_ARRAY -> {
+            checkLenient()
+            PEEKED_NULL.also { peeked = it }
+          }
+          else -> throw syntaxError("Unexpected value")
```
JvmName, line 273: in the original java sources, this code block was a fallthrough for the `';', ','` case; I figured the correct course of action is to duplicate the logic?

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+  }
+
+  @Throws(IOException::class)
+  private fun peekNumber(): Int {
+    var value: Long = 0 // Negative to accommodate Long.MIN_VALUE more easily.
+    var negative = false
+    var fitsInLong = true
+    var last = NUMBER_CHAR_NONE
+    var i: Long = 0
+    charactersOfNumber@ while (true) {
```
JvmName, line 372: I'm not 100% certain that the label is needed, but I've kept it to maintain as much parity w/ the Java sources as possible

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+  }
+
+  @Throws(IOException::class)
+  private fun isLiteral(c: Int): Boolean {
+    return when (c.toChar()) {
+      '/', '\\', ';', '#', '=' -> {
+        checkLenient() // fall-through
+        false
+      }
+      '{', '}', '[', ']', ':', ',', ' ', '\t', '\u000C'/*\f*/, '\r', '\n' -> false
```
JvmName, line 475: > `'\u000C'`

Another fun fact, Kotlin does not recognize `\f` as a valid escape; this was my solution -- although obviously open to other options.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+    pushScope(JsonScope.STREAMING_VALUE)
+    return valueSource!!.buffer()
+  }
+
+  /**
+   * Returns the next character in the stream that is neither whitespace nor a part of a comment.
+   * When this returns, the returned character is always at `buffer.getByte(0)`.
+   */
+  @Throws(IOException::class)
+  private fun nextNonWhitespace(throwOnEof: Boolean): Int {
```
JvmName, line 980: The original java had this return an `int` so I've kept it as `Int` here. 

That being said, I think this could make sense for the return type to be `Char`, since nearly all of the callsites require conversion to `Char` anyway.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+  @Throws(IOException::class)
+  override fun endArray() {
+    var p = peeked
+    if (p == PEEKED_NONE) {
+      p = doPeek()
+    }
+    peeked = if (p == PEEKED_END_ARRAY) {
+      stackSize--
+      pathIndices[stackSize - 1]++
+      PEEKED_NONE
```
JakeWharton, line 107:
```suggestion
    if (p == PEEKED_END_ARRAY) {
      stackSize--
      pathIndices[stackSize - 1]++
      peeked = PEEKED_NONE
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+
+  @Throws(IOException::class)
+  override fun beginObject() {
+    var p = peeked
+    if (p == PEEKED_NONE) {
+      p = doPeek()
+    }
+    peeked = if (p == PEEKED_BEGIN_OBJECT) {
+      pushScope(JsonScope.EMPTY_OBJECT)
+      PEEKED_NONE
```
JakeWharton, line 121:
```suggestion
    if (p == PEEKED_BEGIN_OBJECT) {
      pushScope(JsonScope.EMPTY_OBJECT)
      peeked = PEEKED_NONE
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JakeWharton: I prefer the split version

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+          }
+          else -> {
+            checkLenient()
+            if (isLiteral(c.code)) {
+              PEEKED_UNQUOTED_NAME.also { peeked = it }
+            } else {
+              throw syntaxError("Expected name")
+            }
+          }
+        }
```
JakeWharton, line 226:
```suggestion
        val next = when (val c = nextNonWhitespace(true).toChar()) {
          '"' -> {
            buffer.readByte() // consume the '\"'.
            PEEKED_DOUBLE_QUOTED_NAME
          }
          '\'' -> {
            buffer.readByte() // consume the '\''.
            checkLenient()
            PEEKED_SINGLE_QUOTED_NAME
          }
          '}' -> if (peekStack != JsonScope.NONEMPTY_OBJECT) {
            buffer.readByte() // consume the '}'.
            PEEKED_END_OBJECT
          } else {
            throw syntaxError("Expected name")
          }
          else -> {
            checkLenient()
            if (isLiteral(c.code)) {
              PEEKED_UNQUOTED_NAME
            } else {
              throw syntaxError("Expected name")
            }
          }
        }
        peeked = next
        return next
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JakeWharton: Doesn't seem needed

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+          p = 0
+        }
+        else -> return c.code
+      }
+    }
+    return if (throwOnEof) {
+      throw EOFException("End of input")
+    } else {
+      -1
+    }
```
JakeWharton, line 1039:
```suggestion
    if (throwOnEof) {
      throw EOFException("End of input")
    }
    return -1
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+    /*
+     * This code uses ugly local variables 'p' and 'l' representing the 'pos'
+     * and 'limit' fields respectively. Using locals rather than fields saves
+     * a few field reads for each whitespace character in a pretty-printed
+     * document, resulting in a 5% speedup. We need to flush 'p' to its field
+     * before any (potentially indirect) call to fillBuffer() and reread both
+     * 'p' and 'l' after any (potentially indirect) call to the same method.
+     */
+    var p: Long = 0
+    loop@ while (source.request((p + 1))) {
```
JakeWharton, line 990:
```suggestion
    while (source.request((p + 1))) {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+      pathIndices[stackSize - 1]++
+      return peekedLong.toDouble()
+    }
+    when {
+      p == PEEKED_NUMBER -> peekedString = buffer.readUtf8(peekedNumberLength.toLong())
+      p == PEEKED_DOUBLE_QUOTED -> peekedString = nextQuotedValue(DOUBLE_QUOTE_OR_SLASH)
+      p == PEEKED_SINGLE_QUOTED -> peekedString = nextQuotedValue(SINGLE_QUOTE_OR_SLASH)
+      p == PEEKED_UNQUOTED -> peekedString = nextUnquotedValue()
+      p != PEEKED_BUFFERED -> throw JsonDataException("Expected a double but was " + peek() + " at path " + path)
+    }
```
JakeWharton, line 692:
```suggestion
    peekedString = when {
      p == PEEKED_NUMBER -> buffer.readUtf8(peekedNumberLength.toLong())
      p == PEEKED_DOUBLE_QUOTED -> nextQuotedValue(DOUBLE_QUOTE_OR_SLASH)
      p == PEEKED_SINGLE_QUOTED -> nextQuotedValue(SINGLE_QUOTE_OR_SLASH)
      p == PEEKED_UNQUOTED -> nextUnquotedValue()
      p != PEEKED_BUFFERED -> throw JsonDataException("Expected a double but was " + peek() + " at path " + path)
    }
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+    peeked = PEEKED_BUFFERED
+    val result: Double = try {
+      peekedString!!.toDouble()
```
JakeWharton, line 695: We can eliminate this `!!` by holding the result of the `when` in a local variable as well as writing it to `peekedString`.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1169 @@
...
+    }
+
+    // Advance the path and clear peeked if we haven't already.
+    if (peeked != PEEKED_NONE) {
+      pathIndices[stackSize - 1]++
+      peeked = PEEKED_NONE
+    }
+    valueSource = JsonValueSource(source, prefix, state, valueSourceStackSize)
+    pushScope(JsonScope.STREAMING_VALUE)
+    return valueSource!!.buffer()
```
JakeWharton, line 972: We can eliminate this `!!` by holding the new instance in a local variable as well as writing it to `valueSource`.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JvmName: fixed in 730b6f7

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+   * If non-null, the most recent value read was [.readJsonValue]. The caller may be
+   * mid-stream so it is necessary to call [JsonValueSource.discard] to get to the end of the
+   * current JSON value before proceeding.
+   */
+  private var valueSource: JsonValueSource? = null
+
+  constructor(source: BufferedSource?) {
+    if (source == null) {
+      throw NullPointerException("source == null")
+    }
```
JakeWharton, line 58:
```suggestion
  constructor(source: BufferedSource) {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+  /** The input JSON.  */
+  private val source: BufferedSource
+  private val buffer: Buffer
+  private var peeked = PEEKED_NONE
+
+  /**
+   * A peeked value that was composed entirely of digits with an optional leading dash. Positive
+   * values may not have a leading 0.
+   */
+  private var peekedLong: Long = 0
```
JakeWharton, line 37: nit:
```suggestion
  private var peekedLong = 0L
```
The lack of a type on the `peekedNumberLength` stands out to me. Want both to either specify the type or not.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+    }
+    return p != PEEKED_END_OBJECT && p != PEEKED_END_ARRAY && p != PEEKED_EOF
+  }
+
+  @Throws(IOException::class)
+  override fun peek(): Token {
+    var p = peeked
+    if (p == PEEKED_NONE) {
+      p = doPeek()
+    }
```
JakeWharton, line 157: In a follow-up we should extract this to a `private inline fun`. It appears like 30 times in this file and will eliminate the usage of `var` at all the call-sites.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+          peeked = PEEKED_NULL
+          peeked
+        }
+        else -> throw syntaxError("Unexpected value")
+      }
+      '\'' -> {
+        checkLenient()
+        buffer.readByte() // Consume '\''.
+        peeked = PEEKED_SINGLE_QUOTED
+        return peeked
```
JakeWharton, line 300: I know we changed this pattern from `return PEEKED_WHATEVER.also { peeked = it }`, but the new format (which I prefer visually) may be slower because it re-reads the value from the field rather than reusing the local. It may actually be worth switching back since performance is of the utmost importance at this layer.

@swankjesse opinions?

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+
+    // We've found the keyword followed either by EOF or by a non-literal character.
+    buffer.skip(length.toLong())
+    peeked = peeking
+    return peeked
+  }
+
+  @Throws(IOException::class)
+  private fun peekNumber(): Int {
+    var value: Long = 0 // Negative to accommodate Long.MIN_VALUE more easily.
```
JakeWharton, line 385:
```suggestion
    var value = 0L // Negative to accommodate Long.MIN_VALUE more easily.
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+    var negative = false
+    var fitsInLong = true
+    var last = NUMBER_CHAR_NONE
+    var i: Long = 0
```
JakeWharton, line 389:
```suggestion
    var i = 0L
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+        } catch (ignored: NumberFormatException) {
+          // Fall back to parse as a BigDecimal below.
+        }
+      }
+      p != PEEKED_BUFFERED -> {
+        throw JsonDataException("Expected a long but was " + peek() + " at path " + path)
+      }
+    }
+    peeked = PEEKED_BUFFERED
+    val result: Long = try {
```
JakeWharton, line 756:
```suggestion
    val result = try {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+  private fun nextNonWhitespace(throwOnEof: Boolean): Int {
+    /*
+     * This code uses ugly local variables 'p' and 'l' representing the 'pos'
+     * and 'limit' fields respectively. Using locals rather than fields saves
+     * a few field reads for each whitespace character in a pretty-printed
+     * document, resulting in a 5% speedup. We need to flush 'p' to its field
+     * before any (potentially indirect) call to fillBuffer() and reread both
+     * 'p' and 'l' after any (potentially indirect) call to the same method.
+     */
+    var p: Long = 0
```
JakeWharton, line 1007:
```suggestion
    var p = 0L
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+    while (true) {
+      val index = source.indexOfElement(runTerminator)
+      if (index == -1L) throw syntaxError("Unterminated string")
+      if (buffer[index].asChar() == '\\') {
+        buffer.skip(index + 1)
+        readEscapeCharacter()
+      } else {
+        buffer.skip(index + 1)
+        return
+      }
```
JakeWharton, line 825:
```suggestion
      val terminator = buffer[index].asChar()
      buffer.skip(index + 1)
      if (terminator == '\\') {
        readEscapeCharacter()
      } else {
        return
      }
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+          pathIndices[stackSize - 1]++
+          return result
+        } catch (ignored: NumberFormatException) {
+          // Fall back to parse as a double below.
+        }
+      }
+      p != PEEKED_BUFFERED -> throw JsonDataException("Expected an int but was ${peek()} at path $path")
+    }
+    peeked = PEEKED_BUFFERED
+    val asDouble: Double = try {
```
JakeWharton, line 867:
```suggestion
    val asDouble = try {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+    /*
+     * This code uses ugly local variables 'p' and 'l' representing the 'pos'
+     * and 'limit' fields respectively. Using locals rather than fields saves
+     * a few field reads for each whitespace character in a pretty-printed
+     * document, resulting in a 5% speedup. We need to flush 'p' to its field
+     * before any (potentially indirect) call to fillBuffer() and reread both
+     * 'p' and 'l' after any (potentially indirect) call to the same method.
+     */
+    var p: Long = 0
+    while (source.request((p + 1))) {
```
JakeWharton, line 1008:
```suggestion
    while (source.request(p + 1)) {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+      val c = buffer[p++].asChar()
+      when (c) {
+        '\n', ' ', '\r', '\t' -> continue
+      }
+      buffer.skip((p - 1))
```
JakeWharton, line 1013:
```suggestion
      buffer.skip(p - 1)
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+   * Advances the position until after the next newline character. If the line is terminated by
+   * "\r\n", the '\n' must be consumed as whitespace by the caller.
+   */
+  @Throws(IOException::class)
+  private fun skipToEndOfLine() {
+    val index = source.indexOfElement(LINEFEED_OR_CARRIAGE_RETURN)
+    buffer.skip(if (index != -1L) index + 1 else buffer.size)
+  }
+
+  /** Skips through the next closing block comment.  */
```
JakeWharton, line 1076:
```suggestion
  /** Skips through the next closing block comment. */
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+        peeking = PEEKED_NULL
+      }
+      else -> return PEEKED_NONE
+    }
+
+    // Confirm that chars [1..length) match the keyword.
+    val length = keyword.length
+    for (i in 1 until length) {
+      val i_long = i.toLong()
+      if (!source.request((i_long + 1))) {
```
JakeWharton, line 365:
```suggestion
      if (!source.request(i_long + 1)) {
```

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1186 @@
...
+
+  @Throws(IOException::class)
+  private fun peekNumber(): Int {
+    var value: Long = 0 // Negative to accommodate Long.MIN_VALUE more easily.
+    var negative = false
+    var fitsInLong = true
+    var last = NUMBER_CHAR_NONE
+    var i: Long = 0
+    while (true) {
+      if (!source.request((i + 1))) {
```
JakeWharton, line 391:
```suggestion
      if (!source.request(i + 1)) {
```

Bbnature99 (approved):

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JvmName: 🤦🏾 clearly I missed some stuff, thank you

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JvmName: is there a way to benchmark this? I don't think Moshi has any benchmarking code (besides #773)? 

Obviously I don't mind reverting this change -- mostly asking out of curiosity

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
JvmName: done in 98fdad3

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1115 @@
...
+    // Make sure our buffer has as many bytes as the source's buffer. This is necessary because
+    // JsonUtf8Reader assumes any data it has peeked (like the peekedNumberLength) are buffered.
+    try {
+      sourcePeek.require(copyFrom.buffer.size)
+    } catch (e: IOException) {
+      throw AssertionError()
+    }
+  }
+
+  @Throws(IOException::class)
```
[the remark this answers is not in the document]
JakeWharton: This type is not public so no one sees these. They're just leftover noise from the J2K converter.

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1115 @@
...
+  private fun skipUnquotedValue() {
+    val i = source.indexOfElement(UNQUOTED_STRING_TERMINALS)
+    buffer.skip(if (i != -1L) i else buffer.size)
+  }
+
+  @Throws(IOException::class)
+  override fun nextInt(): Int {
+    val p = peekIfNone()
+    var result: Int
```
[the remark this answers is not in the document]
JvmName: oh nice! I looked through the code, I don't think I saw any other instances where this could be applied?

moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.kt
```diff
@@ -0,0 +1,1082 @@
...
+   * Advances the position until after the next newline character. If the line is terminated by
+   * "\r\n", the '\n' must be consumed as whitespace by the caller.
+   */
+  private fun skipToEndOfLine() {
+    val index = source.indexOfElement(LINEFEED_OR_CARRIAGE_RETURN)
+    buffer.skip(if (index != -1L) index + 1 else buffer.size)
+  }
+
+  /** Skips through the next closing block comment. */
+  private fun skipToEndOfBlockComment(): Boolean {
```
codefromthecrypt, line 968: this is a drive by thing and it is not new to the port, but in case you want to solve it, this doesn't work if someone nested C-style comments. ex `[/* comment /* nested */ */ true]`

JakeWharton: Comments already aren't part of the standard so I'm not sure supporting that is necessary.

PR #1405, opened 2021-10-22
Only assign setters if present

ZacSweers:
Rather than generating

```kotlin
val result: Content
result = Content()
result.content = if (contentSet) content else result.content
result.text = if (textSet) text else result.text
```

This will now conditionally set

```kotlin
val result: Content
result = Content()
if (contentSet) {
  result.content = content
}
if (textSet) {
  result.text = text
}
```


swankjesse (commented):
# 👍🏻

LGTM

PR #1407, opened 2021-10-22
Parameterize kotlin test infra on CI

ZacSweers:
Before - our `DualKotlinTest` never actually ran with the `KotlinJsonAdapterFactory` due to there being no easy way to _force_ it to run even when there are generated adapters. This instead splits out codegen-only tests and wires a matrix in CI to test this instead, allowing us to cover the reflect case by _not_ generating code in that matrix on CI.

Fixed a an issue along the way with handling inline classes. We still don't formally support them, but this test has been helping us get a feel for what that would look like.

### DIFF
--- a/kotlin/tests/codegen-only/src/test/kotlin/com/squareup/moshi/kotlin/codegen/MixingReflectAndCodeGen.kt
+++ b/kotlin/tests/codegen-only/src/test/kotlin/com/squareup/moshi/kotlin/codegen/MixingReflectAndCodeGen.kt
@@ -32,10 +32,10 @@ class MixingReflectAndCodeGen {
     val reflectionAdapter = moshi.adapter<UsesReflectionAdapter>()
 
     assertThat(generatedAdapter.toString())
-      .isEqualTo("GeneratedJsonAdapter(KotlinJsonAdapterTest.UsesGeneratedAdapter).nullSafe()")
+      .isEqualTo("GeneratedJsonAdapter(MixingReflectAndCodeGen.UsesGeneratedAdapter).nullSafe()")
     assertThat(reflectionAdapter.toString())
       .isEqualTo(
-        "KotlinJsonAdapter(com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest" +
+        "KotlinJsonAdapter(com.squareup.moshi.kotlin.codegen.MixingReflectAndCodeGen" +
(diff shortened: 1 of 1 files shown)


kotlin/tests/codegen-only/build.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+  REFLECT, KAPT, KSP
+}
+
+val testMode = findProperty("kotlinTestMode")?.toString()
+  ?.let(TestMode::valueOf)
+  ?: KSP
+
+when (testMode) {
+  REFLECT -> {
+    // Default to KSP. This is a CI-only thing
```
swankjesse, line 38: Nice.

kotlin/tests/codegen-only/src/test/kotlin/com/squareup/moshi/kotlin/codegen/MixingReflectAndCodeGen.kt
```diff
@@ -0,0 +1,48 @@
...
+package com.squareup.moshi.kotlin.codegen
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.moshi.JsonClass
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.adapter
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import org.junit.Test
+
+class MixingReflectAndCodeGen {
```
swankjesse, line 25: # 🔥

kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
```diff
@@ -366,19 +326,22 @@ class DualKotlinTest(useReflection: Boolean) {
...
+        {"text":"root","r":{"number":0,"r":{"text":"grand child 1"},"t":{"number":1}},"t":{"text":"child 1"}}
         """.trimIndent()
       )
   }
 
   @JsonClass(generateAdapter = true)
   open class Node<T : Node<T, R>, R : Node<R, T>> {
-    var t: T? = null
+    // kotlin-reflect doesn't preserve ordering, so put these in alphabetical order so that
+    // both reflective and code gen tests work the same
```
swankjesse, line 342: Great comments.

swankjesse (commented):
LGTM!

PR #1410, opened 2021-10-25
Switch from internal to experimental internal annotation

ZacSweers:
This is functionally no different than before, but allows for external use of these APIs for consumers that are willing to forgo support/API stability. This allows use of these processing-agnostic APIs in multiple contexts while reusing the same logic Moshi uses today for generating adapters.


JvmName (approved):

swankjesse:
LGTM

PR #1411, opened 2021-10-27
Add JSON Decode Ordering Parity

hardwil:
This change is replacing `LinkedHashMap` with `TreeMap` in `KotlinJsonAdapter` to match the ordering in the `ClassJsonAdapter`, so that both Java and Kotlin instances will have ordering parity.

When migrating code from Java to Kotlin, I found that some tests failed due to entry ordering changes between the respective adapters.


swankjesse:
I don't like that ClassJsonAdapter needs to sort fields either; I think we do that for determinism across JVMs. Retaining field order seems like the nicest user interface.

Also note that fixing this would be very invasive to anyone upgrading.

swankjesse:
Original PR that sorted Java classes:
https://github.com/square/moshi/pull/19

ZacSweers:
Do we try to offer an opt-out temporarily to ease migration ([kinda like what coroutines does](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.6.0-RC))

ZacSweers (commented):
Let's do this in ClassJsonAdapter as well and add a system property to restore the previous behavior as an escape hatch for people that need it. @hardwil would you be up for doing that? If not I can continue this work too

PR #1412, opened 2021-10-30
Prototype support for Optional

ZacSweers:
This is a prototype of what `Optional` support would look like. This implements it in two parts:
1. A new `OptionalJsonAdapter` + associated factory. This essentially works by associating nulls with `Optional.absent()` and used for properties that are either present or explicitly `null`.
2. Reflective support in `ClassJsonAdapter` and `KotlinJsonAdapter` for supporting defaulting absent values to `Optional.empty()`.

Open questions:
1. Do we really want to support this in Kotlin? If so, how do we handle the following:
  a. `Optional<String>?`
  b. Do we support this in code gen too?
2. Do we support just `j.u.Optional` or do we want to do something more generic (to support Guava, KOptional, etc)? How would that look if so?
3. Do we want to generify absent properties in general? Maybe an optional `JsonAdapter.ifAbsent()` or separate `AbsenceHandler` interface?
  - This would obviate 1 and 2 as we just need to support this API and delegate to whatever adapters folks bring.

### DIFF
--- a/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
+++ b/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt
@@ -28,6 +28,7 @@ import com.squareup.moshi.internal.Util.resolve
 import com.squareup.moshi.rawType
 import java.lang.reflect.Modifier
 import java.lang.reflect.Type
+import java.util.Optional
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
 import kotlin.reflect.KMutableProperty1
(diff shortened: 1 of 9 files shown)


ZacSweers:
Pushed an absence handler API demo instead, which I think came out pretty clean all things considered!

ZacSweers:
Should say I'm not thrilled about the of adding this directly to the JsonAdapter API, but without a clear/easy way to access delegated adapters it's not clear how to unpack an opt-in interface's APIs

lassem:
In Kotlin I'd use this to fall back to any default arguments on data classes. In our existing implementation this is achieved via a work around utilising a custom `@DefaultIfNull` annotation and deserialising via `Map` first, which of course is additional overhead.

lassem:
That is true. However, the aforementioned `@DefaultIfNull` annotation would also use the default if the json specifically said `null`

ZacSweers:
I don't see how it would since an absent key would not be present in a map. In any case let's keep this PR on topic

PR #1416, opened 2021-11-03
Support externally-compiled transient property recognition in KSP

ZacSweers:
For now it doesn't work due to https://github.com/google/ksp/issues/710, but it will eventually work once that's fixed. This puts in a check and a toe-hold comment in tests for it


JvmName (approved):

swankjesse (commented):
LGTM

Issue #1301, opened 2021-02-12
Moshi needs an @JsonIgnore annotation, so fields can be independently excluded from serialization


bpappin:
**tl;dr** Using _transient_ keyword to ignore fields causes side effects, and is impossible in a kotlin data class. Moshi needs a _@JsonIgnore_ annotation similar to Rooms _@Ignore_, or similar to the inverse functionality found with Gson _@Expose_, so that we can isolate fields Moshi should ignore from the rest of the code, and avoid fields marked _transient_ from being interpreted unexpectedly. 

In order to ignore a field in Moshi, the transient keyword is used.
However this causes side effects with other libraries that also interpret the transient keyword.

Kotlin example:
Consider a data class that can be loaded from a remote source, deserialized through moshi, but also has a local representation stored by Room:

```
@Entity(tableName = "my_data")
@JsonClass(generateAdapter = true)
data class MyData(
    @PrimaryKey(autoGenerate = true)
    @ColumnInfo(name = "local_id") 
    @Json(name = "remote_id") 
    val id:String, 
   
    @ColumnInfo(name = "local_name")
    @Json(name = "remote_name")   
    val name:String, 

    @ColumnInfo(name = "local_url") 
    // should be ignored by Moshi in both directions
    var url:String? = "http://example.com"
)
```

In a Java class similar to that Kotlin data class, the "url" field would need a _transient_ qualifier, **which would also cause Room to ignore it**.

**All of these problems can be alleviated by including a different mechanism for ignoring fields, such as including a _@JsonIgnore_ annotation.**

ZacSweers:
The right solution is not to mix your domain models. It can be a convenient shortcut, but when they incur confusing and asymmetric side effects you end up in situations like this. 

We have considered adding a transient option to `@Json` for computed properties, but wouldn't do it for the domain conflation scenario you've described.

bpappin:
First of all, I disagree, 80% of this is about maintenance, and how a model is structures is not defined by Moshi. 

It's not for Moshi to determine the internal architecture of an application, be it a new greenfield build or some legacy project that someone needs to work with. 

**The fact of the matter is that Moshi should not be using "transient" to ignore fields.**

it's extremely lame that this FR was close without properly evaluating the issues involved.

bpappin:
The fact is that sometimes properties need to be hidden from Moshi without using the transient keyword.

JakeWharton:
[...]

Issue #1262, opened 2020-10-23
Conflicts using `transient` (it is also used by other libraries)


goa:
As mentioned in the last comments of [this issue](https://github.com/square/moshi/issues/94), other libraries such as Realm and Cupboard are also using `transient` to exclude fields.

What if we need a field to be excluded from JSON serialization, but nevertheless be written to the Realm database?
If `transient` is used, the field will be excluded from both.

I wonder if Moshi could use another dedicated annotation to exclude fields, or at least an annotation to force inclusion of `transient` fields.

JakeWharton:
It's not a good idea to mix model objects across serialization formats. You either should persist the raw JSON or split the models for the networking layer and the persistence layer. Changes in schema shouldn't affect JSON serialization. Changes in JSON format shouldn't affect schema.

goa:
While I agree that that mixing model objects is not the best approach, it's usually much less work when developing smaller to medium apps. That is possible for example with GSON and Realm, since GSON has its own approach for including / excluding fields, but it is impossible with Moshi and Realm since both libraries rely on `transient`.

While I understand that `transient` is The Right Way, I believe that adding an alternative way to exclude (or force-include transient) fields would be helpful in many situations to save code and increase adoption of Moshi in older projects. 

I guess that if there are no other requests for this, the issue could be closed.

davidcode43w:
I'm really holded back in GSON becouse of this missing feature on Moshi, and I don't want to be like this :'(

Also I would like to have more controll over serialization and deserialization differences at runtime, for example, I work with a restfull API that offer data with ID fields, I want to deserialize this field for sync propouses but I don't want to serialize an ID when I post a new object to the API becouse it'll be useless, the API will override the value for obvious reasons.

 According to the documentation: 
> Transient fields are omitted when writing JSON. When reading JSON, the field is skipped even if the JSON contains a value for the field. Instead it will get a default value.
>
So my `@Exclude(serialize=false, deserialize=true)` can't be safely migrated to Moshi.

ZacSweers:
You should solve this case in your code base by implementing custom Moshi json adapters for types that need special behavior for different properties.

davidcode43w:
> You should solve this case in your code base by implementing custom Moshi json adapters for types that need special behavior for different properties.

Yes but I'll have to write the serializer by myself in all my datatypes so why do I use Moshi anyway, the idea is to save all the work...

davidcode43w:
> 
> 
[...]

PR #1417, opened 2021-11-05
Add `@Json.ignore` 

### LINKED ISSUE #609: Exclude property without backing field from deserialization

### LINKED ISSUE #1151: Codegen Adapters trying to set properties without backing fields

### LINKED ISSUE #1259: Kotlin codegen causes compile error for private computed val

ZacSweers:
This introduces a new `ignore` property to the `@Json` annotation. It's semantically similar to `transient`, but without being JVM-specific or relying on backing fields.

```kotlin
class Example(...) {
  @Json(ignore = true)
  val someComputedProperty: String get() = "hi!"
}
```

This property defaults to `false` and also makes `Json.name` no longer required, instead opting for a "known default value" that we check against to see if it's been set before.

This resolves a number of outstanding issues around handling transience in Kotlin in particular!
* resolves #609
* resolves #1259
* resolves #1151
* Also helps with past issues like #1301, #1262, #1153

It is not supported in record classes or enums, which cannot reasonably support this. Enums could possibly support it, but I couldn't find an obvious example of how this is handled in other parsers and not sure what the right behavior should be.

This adds support in reflective Java, reflective Kotlin, kapt code gen, and KSP code gen. Note that `@Transient`/`transient` are still recognized as well.

### DIFF
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
@@ -510,7 +510,7 @@ private fun List<AnnotationSpec>?.qualifiers(
 
 private fun List<AnnotationSpec>?.jsonName(): String? {
   if (this == null) return null
-  return find { it.typeName == JSON }?.let { annotation ->
+  val name = find { it.typeName == JSON }?.let { annotation ->
     val mirror = requireNotNull(annotation.tag<AnnotationMirror>()) {
       "Could not get the annotation mirror from the annotation spec"
     }
@@ -518,6 +518,7 @@ private fun List<AnnotationSpec>?.jsonName(): String? {
       it.key.simpleName.contentEquals("name")
     }.value.value as String
   }
+  return name?.takeUnless { it == Json.UNSET_NAME }
 }
 
 private fun String.escapeDollarSigns(): String {
--- a/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
+++ b/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
@@ -64,6 +64,7 @@ import javax.tools.Diagnostic.Kind.WARNING
 
 private val JSON_QUALIFIER = JsonQualifier::class.java
 private val JSON = Json::class.asClassName()
+private val TRANSIENT = Transient::class.asClassName()
 private val OBJECT_CLASS = ClassName("java.lang", "Object")
 private val VISIBILITY_MODIFIERS = setOf(
   KModifier.INTERNAL,
@@ -95,7 +96,8 @@ internal fun primaryConstructor(
       type = parameter.type,
       hasDefault = parameter.defaultValue != null,
       qualifiers = parameter.annotations.qualifiers(messager, elements),
-      jsonName = parameter.annotations.jsonName()
+      jsonName = parameter.annotations.jsonName(),
+      jsonIgnore = parameter.annotations.jsonIgnore(),
     )
   }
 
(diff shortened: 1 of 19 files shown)


kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
```diff
@@ -510,15 +514,34 @@ private fun List<AnnotationSpec>?.qualifiers(
...
+  return elementValue<String>("name").takeUnless { it == Json.UNSET_NAME }
+}
+
+private fun AnnotationSpec.jsonIgnore(): Boolean {
+  return elementValue<Boolean>("ignore") ?: false
+}
+
+private fun <T> AnnotationSpec.elementValue(name: String): T? {
+  val mirror = requireNotNull(tag<AnnotationMirror>()) {
+    "Could not get the annotation mirror from the annotation spec"
```
Egorand, line 539: Should we print any diagnostics data here to make it easier to triage, or is it a "this should never happen" case?

Egorand (approved):
Nice!

kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/apt/metadata.kt
ZacSweers: The latter!

moshi/src/main/java/com/squareup/moshi/Json.java
```diff
@@ -39,5 +40,17 @@
...
+  String UNSET_NAME = "\u0000";
+
+  /** The name of the field when encoded as JSON. */
+  String name() default UNSET_NAME;
+
+  /**
+   * If true, this field/property will be ignored. This is semantically similar to use of {@code
+   * transient} on the JVM.
+   *
+   * <p><strong>Note:</strong> this has no effect in enums or record classes.
```
swankjesse, line 53: 👍🏻

swankjesse (commented):
LGTM!

PR #1422, opened 2021-11-15
Update to KSP 1.0.1 and use new jvm modifiers resolver API

ZacSweers:
This allows us to fully support transient across compilation boundaries

### DIFF
--- a/kotlin/tests/extra-moshi-test-module/src/main/kotlin/com/squareup/moshi/kotlin/codegen/test/extra/AbstractClassInModuleA.kt
+++ b/kotlin/tests/extra-moshi-test-module/src/main/kotlin/com/squareup/moshi/kotlin/codegen/test/extra/AbstractClassInModuleA.kt
@@ -18,9 +18,9 @@ package com.squareup.moshi.kotlin.codegen.test.extra
 import com.squareup.moshi.Json
 
 public abstract class AbstractClassInModuleA {
-  // Ignored to ensure processor sees them across module boundaries.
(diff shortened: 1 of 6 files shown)


ZacSweers:
KCT needs to update https://github.com/tschuchortdev/kotlin-compile-testing/issues/222

Egorand (approved):

ZacSweers:
Copying KSP support for KCT for now as they are lagging in updates

Egorand (approved):

ZacSweers:
I think we need to wait for KSP 1.6.0-1.0.1 because it looks like the new effective java modifiers API is not in the 1.0.1-RC

kotlin/tests/extra-moshi-test-module/src/main/kotlin/com/squareup/moshi/kotlin/codegen/test/extra/AbstractClassInModuleA.kt
```diff
@@ -18,9 +18,9 @@ package com.squareup.moshi.kotlin.codegen.test.extra
 import com.squareup.moshi.Json
 
 public abstract class AbstractClassInModuleA {
-  // Ignored to ensure processor sees them across module boundaries.
-  // @Transient doesn't work for this case because it's source-only and jvm modifiers aren't currently visible in KSP.
```
swankjesse, line 22: yay

swankjesse (commented):
LGTM

PR #1425, opened 2021-11-16
Upgrade to Kotlin 1.6 + always instantiate annotations

ZacSweers:
This fully upgrades to Kotlin 1.6 and removes the controls for whether or not to instantiate annotations. Instead, this now just requires Kotlin 1.6 going forward.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,19 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kotlin-version: [ '1.5.31', '1.6.0-RC' ]
-        # TODO add back KSP 1.6.0-1.0.1 once it's out
-        ksp-version: [ '1.5.31-1.0.1' ]
         kotlin-test-mode: [ 'REFLECT', 'KSP', 'KAPT' ]
-        exclude:
-#          - kotlin-version: '1.5.31'
-#            ksp-version: '1.6.0-RC-1.0.1-RC'
-          - kotlin-version: '1.6.0-RC'
-            ksp-version: '1.5.31-1.0.1'
-
-    env:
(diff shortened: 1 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -48,10 +36,10 @@ jobs:
           java-version: '17'
 
       - name: Test
-        run: ./gradlew build check --stacktrace -PkotlinTestMode=${{ matrix.kotlin-test-mode }} -PkotlinVersion=${{ matrix.kotlin-version }}
+        run: ./gradlew build check --stacktrace -PkotlinTestMode=${{ matrix.kotlin-test-mode }}
 
       - name: Publish (default branch only)
-        if: github.repository == 'square/moshi' && github.ref == 'refs/heads/master' && matrix.kotlin-version == '1.5.31' && matrix.kotlin-test-mode == 'reflect'
+        if: github.repository == 'square/moshi' && github.ref == 'refs/heads/master' && matrix.kotlin-version == '1.6.0' && matrix.kotlin-test-mode == 'reflect'
```
Egorand, line 42: Can remove `matrix.kotlin-version == '1.6.0'` since it's not in the matrix anymore?

Egorand (approved):
Nice cleanup! 👍

.github/workflows/build.yml
ZacSweers: Ah good catch, will fix

ZacSweers, line 42:
```suggestion
        if: github.repository == 'square/moshi' && github.ref == 'refs/heads/master' && matrix.kotlin-test-mode == 'reflect'
```

n0m0r3pa1n:
Any ideas when this will be available in maven?

swankjesse (commented):
Love to see simplifications like this.

Issue #1415, opened 2021-11-02
Missing required field within a JSON array exception message has incorrect off-by-one path.


behrozsaadat:
When trying to deserialize JSON into a Data class with a missing required parameter inside of an Array, the path that is output in the message is off-by-one.

Test reproducing the problem: (I wrote the test inside of the `DualKotlinTest` file in Moshi)

https://gist.github.com/behrozsaadat/b49e250c8ff47e6bfeee0f388b64c144

ZacSweers:
> the path that is output in the message is off-by-one.

Can you add an example of the message you get from the given repro?

behrozsaadat:
> > the path that is output in the message is off-by-one.
> 
> Can you add an example of the message you get from the given repro?

Added to the original post.

ZacSweers:
Seems simple enough to fix, want to send a PR?

behrozsaadat:
> Seems simple enough to fix, want to send a PR?

I took a stab at fixing this but it seems to be a bit more complicated than I thought. 

Naively changing [this line](https://github.com/behrozsaadat/moshi/blob/538890e8c0a8178913030d80af306707ce67cb67/moshi/src/main/java/com/squareup/moshi/JsonScope.java#L63) to:

```
result.append('[').append(pathIndices[i] - 1).append(']');
```

Seems to have caused breakages in 100+ other tests. Stepping through the `pathIndices` logic doesn't reveal and immediately obvious bugs either.

SimonMarquis:
Unfortunately, this seems to be the expected behavior.
In [JsonUtf8Reader.endObject()](https://github.com/square/moshi/blob/aaadbf496354c5650202377d100a9b420f0cd26e/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java#L176-L191) and [JsonUtf8Reader.endArray()](https://github.com/square/moshi/blob/aaadbf496354c5650202377d100a9b420f0cd26e/moshi/src/main/java/com/squareup/moshi/JsonUtf8Reader.java#L146-L159), we bump the current path index with `pathIndices[stackSize - 1]++;`.

[...]

PR #1435, opened 2021-11-28
Fix unexpected path when a value is absent from an object

SimonMarquis:
Keep track of the current object path to format the JsonDataException message.

See square/moshi#1415


SimonMarquis:
Oups, the 1st push had a typo: https://github.com/square/moshi/compare/32d694751e555755b88884f3b3f1e4e0d8939041..7043605568338f8c3fbe298a32e4dd02626fb493

SimonMarquis:
I guess this change must also be applied on the AdapterGenerator as well. I'll try to do that.

swankjesse (commented):
Great PR. This is ALMOST perfect. But I have one reservation and it’s a big one. 

`JsonReader.getPath()` is a big slow function and this adds that function to the inner loop of every single object parse. The only sufficiently-performant fix here gets the path only if we actually encounter a problem, and not otherwise.

Unfortunately at the time we know we have a problem it’s difficult to recover that path.

But can we do it anyway? I think we probably can! We need to get the path immediately before calling endObject().

swankjesse:
Here’s a work-in-progress that gets your test passing without precomputing the path. It does introduce more work in the common case, but no extra allocations or method calls.
https://github.com/square/moshi/compare/master...jwilson.1130.more_accurate_paths

SimonMarquis:
I completely get that we should not impact the critical path.
Unfortunately I don't think your version working by the look at the unit tests results, or am I missing something here?

swankjesse:
@SimonMarquis 

My goal is to introduce a behavior change here, which means fixing up a bunch of tests. Currently a JsonReader’s path represents the _previous_ name after its value has been read. This behavior is verified by this test, that is failing on my commit:

```
    JsonReader reader = newReader("{\"a\": 5, \"b\": 5, \"c\": 5, \"d\": 5}");
    reader.beginObject();
    assertEquals("$.", reader.getPath());

    assertEquals(0, reader.selectName(abc));
    assertEquals("$.a", reader.getPath());
    assertEquals(5, reader.nextInt());
    assertEquals("$.a", reader.getPath()); // path is still 'a', even though a’s value has been processed
```

With my experiment we’d introduce a behavior change, where the path would drop the name after its value has been read. We’d need to change the tests to match the new behavior. The test above would change like so:

```
    JsonReader reader = newReader("{\"a\": 5, \"b\": 5, \"c\": 5, \"d\": 5}");
    reader.beginObject();
    assertEquals("$.", reader.getPath());

    assertEquals(0, reader.selectName(abc));
    assertEquals("$.a", reader.getPath());
    assertEquals(5, reader.nextInt());
    assertEquals("$.", reader.getPath()); // different!
```

I’m okay with this, especially since it improves the missing field error.

SimonMarquis:
Thank you for clarifying that.
I'll try to update the remaining tests and JsonReaders.

SimonMarquis:
How do you see the path being affected by `JsonReader.skipName()` and `JsonReader.promoteNameToValue()`?
I'm not sure how it should behave, wether we should ignore the value or simulate it and push it to the path stack.

SimonMarquis:
I can see two behaviors at the moment:

- `JsonValueReader`:
  ```kotlin 
  @Test
  public void promoteNameToValue() throws Exception {
    JsonReader reader = factory.newReader("{\"a\":1}");
    reader.beginObject();
    reader.promoteNameToValue();
    assertThat(reader.getPath()).isEqualTo("$.a");
    assertThat(reader.peek()).isEqualTo(JsonReader.Token.STRING);
    assertThat(reader.nextString()).isEqualTo("a");
    assertThat(reader.getPath()).isEqualTo("$.a"); /* 👀 */
    /* ... */
  }
  ```

- `JsonUtf8Reader`:
  ```kotlin 
  @Test
  public void promoteNameToValue2() throws Exception {
    JsonReader reader = factory.newReader("{\"a\":1}");
    reader.beginObject();
    reader.promoteNameToValue();
    assertThat(reader.getPath()).isEqualTo("$.a");
    assertThat(reader.peek()).isEqualTo(JsonReader.Token.STRING);
    assertThat(reader.nextString()).isEqualTo("a");
    assertThat(reader.getPath()).isEqualTo("$."); /* 👀 */
    /* ... */
  }
  ```

swankjesse:
@SimonMarquis ooooh good question, and I see how this is problematic. I think this behavior is nicer:

```
  assertThat(reader.getPath()).isEqualTo("$.a"); /* 👀 */
```

I prefer that because if there’s a problem **within** the value for `a` we want the path to indicate that. I’m not sure if that’s super tricky to achieve!

SimonMarquis:
There is one remaining class of issues I'm facing with [this test](https://github.com/square/moshi/blob/3cbd5c327a0e2c1d150fdaa2c04202dfa82450e8/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt#L148-L158):

`KotlinJsonAdapter` correctly decodes a `null` value, but throws because it it must not be nullable. But here it's too late, the path already changed.

https://github.com/square/moshi/blob/3cbd5c327a0e2c1d150fdaa2c04202dfa82450e8/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt#L86-L94

We could fix that by first checking if the reader is peeking a NULL token, but it wouldn't solve other variants of this bug like in [this test](https://github.com/square/moshi/blob/3cbd5c327a0e2c1d150fdaa2c04202dfa82450e8/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt#L160-L179).
The token might not be NULL but the delegate could return a null value.

SimonMarquis:
> We could fix that by first checking if the reader is peeking a NULL token

Well actually I found out it's an already expected behavior to first let the delegate parse the value (see [delegatesToInstalledAdaptersBeforeNullChecking](https://github.com/square/moshi/blob/3cbd5c327a0e2c1d150fdaa2c04202dfa82450e8/kotlin/tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt#L190))

swankjesse:
@SimonMarquis in that specific case I don’t think the failure is ambiguous, that unexpected null exception tells you the JSON name that was null and the path to the object that enclosed it.

SimonMarquis:
I've updated the `JsonUtf8Reader` to correctly report the last `peekedString` to fix `promoteValueToName` tests.
But to be honest, I'm not confident at all with the solution I came up with.
I'd like to get your feedback on this. 🙏

cleidiano:
Hi everyone, this PR have been stalling for a while, any plan to merge it?

PR #1436, opened 2021-11-30
Omit adapter field names now that we directly instantiate annotations

ZacSweers:
This is no longer necessary!


bryanstern (approved):

swankjesse:
(this ended up reverted here: https://github.com/square/moshi/pull/1439)

PR #1442, opened 2021-12-04
Fix a crash when combining generics and records

swankjesse:
Without the fix the test fails like so:

    No JsonAdapter for T (with no annotations)
    for T
    for java.util.List<T>
    for com.squareup.moshi.records.RecordsTest$IndirectGenerics<java.lang.Long>
    for class com.squareup.moshi.records.RecordsTest$HasIndirectGenerics
    java.lang.IllegalArgumentException: No JsonAdapter for T (with no annotations)
    for T
    for java.util.List<T>
    for com.squareup.moshi.records.RecordsTest$IndirectGenerics<java.lang.Long>
    for class com.squareup.moshi.records.RecordsTest$HasIndirectGenerics
      at com.squareup.moshi.Moshi$LookupChain.exceptionWithLookupStack(Moshi.java:389)
      at com.squareup.moshi.Moshi.adapter(Moshi.java:158)
      at com.squareup.moshi.Moshi.adapter(Moshi.java:106)
      at com.squareup.moshi.Moshi.adapter(Moshi.java:75)

### DIFF
--- a/moshi/records-tests/src/test/java/com/squareup/moshi/records/RecordsTest.java
+++ b/moshi/records-tests/src/test/java/com/squareup/moshi/records/RecordsTest.java
@@ -157,6 +157,21 @@ public final class RecordsTest {
   }
 
   @Test
+  public void indirectGenerics() throws IOException {
+    var value =
+        new HasIndirectGenerics(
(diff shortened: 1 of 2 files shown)


ZacSweers (approved):
Nice catch. Just needs a spotlessApply fix

PR #1446, opened 2021-12-05
Don't mangle property names that contain a dollar sign

swankjesse:
Without this fix the new test fails like this:

    value of: toJson(...)
    expected: {"$a":"apple","$b":"banana"}
    but was : {"${'$'}a":"apple","$b":"banana"}


ZacSweers (approved):

swankjesse:
Ooooh tests show we also had this bug in Kapt! Yay bug compatibility!

PR #1449, opened 2021-12-06
Don't use dynamic proxies to read annotation values

swankjesse:
Including such a powerful mechanism is overkill for our narrow use cases.


ZacSweers (approved):

swankjesse:
Lets just wait for KSP to have a stable API for this.

PR #1450, opened 2021-12-09
Switch to the Vanniktech base plugin for publishing

swankjesse:
Move configuration out of build.gradle files and into build.gradle.kts files.
Sign published builds.
Support publishing release builds from GitHub actions.


settings.gradle.kts
```diff
@@ -26,10 +26,13 @@ include(":moshi")
 include(":moshi:japicmp")
 include(":moshi:records-tests")
 include(":adapters")
+project(":adapters").name = "moshi-adapters"
```
swankjesse, line 29: In a follow-up I’d like to move the files around so we don’t need this renaming hack.

build.gradle.kts
```diff
@@ -106,30 +112,54 @@ subprojects {
...
+      publishToMavenCentral(SonatypeHost.DEFAULT)
+      signAllPublications()
+      pom {
+        description.set("A modern JSON API for Android and Java")
+        name.set(project.name)
+        url.set("https://github.com/square/moshi/")
+        licenses {
+          license {
+            name.set("The Apache Software License, Version 2.0")
+            url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
```
ZacSweers, line 150:
```suggestion
            url.set("https://www.apache.org/licenses/LICENSE-2.0.txt")
```

build.gradle.kts
```diff
@@ -106,30 +112,54 @@ subprojects {
...
-        outputDirectory.set(rootDir.resolve("docs/1.x"))
-        dokkaSourceSets.configureEach {
-          skipDeprecated.set(true)
-          externalDocumentationLink {
-            url.set(URL("https://square.github.io/okio/2.x/okio/"))
+  plugins.withId("com.vanniktech.maven.publish.base") {
+    configure<MavenPublishBaseExtension> {
+      publishToMavenCentral(SonatypeHost.DEFAULT)
+      signAllPublications()
+      pom {
```
ZacSweers, line 143: FYI there's a helper function for loading these from the same properties as before if ya want

ZacSweers (approved):

build.gradle.kts
swankjesse: Things I hate from most to hate least:

 * Gradle Groovy builds. Many ways to do each thing. Mistakes fail silently. Not discoverable.
 * Gradle properties files. One way to do each thing. Fails silently. Not discoverable.
 * Gradle Kotlin builds. IntelliJ tells me when I make a mistake. Discoverable.

So configuring stuff without the safety net of a type checker? Not unless I have no other choice.

PR #1320, opened 2021-03-19
Convert moshi-adapters to Kotlin

ZacSweers:
Ref #1252

PR #1460, opened 2021-12-11
Convert adapters to Kotlin

ZacSweers:
Ref #1252

Continuation from #1320 with CR comments addressed


moshi-adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException
+import java.util.Date
+
+@Deprecated(
+  """This class moved to avoid a package name conflict in the Java Platform Module System.
+      The new class is com.squareup.moshi.adapters.Rfc3339DateJsonAdapter.""",
+  replaceWith = ReplaceWith("com.squareup.moshi.adapters.Rfc3339DateJsonAdapter"),
+  level = DeprecationLevel.ERROR
+)
+public class Rfc3339DateJsonAdapter : JsonAdapter<Date>() {
```
ZacSweers, line 28: Maybe we should drop this in Moshi 2.0?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * @receiver ISO string to parse in the appropriate format.
+ * @return the parsed date
+ */
+internal fun String.parseIsoDate(): Date {
+  return try {
+    var offset = 0
+
+    // extract year
+    val year = parseInt(this, offset, 4.let { offset += it; offset })
```
JvmName, line 84: > `4.let { offset += it; offset }`

oof, that takes a second to grok, but I can't think of a better oneliner 🤷🏾

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
```diff
@@ -0,0 +1,271 @@
...
+    var minutes = 0
+    var seconds = 0
+    // always use 0 otherwise returned date will include millis of current time
+    var milliseconds = 0
+
+    // if the value has no time component (and no time zone), we are done
+    val hasT = checkOffset(this, offset, 'T')
+    if (!hasT && this.length <= offset) {
+      val calendar: Calendar = GregorianCalendar(year, month - 1, day)
+      return calendar.time
```
JvmName, line 108: would it make sense to combine this as `GregorianCalendar(**).time`?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  override fun create(type: Type, annotations: Set<Annotation?>, moshi: Moshi): JsonAdapter<*>? {
+    if (type.rawType != baseType || annotations.isNotEmpty()) {
+      return null
+    }
+    val jsonAdapters: List<JsonAdapter<Any>> = List(subtypes.size) { i ->
+      moshi.adapter(subtypes[i])
+    }
```
JvmName, line 176: nit: is this `List` construction more idiomatic than `subtypes.map { moshi.adapter(it) }`?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt
```diff
@@ -0,0 +1,274 @@
...
+        val labelIndex = reader.selectString(labelOptions)
+        if (labelIndex == -1 && fallbackJsonAdapter == null) {
+          throw JsonDataException(
+            "Expected one of " +
+              labels +
+              " for key '" +
+              labelKey +
+              "' but found '" +
+              reader.nextString() +
+              "'. Register a subtype for this label."
```
JvmName, line 222: nit: would it make sense to interp this string?

JvmName (approved):

moshi-adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.kt
Egorand, line 28: Would it make sense to just do `by Rfc3339DateJsonAdapter()` here?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
```diff
@@ -0,0 +1,271 @@
+/*
+ * Copyright (C) 2011 Square, Inc.
```
Egorand, line 2: Original copyright was different - `Copyright (C) 2011 FasterXML, LLC`

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
Egorand: Would `run { offset += 4; offset }` work?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
```diff
@@ -0,0 +1,271 @@
...
+          /*
+           * 22-Jan-2015, tatu: Looks like canonical version has colons, but we may be given
+           *    one without. If so, don't sweat.
+           *   Yes, very inefficient. Hopefully not hit often.
+           *   If it becomes a perf problem, add 'loose' comparison instead.
+           */
+          val cleaned = act.replace(":", "")
+          if (cleaned != timezoneId) {
+            throw IndexOutOfBoundsException(
+              "Mismatching time zone indicator: " +
```
Egorand, line 168: nit: use string interpolation here?

moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt
```diff
@@ -0,0 +1,274 @@
+/*
+ * Copyright (C) 2011 Square, Inc.
```
Egorand, line 2: Original is `Copyright (C) 2011 Google Inc.`

moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt
```diff
@@ -0,0 +1,274 @@
...
+    /** Single-element options containing the label's key only.  */
+    private val labelKeyOptions: Options = Options.of(labelKey)
+
+    /** Corresponds to subtypes.  */
+    private val labelOptions: Options = Options.of(*labels.toTypedArray())
+
+    override fun fromJson(reader: JsonReader): Any? {
+      val peeked = reader.peekJson()
+      peeked.setFailOnUnknown(false)
+      val labelIndex = peeked.use(::labelIndex)
```
Egorand, line 197: Nice!

Egorand (approved):

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
JvmName: I thought about that, it's nice that `4.` is at the front of the `let` construction though

moshi-adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.kt
```diff
@@ -0,0 +1,111 @@
...
+
+  @Throws(IOException::class)
+  override fun fromJson(reader: JsonReader): T? {
+    val index = reader.selectString(options)
+    if (index != -1) return constants[index]
+    val path = reader.path
+    if (!useFallbackValue) {
+      val name = reader.nextString()
+      throw JsonDataException(
+        "Expected one of ${nameStrings.toList()} but was $name at path $path"
```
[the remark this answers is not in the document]
ZacSweers: We could but tbh we're throwing an exception here and I'm not worried about perf there vs just simple readability. Going to just remove toList() instead

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
ZacSweers: yeah it's a little wonky and one of those places that you can't write it the same way in kotlin

moshi-adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.kt
ZacSweers: Unfortunately can't since JsonAdapter is not an interface

moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt
ZacSweers: good catch

moshi-adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.kt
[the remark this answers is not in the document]
JvmName: uh it's an array so `.toString` will produce garbage without either `joinToString` or `toList` right?

PR #1461, opened 2021-12-11
Convert util package to Kotlin

ZacSweers:
Ref #1252


moshi-adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.java
```diff
@@ -68,7 +68,7 @@
       nameStrings = new String[constants.length];
       for (int i = 0; i < constants.length; i++) {
         String constantName = constants[i].name();
-        nameStrings[i] = jsonName(constantName, enumType.getField(constantName));
+        nameStrings[i] = jsonName(enumType.getField(constantName), constantName);
```
ZacSweers, line 71: I think this is a problem for code gen actually :(

ZacSweers: actually should be ok

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
```diff
@@ -0,0 +1,643 @@
...
+        val original = toResolve
+        val ownerType: Type? = original.ownerType
+        val newOwnerType = ownerType?.let {
+          ownerType.resolve(context, contextRawType, visitedTypeVariables)
+        }
+        var changed = newOwnerType !== ownerType
+        var args = original.actualTypeArguments
+        var t = 0
+        val length = args.size
+        while (t < length) {
```
JvmName, line 249: isn't this equivalent, or am I missing something
```kotlin
for(t in 0..args.size)
```

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
```diff
@@ -0,0 +1,643 @@
...
+  if (toResolve == rawType) {
+    return context
+  }
+
+  // we skip searching through interfaces if unknown is an interface
+  if (toResolve.isInterface) {
+    val interfaces = rawType.interfaces
+    var i = 0
+    val length = interfaces.size
+    while (i < length) {
```
JvmName, line 311:
```suggestion
	for(i in 0..interfaces.size) {
```

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
```diff
@@ -0,0 +1,643 @@
...
+    throw RuntimeException("Failed to find the generated JsonAdapter class for $type", e)
+  } catch (e: NoSuchMethodException) {
+    if (possiblyFoundAdapter != null && type !is ParameterizedType && possiblyFoundAdapter.typeParameters.isNotEmpty()) {
+      throw RuntimeException(
+        "Failed to find the generated JsonAdapter constructor for '" +
+          type +
+          "'. Suspiciously, the type was not parameterized but the target class '" +
+          possiblyFoundAdapter.canonicalName +
+          "' is generic. Consider using " +
+          "Types#newParameterizedType() to define these missing type variables.",
```
JvmName, line 416:
```suggestion
        "Failed to find the generated JsonAdapter constructor for '$type'. Suspiciously, the type was not parameterized but the target class '${possiblyFoundAdapter.canonicalName}' is generic. Consider using Types#newParameterizedType() to define these missing type variables."
```

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
ZacSweers: Close! `for(t in args.indices) {`

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
ZacSweers: Did this locally

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
ZacSweers: Same thing with indices

PR #1462, opened 2021-12-12
Convert Moshi.java to Kotlin

ZacSweers:
Ref #1252


moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+import javax.annotation.CheckReturnValue
+
+/**
+ * Coordinates binding between JSON values and Java objects.
+ *
+ * Moshi instances are thread-safe, meaning multiple threads can safely use a single instance
+ * concurrently.
+ */
+public class Moshi internal constructor(builder: Builder) {
+  private val factories: List<JsonAdapter.Factory> = buildList {
```
Egorand, line 35: Don't recall if `buildList` and `buildSet` are stable already or not, if not - should we avoid using them?

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+  @CheckReturnValue
+  public fun <T> adapter(type: Type, annotations: Set<Annotation>): JsonAdapter<T> {
+    return adapter(type, annotations, fieldName = null)
+  }
+
+  /**
+   * @param fieldName An optional field name associated with this type. The field name is used as a
+   * hint for better adapter lookup error messages for nested structures.
+   */
+  @CheckReturnValue // Factories are required to return only matching JsonAdapters.
```
Egorand, line 81: It seems that here and on the next method this comment was attached to `@SuppressWarnings` and not `@CheckReturnValue`. If `@SuppressWarnings` is not needed anymore, should we remove the comment?

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+    internal val factories: MutableList<JsonAdapter.Factory> = ArrayList()
+    internal var lastOffset = 0
+
+    public fun <T> add(type: Type, jsonAdapter: JsonAdapter<T>): Builder = apply {
+      add(
+        newAdapterFactory(
+          type,
+          jsonAdapter
+        )
+      )
```
Egorand, line 176: Or was there a reason for this formatting?
```suggestion
      add(newAdapterFactory(type, jsonAdapter))
```

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    @Suppress("unused")
+    public fun <T> addLast(type: Type, jsonAdapter: JsonAdapter<T>): Builder = apply {
+      addLast(
+        newAdapterFactory(
+          type,
+          jsonAdapter
+        )
+      )
```
Egorand, line 202:
```suggestion
      addLast(newAdapterFactory(type, jsonAdapter))
```

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Returns a JSON adapter that was already created for this call, or null if this is the first
+     * time in this call that the cache key has been requested in this call. This may return a
+     * lookup that isn't yet ready if this lookup is reentrant.
+     */
+    fun <T> push(type: Type, fieldName: String?, cacheKey: Any): JsonAdapter<T>? {
+      // Try to find a lookup with the same key for the same call.
+      var i = 0
+      val size = callLookups.size
+      while (i < size) {
```
Egorand, line 259: For seemed to be a better fit here..

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+     */
+    fun pop(success: Boolean) {
+      stack.removeLast()
+      if (!stack.isEmpty()) return
+      lookupChainThreadLocal.remove()
+      if (success) {
+        synchronized(adapterCache) {
+          var i = 0
+          val size = callLookups.size
+          while (i < size) {
```
Egorand, line 297: Same here, `for` was better

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    override fun toString(): String {
+      return adapter?.toString() ?: super.toString()
+    }
+  }
+
+  internal companion object {
+    @JvmField
+    val BUILT_IN_FACTORIES: List<JsonAdapter.Factory> = buildList(5) {
```
Egorand, line 354:
```suggestion
    val BUILT_IN_FACTORIES: List<JsonAdapter.Factory> = buildList(6) {
```

moshi/src/main/java/com/squareup/moshi/Moshi.kt
```diff
@@ -0,0 +1,388 @@
...
+      return checkNotNull(adapter) { "JsonAdapter isn't ready" }.body()
+    }
+
+    override fun toString(): String {
+      return adapter?.toString() ?: super.toString()
+    }
+  }
+
+  internal companion object {
+    @JvmField
```
Egorand, line 353: Can remove once all internal code is konverted?

Egorand (approved):

moshi/src/main/java/com/squareup/moshi/Moshi.kt
ZacSweers: they're stable now 👍 

https://kotlinlang.org/docs/whatsnew16.html#stable-collection-builders

moshi/src/main/java/com/squareup/moshi/Moshi.kt
ZacSweers: You can't write for loops in Kotlin with multiple variables

moshi/src/main/java/com/squareup/moshi/Moshi.kt
ZacSweers: same as above!

moshi/src/main/java/com/squareup/moshi/Moshi.kt
ZacSweers: Moved in 70091a9

moshi/src/main/java/com/squareup/moshi/Moshi.kt
ZacSweers: We could, though the only other place it's used is from MoshiTest and we won't be converting that to Kotlin right?

PR #1466, opened 2021-12-21
Convert classes in moshi package to kotlin

spencergriffin:
Ref #1252 

Japicmp has the following changes that I'm unsure about:
```Comparing binary compatibility of  against 
***! MODIFIED CLASS: PUBLIC (<- PACKAGE_PROTECTED) FINAL com.squareup.moshi.JsonValueWriter  (not serializable)
	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
***! MODIFIED CLASS: PUBLIC ABSTRACT com.squareup.moshi.JsonWriter  (not serializable)
	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
	---! REMOVED METHOD: PUBLIC(-) ABSTRACT(-) com.squareup.moshi.JsonWriter value(boolean)
		---  REMOVED EXCEPTION: java.io.IOException
+++  NEW CLASS: PUBLIC(+) FINAL(+) kotlin.collections.builders.SerializedCollection  (changed from Serializable to Externalizable)
	+++  CLASS FILE FORMAT VERSION: 50.0 <- n.a.```

The java version of JsonWriter has methods overloading both boolean and Boolean, and I'm not sure that works in Kotlin.


ZacSweers:
Please split these into smaller PRs and take note of existing ones! This is basically unreviewable, sorry!

PR #1475, opened 2021-12-27
Convert `JsonAdapter` to Kotlin

ZacSweers:
Note that there's more to be done here I think, namely exploring removing the NonNull adapter and making the nullSafe() adapter public so that nullability is directly in the API. Saving that for another day though


moshi/japicmp/build.gradle.kts
```diff
@@ -30,6 +30,9 @@ val japicmp = tasks.register<JapicmpTask>("japicmp") {
     "com.squareup.moshi.internal.NullSafeJsonAdapter", // Internal.
     "com.squareup.moshi.internal.Util" // Internal.
   )
+  methodExcludes = listOf(
+    "com.squareup.moshi.JsonAdapter#indent(java.lang.String)" // Was unintentionally open before
```
ZacSweers, line 34: @swankjesse @JakeWharton let me know if this is right,  can make it `open` if it was intentional before

moshi/src/main/java/com/squareup/moshi/JsonAdapter.kt
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * Returns a JSON adapter equal to this JSON adapter, but that refuses null values. If null is
+   * read or written this will throw a [JsonDataException].
+   *
+   * Note that this adapter will not usually be invoked for absent values and so those must be
+   * handled elsewhere. This should only be used to fail on explicit nulls.
+   */
+  @CheckReturnValue
+  public fun nonNull(): JsonAdapter<T> {
+    return if (this is NonNullJsonAdapter<*>) {
```
JvmName, line 180: nit -- same thought here

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonAdapter.kt
```diff
@@ -0,0 +1,298 @@
...
+   * Returns a JSON adapter equal to this JSON adapter, but with support for reading and writing
+   * nulls.
+   */
+  @CheckReturnValue
+  public fun nullSafe(): JsonAdapter<T> {
+    return if (this is NullSafeJsonAdapter<*>) {
+      this
+    } else {
+      NullSafeJsonAdapter(this)
+    }
```
JvmName, line 168: nit -- curious about `when` vs `if` here --

```suggestion
    return when(this)
      is NullSafeJsonAdapter<*> -> this
      else -> NullSafeJsonAdapter(this)
```

moshi/src/main/java/com/squareup/moshi/JsonAdapter.kt
```diff
@@ -0,0 +1,298 @@
...
+ */
+package com.squareup.moshi
+
+import com.squareup.moshi.internal.NonNullJsonAdapter
+import com.squareup.moshi.internal.NullSafeJsonAdapter
+import okio.Buffer
+import okio.BufferedSink
+import okio.BufferedSource
+import okio.IOException
+import org.intellij.lang.annotations.Language
```
Egorand, line 24: Oh interesting, which artifact does this come from, kotlin-stdlib?

Egorand (approved):

moshi/src/main/java/com/squareup/moshi/JsonAdapter.kt
ZacSweers: yup, JetBrains annotations includes it

PR #1486, opened 2022-01-07
Convert JsonUtf8Writer to Kotlin

ZacSweers:


moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.kt
```diff
@@ -0,0 +1,387 @@
...
+    pushScope(empty)
+    pathIndices[stackSize - 1] = 0
+    sink.writeByte(openBracket.code)
+    return this
+  }
+
+  /** Closes the current scope by appending any necessary whitespace and the given bracket. */
+  private fun close(empty: Int, nonempty: Int, closeBracket: Char): JsonWriter {
+    val context = peekScope()
+    check(!(context != nonempty && context != empty)) { "Nesting problem." }
```
JvmName, line 95: i think the `!` can be propagated here?

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.kt
ZacSweers: 4e68851

PR #1489, opened 2022-01-08
Convert JsonValueSource to Kotlin

ZacSweers:


moshi/src/main/java/com/squareup/moshi/JsonValueSource.kt
```diff
@@ -0,0 +1,208 @@
...
+            limit = index + 1
+          }
+          '\'' -> {
+            state = STATE_SINGLE_QUOTED
+            limit = index + 1
+          }
+          '/' -> {
+            source.require(index + 2)
+            when (buffer[index + 1]) {
+              '/'.code.toByte() -> {
```
JvmName, line 110: does it make a difference, either from perf or readability, to coerce `buffer[index+1]` to `Char` instead of coercing `'/'` to a `Byte`?

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonValueSource.kt
ZacSweers: I don't think it helps with readability too much and we'd like to keep things at a byte

PR #1491, opened 2022-01-08
Convert JsonValueWriter to Kotlin

ZacSweers:


moshi/src/main/java/com/squareup/moshi/JsonValueWriter.kt
```diff
@@ -0,0 +1,265 @@
...
+    return this
+  }
+
+  override fun value(value: Double): JsonWriter {
+    require(
+      !(
+        !lenient &&
+          (java.lang.Double.isNaN(value) || value == Double.NEGATIVE_INFINITY || value == Double.POSITIVE_INFINITY)
+        )
+    ) { "Numeric values must be finite, but was $value" }
```
JvmName, line 152: can the `!` be propagated so this can be simplified?

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.kt
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    // Everything else gets converted to a BigDecimal.
+    val bigDecimalValue = if (value is BigDecimal) value else BigDecimal(value.toString())
+    if (promoteValueToName) {
+      promoteValueToName = false
+      return name(bigDecimalValue.toString())
+    }
+    add(bigDecimalValue)
+    pathIndices[stackSize - 1]++
```
JvmName, line 194:
```suggestion
    return when(value){
		is Byte, is Short, is Int, is Long -> value(value.toLong())     // If it's trivially converted to a long, do that.
		is Float, is Double -> value(value.toDouble()) // If it's trivially converted to a double, do that.
		null -> nullValue()
		else -> { // Everything else gets converted to a BigDecimal.
		val bigDecimalValue = if (value is BigDecimal) value else BigDecimal(value.toString())
		if (promoteValueToName) {
			promoteValueToName = false
			return name(bigDecimalValue.toString())
    	}
    	add(bigDecimalValue)
    	pathIndices[stackSize - 1]++
		}
	}



    
        
    
```

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.kt
ZacSweers: oof yeah let me clean this up

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.kt
```diff
@@ -0,0 +1,265 @@
...
+  var stack = arrayOfNulls<Any>(32)
+  private var deferredName: String? = null
+
+  init {
+    pushScope(EMPTY_DOCUMENT)
+  }
+
+  fun root(): Any? {
+    val size = stackSize
+    check(size <= 1 && (size != 1 || scopes[0] == NONEMPTY_DOCUMENT)) { "Incomplete document" }
```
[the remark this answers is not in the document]
ZacSweers: yeah I tried to clean up the readability a bit

moshi/src/main/java/com/squareup/moshi/JsonValueWriter.kt
ZacSweers: 21b6dc4

PR #1493, opened 2022-01-08
Convert RecordJsonAdapter to Kotlin

ZacSweers:
Requires a bit of extra tweaking for the kotlin compilation task


moshi/build.gradle.kts
```diff
@@ -11,20 +12,27 @@ plugins {
 }
 
 val mainSourceSet by sourceSets.named("main")
-val java16 by sourceSets.creating {
+val java16: SourceSet by sourceSets.creating {
```
ZacSweers, line 15: it was nagging me about a platform type so made it explicit

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.kt
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * **NOTE:** Java records require JDK 16 or higher.
+ */
+internal class RecordJsonAdapter<T>(
+  private val constructor: MethodHandle,
+  private val targetClass: String,
+  componentBindings: Map<String, ComponentBinding<Any>>
+) : JsonAdapter<T>() {
+
+  @JvmRecord
```
ZacSweers, line 43: It was a record before, but doesn't truuuuuly matter here since we have access to a data class. Did appreciate being able to use a JvmRecord though, but can remove since it's ultimately irrelevant

ZacSweers:
```suggestion
```

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.kt
```diff
@@ -0,0 +1,174 @@
...
+        if (replaced != null) {
+          throw IllegalArgumentException(
+            "Conflicting components:\n    ${replaced.componentName}\n    ${componentBinding.componentName}"
+          )
+        }
+        component.type
+      }
+
+      val constructor = try {
+        lookup.findConstructor(rawType, methodType(knownNotNull(Void::class.javaPrimitiveType), componentRawTypes))
```
ZacSweers, line 140: Should we save off the `knownNotNull(Void::class.javaPrimitiveType)` as a top-level property?

Egorand: Is it just a better-documented `!!`? If it's used extensively yeah can be a top-level property.

Egorand (approved):

moshi/src/main/java16/com/squareup/moshi/RecordJsonAdapter.kt
ZacSweers: It skips the nullcheck

PR #1496, opened 2022-01-10
Inline Moshi extension functions to Moshi itself

ZacSweers:
No need to be separate anymore and saves us an import. Still binary-compatible with before!

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
+++ b/moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+
 package com.squareup.moshi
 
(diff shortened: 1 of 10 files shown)


moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
```diff
@@ -25,16 +27,19 @@ import kotlin.reflect.typeOf
  * @return a [JsonAdapter] for [T], creating it if necessary. Note that while nullability of [T]
  *         itself is handled, nested types (such as in generics) are not resolved.
  */
+@Deprecated("Use the Moshi instance version instead", level = DeprecationLevel.HIDDEN)
```
JvmName, line 30: is there a refactoring helper that can be added here, or is that infeasible for extension-to-instance?

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/-MoshiKotlinExtensions.kt
ZacSweers: It's hidden and code simply needs to just be recompiled for this to work now, no refactoring needed!

PR #1477, opened 2021-12-30
Convert JsonWriter to Kotlin

Goooler:
Ref #1252.

PR #1501, opened 2022-01-12
Finish up JsonWriter Kotlin conversion

ZacSweers:
Finishing up #1477 to proceed with other work


moshi/src/main/java/com/squareup/moshi/JsonWriter.kt
```diff
@@ -0,0 +1,543 @@
...
+    when (value) {
+      is Map<*, *> -> {
+        beginObject()
+        for ((key, value1) in value) {
+          require(key is String) {
+            if (key == null) {
+              "Map keys must be non-null"
+            } else {
+              "Map keys must be of type String: ${key.javaClass.name}"
+            }
```
JvmName, line 401: nit -- i feel like this'd be more idiomatic as 2 `require`s? 
```suggestion
		 requireNotNull(key) { "Map keys must be not-null" } 
		 require(key is String) { "Map keys must be of type String: ${key.javaClass.name}" }
```

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonWriter.kt
```diff
@@ -0,0 +1,543 @@
...
+   *
+   * @return this writer.
+   * @see JsonReader.readJsonValue
+   */
+  @Throws(IOException::class)
+  public fun jsonValue(value: Any?): JsonWriter {
+    when (value) {
+      is Map<*, *> -> {
+        beginObject()
+        for ((key, value1) in value) {
```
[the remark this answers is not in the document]
ZacSweers: e0477a4

moshi/src/main/java/com/squareup/moshi/JsonWriter.kt
ZacSweers: dig, though github won't let me add suggestion from here. Let me push

ZacSweers: 16de888

moshi/src/main/java/com/squareup/moshi/JsonWriter.kt
```diff
@@ -0,0 +1,538 @@
...
+  protected var pathNames: Array<String?> = arrayOfNulls(32)
+
+  @JvmField
+  protected var pathIndices: IntArray = IntArray(32)
+
+  /**
+   * A string containing a full set of spaces for a single level of indentation, or null for no
+   * pretty printing.
+   */
+  @JvmField
```
Goooler, line 153: Seems these `@JvmField` could be removed now?

PR #1505, opened 2022-01-16
Convert JsonReader to Kotlin

ZacSweers:
Made it sealed along the way like JsonWriter

Some of the properties had different docs for setting and getting, which dokka doesn't seem to have a nice way to differentiate. Went with a little marked up form but open to suggestions.


moshi/src/main/java/com/squareup/moshi/JsonReader.kt
```diff
@@ -0,0 +1,672 @@
...
+
+  private var tags: MutableMap<Class<*>, Any>? = null
+
+  protected constructor() {
+    scopes = IntArray(32)
+    pathNames = arrayOfNulls(32)
+    pathIndices = IntArray(32)
+  }
+
+  // Package-private to control subclasses.
```
JvmName, line 245: errant comment
```suggestion
```

moshi/src/main/java/com/squareup/moshi/JsonReader.kt
```diff
@@ -0,0 +1,672 @@
...
+    internal val strings: Array<out String>,
+    internal val doubleQuoteSuffix: OkioOptions
+  ) {
+    /** Returns a copy of this [Option's][Options] strings. */
+    public fun strings(): List<String> {
+      return buildList(strings.size) {
+        for (string in strings) {
+          add(string)
+        }
+      }
```
JvmName, line 590: couldn't you just say `strings.toList()` or ` strings.map { it }`?

JvmName (approved):

moshi/src/main/java/com/squareup/moshi/JsonReader.kt
ZacSweers: Collection builders return (truly) immutable instances

PR #1508, opened 2022-01-17
Convert LinkedHashTreeMap to Kotlin

ZacSweers:
Tried to keep this as close to the original impl as possible with some minor tweaks to better handle places where we skimmed known nullability that Kotlin isn't happy about.


moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
```diff
@@ -0,0 +1,775 @@
...
+ * removal.
+ *
+ * This implementation was derived from Android 4.1's TreeMap and LinkedHashMap classes.
+ */
+internal class LinkedHashTreeMap<K, V>
+/**
+ * Create a tree map ordered by [comparator]. This map's keys may only be null if [comparator] permits.
+ *
+ * @param comparator the comparator to order elements with, or null to use the natural
+ *     ordering.
```
Goooler, line 23:
```suggestion
 * @param comparator the comparator to order elements with, or null to use the natural ordering.
```

ZacSweers: Hi Gooler, this is a needless formatting suggestion as it is correct as is. Since this is a complex PR I'd like to ask that you avoid trying to contribute further changes like this as they just add noise that distract from the more important parts of the review.

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
```diff
@@ -0,0 +1,775 @@
...
+  private var modCount = 0
+  private var threshold = table.size / 2 + table.size / 4 // 3/4 capacity
+  private var entrySet: EntrySet? = null
+  private var keySet: KeySet? = null
+
+  override val keys: MutableSet<K>
+    get() {
+      val result: KeySet? = keySet
+      return result ?: KeySet().also { keySet = it }
+    }
```
Goooler, line 42:
```suggestion
    get() = keySet ?: KeySet().also { keySet = it }
```

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
```diff
@@ -0,0 +1,775 @@
...
+    created.mutableValue = value
+    return result
+  }
+
+  override val entries: MutableSet<MutableMap.MutableEntry<K, V>>
+    get() = entrySet ?: EntrySet().also { entrySet = it }
+
+  override fun get(key: K) = findByObject(key)?.value
+
+  override fun containsKey(key: K) = findByObject(key) != null
```
Goooler, line 56:
```suggestion
  override fun get(key: K): V? = findByObject(key)?.value

  override fun containsKey(key: K): Boolean = findByObject(key) != null
```

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
ZacSweers: Please stop, per my other comment

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
ZacSweers: Please stop, per my other comment

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
```diff
@@ -0,0 +1,775 @@
...
+      val next = e!!.next
+      e.prev = null
+      e.next = null
+      e = next
+    }
+    header.prev = header
+    header.next = header.prev
+  }
+
+  override fun remove(key: K) = removeInternalByKey(key)?.value
```
Goooler, line 76:
```suggestion
  override fun remove(key: K): V? = removeInternalByKey(key)?.value
```

ZacSweers: Please take no for an answer

moshi/src/main/java/com/squareup/moshi/LinkedHashTreeMap.kt
```diff
@@ -0,0 +1,775 @@
...
+   */
+  private fun rebalance(unbalanced: Node<K, V>?, insert: Boolean) {
+    var node = unbalanced
+    while (node != null) {
+      val left = node.left
+      val right = node.right
+      val leftHeight = left?.height ?: 0
+      val rightHeight = right?.height ?: 0
+      val delta = leftHeight - rightHeight
+      if (delta == -2) {
```
JvmName, line 393: real minor nit -- seems like this could be a nice `when` instead of if/else?

JvmName (approved):

Issue #1522, opened 2022-02-21
ksp 1.0.3 breaks build


giangpham96:
After I update my ksp from `1.6.10-1.0.2` to `1.6.10-1.0.3`, the build breaks. I am not sure if this is a ksp bug or moshi bug. The log doesn't say much
```
e: [ksp] java.lang.IllegalStateException: Required value was null.

	at shadow.com.squareup.kotlinpoet.ksp.UtilsKt.toClassNameInternal(utils.kt:63)
	at shadow.com.squareup.kotlinpoet.ksp.KsClassDeclarationsKt.toClassName(ksClassDeclarations.kt:24)
	at shadow.com.squareup.kotlinpoet.ksp.KsTypesKt.toTypeName(ksTypes.kt:66)
	at shadow.com.squareup.kotlinpoet.ksp.KsTypesKt.toTypeName(ksTypes.kt:57)
> Task :net_entities:kspDevelopmentDebugKotlin
	at com.squareup.moshi.kotlin.codegen.ksp.TargetTypesKt.toPropertySpec(TargetTypes.kt:261)
	at com.squareup.moshi.kotlin.codegen.ksp.TargetTypesKt.declaredProperties(TargetTypes.kt:236)
	at com.squareup.moshi.kotlin.codegen.ksp.TargetTypesKt.targetType(TargetTypes.kt:124)
	at com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessor.adapterGenerator(JsonClassSymbolProcessorProvider.kt:118)
	at com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessor.process(JsonClassSymbolProcessorProvider.kt:91)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$4$1.invoke(KotlinSymbolProcessingExtension.kt:223)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$4$1.invoke(KotlinSymbolProcessingExtension.kt:221)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension.handleException(KotlinSymbolProcessingExtension.kt:312)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension.doAnalysis(KotlinSymbolProcessingExtension.kt:221)
[...]

PR #1527, opened 2022-03-30
KSP error messaging: Include name of the type being processed when creating the adapterGenerator fails

trevjonez:
Follow up to the error/stack I posted in #1522 
After poking around on my project I realized I had some bad generics syntax in my test source set causing the intrinsic null check failure in the ksp task.

Messaging now at least provides a hint as to where to start investigating


ZacSweers (commented):
Apologies for the delay. If you're still interested in contributing this, please update the branch and ping here when it's ready for review again

trevjonez:
@ZacSweers should be good to go now.

ZacSweers:
Looks like you need to fix formatting

Issue #1530, opened 2022-04-14
After updating ksp to 1.6.20-1.0.5, Moshi generates error code for `typealias` of class with type parameters


efemoney:
This may as well be a bug in ksp but after updating kotlin & ksp to 1.6.20-1.0.5, given:

```kotlin
typealias EventMetadata = Map<String, String>

@JsonClass(generateAdapter = true) 
data class ClassUnderTest(
  val metadata: EventMetadata,
)
```

in the generated code, the `Map` loses its type arguments and is therefore a compile error.

yigit:
This might be related to: https://github.com/google/ksp/issues/881

ting-yuan:
The above change was introduced in KSP 1.0.5, where type arguments are now returned for the reference, instead of the complete, underlying type. In this case, it is a reference to an alias and therefore the arguments defined in declaration of the alias are not included.

KotlinPoet [relies on the original behavior](https://github.com/square/kotlinpoet/blob/master/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/ksTypes.kt#L78). [This test](https://github.com/square/kotlinpoet/blob/master/interop/ksp/test-processor/src/test/kotlin/com/squareup/kotlinpoet/ksp/test/processor/TestProcessorTest.kt#L260) caught the issue when I tried to bump KSP to 1.0.5.

Egorand:
Hey @ting-yuan, I'm trying to upgrade KotlinPoet to 1.6.20-1.0.5 and bumped into this issue. Any suggestions on how to migrate the logic? Here's what I've got at this point:

For the following declaration:

```kotlin
genericMapAlias: GenericMapTypeAlias<String, Int>,
```

where

```kotlin
typealias GenericMapTypeAlias<V, K> = Map<K, V>
```

I can access:

- `KSTypeReference.element.typeArguments` = `[String, Int]`
- `KSTypeAlias.typeParameters` = `[V, K]`
- `KSTypeAlias.type.declaration.typeParameters` = `[K, V]`

I can match parameter names manually and put the type arguments in the right order, but wondering if there's a simpler way to do this.

ting-yuan:
I think what you mentioned (match parameter names and put type arguments in the right order, which is how KotlinPoet worked with KSP 1.0.4, right?) is probably the simplest/best way. 

Thanks for fixing this. My apology for not noticing and announcing the behavior change in KSP.

Egorand:
[...]

PR #1531, opened 2022-04-14
Reproducer for #1530

efemoney:
Added a regression test for #1530


efemoney:
Took a second look at this and this test doesnt really repro. i will find some time to update it.

PR #1542, opened 2022-06-02
Make class instances with type parameter info equal to equivalent parameterized types.

moomoomoo309:
This is a _very_ specific bug I've run into, primarily with moshi-adapters. It happens when you get the return type from a method or a type parameter from a method and pass that in as a class instance to an adapter. It can also occur from the type parameters in JsonAdapter<T> if T has bounds.

In those instances, when it calls Types.equals, because one side is a ParameterizedType, and the other is a Class<?>, it immediately returns false without looking at the types themselves. This changes that, so it first checks if the raw types match, then checks if the type parameters match in those cases.


ZacSweers (approved):
This looks reasonable to me, thanks for the tests! @swankjesse could I ask your eyes on this too?

moshi/src/test/java/com/squareup/moshi/TypesTest.java
```diff
@@ -292,6 +300,31 @@ public void arrayEqualsGenericTypeArray() {
...
+    );
+    Method returningE = null;
+    for (Method method: TypesTest.class.getDeclaredMethods()) {
+      if (method.getName().contains("methodReturningE")) {
+        returningE = method;
+        break;
+      }
+    }
+    Type rawType = Types.getRawType(returningE.getGenericReturnType());
+    Truth.assertThat(Types.equals(type, rawType)).isTrue();
```
ZacSweers, line 323: nit: static import `assertThat`

moshi/src/test/java/com/squareup/moshi/TypesTest.java
```diff
@@ -292,6 +300,31 @@ public void arrayEqualsGenericTypeArray() {
     assertThat(Types.equals(Types.arrayOf(String.class), String[].class)).isTrue();
   }
 
+  private E<A, B> methodReturningE() {
+    throw new Error(); // Intentionally not implemented
```
ZacSweers, line 304: nit: use `NotImplementedException`?

moshi/src/main/java/com/squareup/moshi/Types.kt
```diff
@@ -212,11 +212,17 @@ public object Types {
       is Class<*> -> {
         return if (b is GenericArrayType) {
           equals(a.componentType, b.genericComponentType)
+        } else if (b is ParameterizedType && a.rawType == b.rawType) {
+          // Class instance with generic info, from method return types
+          return a.typeParameters.map { it.bounds }.toTypedArray().flatten() == b.actualTypeArguments.toList()
```
ZacSweers, line 217: Can we use this?

```suggestion
          return a.typeParameters.flatMap { it.bounds } == b.actualTypeArguments.toList()
```

moshi/src/test/java/com/squareup/moshi/TypesTest.java
moomoomoo309: Makes sense to me, I used NotImplementedError from Kotlin, since that shouldn't be checked since it extends Error instead of Exception.

moshi/src/main/java/com/squareup/moshi/Types.kt
moomoomoo309: Yes, but I had to change it.bounds to it.bounds.toList(), because flatMap only takes Iterable, not Array.

ZacSweers:
Apologies for the delay, thanks!

PR #1545, opened 2022-06-19
DefaultConstructorMarker is now public

vvb2060:
JetBrains/kotlin@83383ab9

### DIFF
--- a/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
+++ b/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
@@ -23,7 +23,7 @@ import com.squareup.kotlinpoet.ClassName
  * - Keeping the generated adapter class name + public constructor for reflective lookup.
  * - Keeping any used JsonQualifier annotations and the properties they are attached to.
  * - If the target class has default parameter values, also keeping the associated synthetic
- *   constructor as well as the DefaultConstructorMarker type Kotlin adds to it.
+ *   constructor.
  *
  * Each rule is intended to be as specific and targeted as possible to reduce footprint, and each is
  * conditioned on usage of the original target type.
@@ -70,13 +70,10 @@ public data class ProguardConfig(
     if (targetConstructorHasDefaults) {
       // If the target class has default parameter values, keep its synthetic constructor
       //
-      // -keepnames class kotlin.jvm.internal.DefaultConstructorMarker
       // -keepclassmembers @com.squareup.moshi.JsonClass @kotlin.Metadata class * {
       //     synthetic <init>(...);
       // }
       //
-      appendLine("-if class $targetName")
-      appendLine("-keepnames class kotlin.jvm.internal.DefaultConstructorMarker")
       appendLine("-if class $targetName")
       appendLine("-keepclassmembers class $targetName {")
       val allParams = targetConstructorParams.toMutableList()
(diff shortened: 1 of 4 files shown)


moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
```diff
@@ -70,14 +70,11 @@ public data class ProguardConfig(
     if (targetConstructorHasDefaults) {
       // If the target class has default parameter values, keep its synthetic constructor
       //
-      // -keepnames class kotlin.jvm.internal.DefaultConstructorMarker
       // -keepclassmembers @com.squareup.moshi.JsonClass @kotlin.Metadata class * {
       //     synthetic <init>(...);
       // }
       //
       appendLine("-if class $targetName")
-      appendLine("-keepnames class kotlin.jvm.internal.DefaultConstructorMarker")
```
ZacSweers, line 79: We should keep this, we still want to allow R8 to minify when it's truly never used

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
```diff
@@ -61,11 +62,7 @@ private val METADATA: Class<out Annotation>? = try {
 // We look up the constructor marker separately because Metadata might be (justifiably)
 // stripped by R8/Proguard but the DefaultConstructorMarker is still present.
 @JvmField
-public val DEFAULT_CONSTRUCTOR_MARKER: Class<*>? = try {
-  Class.forName("kotlin.jvm.internal.DefaultConstructorMarker")
-} catch (ignored: ClassNotFoundException) {
-  null
-}
+public val DEFAULT_CONSTRUCTOR_MARKER: Class<*> = DefaultConstructorMarker::class.java
```
ZacSweers, line 65: Let's make this work like the `METADATA` field. We do this a particular way to prevent R8 from automatically keeping it

moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
vvb2060: `DefaultConstructorMarker` class is just a flag, keeping the name means keeping everything.  I want to obfuscate it with R8.

moshi/src/main/java/com/squareup/moshi/internal/Util.kt
vvb2060: Using reflection means we have to keep all(This class is just a marker, it only has a name and no content.), using references allows R8 to obfuscate safely.

moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
vvb2060: I have removed all relevant keep rules, which means it is auto deleted when not in use.

vvb2060:
As long as we have a kotlin class, the class is "in use", so it can't be deleted anyway.  
I think it's better to let it be able to be obfuscated.

ZacSweers:
@vvb2060 please do what the comments asked if you want to get these merged

vvb2060:
I can't understand, after following comments, this class will not be obfuscated.

ZacSweers:
I don't think you do. Closing this for now. I don't think we should be directly referencing internal code anyway as it's not part of the public API and subject to change. At least with reflection we can gracefully fall back.

vvb2060:
I'm sorry, it looks like you didn't read my comment at all. Also, this class is closely related to the compiler, it will not change due to binary compatibility.

vvb2060:
Well, how about this? Catch error that this class cannot be found at runtime
```kotlin
public val DEFAULT_CONSTRUCTOR_MARKER: Class<*>? = try {
  DefaultConstructorMarker::class.java
} catch (ignored: NoClassDefFoundError) {
  null
}
```

PR #1547, opened 2022-06-22
Suppress warning in generated code: `Parameter 'moshi' is never used`

yogurtearl:
We are getting this warning on generated type adapters with Kotlin `1.6.21` and moshi `1.14.0-SNAPSHOT` published from [this commit](https://github.com/square/moshi/commit/53bb46c20839fe155cdfcc042553d4acf6237344) :

`Parameter 'moshi' is never used`

`"unused"` doesn't seem to suppress that, but `"UNUSED_PARAMETER"` does work.

I couldn't figure out how to get `kotlin-compile-testing` to output the warning, so I could write a test to assert the warning is gone.


StephanSchuster:
Same here. Would be happy if this gets merged.

ZacSweers (approved):
Sorry for the delay, missed this!

PR #1550, opened 2022-06-24
Moshi 1.14.x updates

ZacSweers:
- Update Kotlin to 1.7.0
- Update KotlinPoet to 1.13.0
- No longer shade kotlinpoet-ksp APIs 
- Update to KSP 1.7.0-1.0.6


yogurtearl:
This is branched from `1.13.x`, so `1.14.x` won't have all the changes that are in the `1.14.0-SNAPSHOT` builds being published from `master`, which is a bit awkward. 

Might want to mention that in the 1.14 release notes. 
Also, might want to switch the `SNAPSHOT`'s from `master` to be `2.0.0-SNAPSHOT`

yogurtearl:
as an example, this reified `adapter` method is in the current `1.14.0-SNAPSHOT` but not in this branch. 

https://github.com/square/moshi/blob/53bb46c20839fe155cdfcc042553d4acf6237344/moshi/src/main/java/com/squareup/moshi/Moshi.kt#L82

yogurtearl:
I built/tested the branch locally and this branch works for us with Kotlin 1.6.21, Compose 1.2.0-rc02, which need the updated ksp/kotlinpoet versions.

ZacSweers:
This PR wouldn't merge to main, it would just be branching from 1.13 and be used to make a 1.14 release with all the above updates. 2.0 is what's happening on main and still needs a bit of work

yogurtearl:
> This PR wouldn't merge to main, it would just be branching from 1.13 and be used to make a 1.14 release with all the above updates. 2.0 is what's happening on main and still needs a bit of work

Yeah, that is what makes it a bit expected that main is publishing `1.14.0-SNAPSHOT` and not `2.0.0-SNAPSHOT` 

fixed here: https://github.com/square/moshi/pull/1552

swankjesse (approved):

mmasny-nemo:
Hello, have you planned the 1.14 release for any specific date yet?

ZacSweers:
[nope](https://www.zacsweers.dev/you-are-not-blocked/)

yogurtearl:
FWIW, we have an internal fork with these changes in this PR applied and it is working well for us. :)

t1u1:
We tried out the new API and it works great!

swankjesse (approved):

PR #1552, opened 2022-06-28
publish `2.0.0-SNAPSHOT` from `main`

yogurtearl:
Right now the builds on `main` are publishing `1.14.0-SNAPSHOT` but it contains stuff that is not intended to go into `1.14` and is only intended for `2.x`

With this change, `main` will produce `2.0.0-SNAPSHOT` instead of `1.14.0-SNAPSHOT`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,7 +28,7 @@ plugins {
 
 allprojects {
   group = "com.squareup.moshi"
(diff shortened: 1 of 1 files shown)


yogurtearl:
PR to publish `1.14.0-SNAPSHOT` from the `moshi_1.14.x` branch for easier validation: https://github.com/square/moshi/pull/1553

PR #1553, opened 2022-06-29
Publish 1.14.0-SNAPSHOT from the 1.14 branch.

yogurtearl:
After `main` is updated to use  `2.0.0-SNAPSHOT` -> https://github.com/square/moshi/pull/1552

then `1.14.0-SNAPSHOT` can be published from the `moshi_1.14.x` branch


ZacSweers:
We're not looking to publish snapshots from this branch

PR #1573, opened 2022-08-29
Remove truth library

sahinakyol:
Motivation:
Is there any specific reason to use the truth testing library? I checked the whole code base and generally, you had used isEqual of type methods. I changed with JUnit assertion methods and I've not changed any test semantics. If there is a good reason then you can decline PR.

Changes:
Remove -> com.google.truth:truth:1.1.3 dependency
Remove -> import static com.google.common.truth.Truth.assertThat in classes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -39,4 +39,3 @@ assertj = "org.assertj:assertj-core:3.11.1"
 junit = "junit:junit:4.13.2"
 kotlinCompileTesting = { module = "com.github.tschuchortdev:kotlin-compile-testing", version.ref = "kotlinCompileTesting" }
 kotlinCompileTesting-ksp = { module = "com.github.tschuchortdev:kotlin-compile-testing-ksp", version.ref ="kotlinCompileTesting" }
-truth = "com.google.truth:truth:1.1.3"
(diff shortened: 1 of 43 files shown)


ZacSweers:
You should definitely raise an issue in the future before PRs like this, we could have saved you a lot of effort. Yes, we intentionally and want to continue using Truth :). Better error messages, better maintenance story, and a consistent API.

qwwdfsad:
(was just passing by, no opinion on this PR)

Could you please show a few examples of better error messages/maintenance story for Truth?

No doubts that `assertThat(message).isInstanceOf(EmptyMessage.class);` is a much better fit than `assertTrue(message instanceof EmptyMessage);`, so the question is rather about equivalent API  (e.g. `assertThat(a).isEqualTo(b)` vs `assertEquals(b, a)`)

JakeWharton:
Here's two I particularly like:

Matching contents but different line breaks.
```
value of: throwable.getMessage()
diff    : (line contents match, but line-break characters differ)
```
Matching contents but trailing newline.
```
expected                               : foo
but contained extra trailing whitespace: \n
```
Both of these are the result of calls to `isEqualTo` and dramatically help in the case where the visual output would appear to be equal despite the character-based content not actually being equal.

Issue #1170, opened 2020-07-06
[1.9.2] NoSuchMethodException when deserializing class with inline-class values and default values


pswaminathan:
Suppose I have the following:

```kotlin
inline class Distance(val millimeters: Long): Comparable<Distance> {
    (truncated)
}

@JsonClass(generateAdapter = true)
data class ScratchEntity(val id: Int, val distance: Distance)
```

This compiles and works fine. But if I give *either* property a default value, e.g.

```kotlin
inline class Distance(val millimeters: Long): Comparable<Distance> {
    (truncated)
}

@JsonClass(generateAdapter = true)
data class ScratchEntity(val id: Int = 1, val distance: Distance)
```

I get the following traceback:

```
Exception in thread "main" java.lang.NoSuchMethodException: mypackage.ScratchEntity.<init>(int, mypackage.Distance, int, kotlin.jvm.internal.DefaultConstructorMarker)
	at java.lang.Class.getConstructor0(Class.java:3082)
	at java.lang.Class.getDeclaredConstructor(Class.java:2178)
	at mypackage.ScratchEntityJsonAdapter.fromJson(ScratchEntityJsonAdapter.kt:59)
	at mypackage.ScratchEntityJsonAdapter.fromJson(ScratchEntityJsonAdapter.kt:19)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:43)
	at org.jetbrains.kotlin.idea.scratch.generated.ScratchFileRunnerGenerated$ScratchFileRunnerGenerated.<init>(tmp.kt:24)
	at org.jetbrains.kotlin.idea.scratch.generated.ScratchFileRunnerGenerated.main(tmp.kt:30)
```

I suspect this is an interaction with inline classes: if I change to `val distance: Int`, all is good.

LMK if there is anything I can do to help diagnose or clarify!

ZacSweers:
inline classes are not supported as they're not stable in kotlin itself yet

pswaminathan:
Sure, and I'm not suggesting this requires an immediate fix—I have a work-around I can use for now for my own use cases. But as inline classes are an extremely useful feature, and ostensibly the goal is to support them as they become stable in Kotlin, it is worth tracking and understanding how, as inline classes evolve and become stable, this would/could be addressed. No?

swankjesse:
Yep, when Kotlin has inline classes we’ll find a way to make ’em work nicely with Moshi.

moomoomoo309:
[...]

PR #1594, opened 2022-12-13
Adding support for inline classes and unsigned integers

amichne:
The intention of this pull request is to address the issues outlined in #1170.

This would also cover the unsigned types `[ULong, UInt, UShort, UByte]`.

I've outlined the thought process and approach for this solution in [this repository](https://github.com/amichne/moshi-value-classes), including the expected behavior, the current behavior (as of version `1.14.0`) and the behavior after this addition of the included adapters.

Any & all feedback is highly appreciated!

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -7,6 +7,9 @@ kotlinCompileTesting = "1.4.9-alpha01"
 kotlinpoet = "1.12.0"
 ksp = "1.7.0-1.0.6"
 ktlint = "0.45.2"
+assertk = "0.25"
+junitJupiter = "5.9.1"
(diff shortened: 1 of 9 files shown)


ZacSweers (changes requested):
Sorry but this is a bit too limited. This isn't something that can be accomplished with with Java reflection alone, nor can we only support it in reflection. If you want to take a shot at this using kotlin-reflect, that would be welcome. Code gen support does need to happen, but that can also be done separately.

A bit of general feedback:
- When contributing, please don't introduce unnecessary extra test libraries or frameworks and instead use what the rest of the codebase uses
  - Or if they're truly needed, they should be explained in the description
- Please keep code tidy. Leftover commented out lines or non-conventional extension utilities can clutter a review up

Issue #181, opened 2016-07-25
Java 8 Date Api Type Adapters


Antrakos:
I was looking for LocalDate and LocalDateTime type adapters, but I didn't found them. Can you add it in future?

swankjesse:
You can probably build adapters using `@ToJson` and `@FromJson` pretty easily as long as you also use our [`java.util.Date` adapter](https://github.com/square/moshi/blob/master/adapters/src/main/java/com/squareup/moshi/Rfc3339DateJsonAdapter.java).

Antrakos:
I have already done it pretty easily. However I think this should be included in library.

kenyee:
`import com.squareup.moshi.JsonAdapter;
import com.squareup.moshi.JsonReader;
import com.squareup.moshi.JsonWriter;
import java.io.IOException;
import java.time.Duration;

public final class IsoDurationJsonAdapter extends JsonAdapter<Duration> {
    @Override public synchronized Duration fromJson(JsonReader reader) throws IOException {
        String string = reader.nextString();
        return Duration.parse(string);
    }

    @Override public synchronized void toJson(JsonWriter writer, Duration value) throws IOException {
        String string = (value == null) ? null : value.toString();
        writer.value(string);
    }
}
`

Another useful one

JakeWharton:
You don't need to take `JsonReader` or `JsonWriter` there. Just accept `String` and return `String`.

JakeWharton:
Also drop `synchronized`. `java.time` APIs are thread-safe unless otherwise noted (these are safe).

kenyee:
The String methods of the JsonAdapter are final unfortunately :-(
Removed synchronized...thanks.  That was from an autoconversion to Kotlin by Android Studio :-)

`
  public final @Nullable T fromJson(String string) throws IOException 

  public final String toJson(@Nullable T value) 

`

NightlyNexus:
Jake meant with `@FromJson` and `@ToJson` methods, you can accept and return a `String` delegate.
like `@FromJson Duration fromJson(String string) {...}` `@ToJson String toJson(Duration duration) {...}`

less code than writing a full JsonAdapter manually.

msrd0:
:+1: I'd definitely appreciate if this made it into the library

raman-nbg:
Still not made it to the library? :(

spyro2000:
[...]

PR #1595, opened 2022-12-21
[FEATURE] New factory API to control the type `labelKey` serialization.

hossain-khan:
## Summary
New API `autoSerializeLabel(boolean)` added to allow the factory can be configured with boolean flag to disable automatic serialization of the type label, and it's respective value. Here is how the configuration may look like in the factory builder.

```java
PolymorphicJsonAdapterFactory.of(Message.class, "type")
    .withSubtype(Success.class, "success")
    .withSubtype(MessageWithType.class, "customType")
    .autoSerializeLabel(false)) // ← New API!
```

This is done for use cases when all known subtypes has serializable type property in the class definition, then it generates double `type` properties when serialized to JSON.

### Background of the change
There are two key reason for requesting this feature (new API).
1. In projects, sub-type classes can have type property which is serializable. Adding `@Json(ignore = true)` is not an option. That results in double type value in the serialized JSON object.
2. There are some use cases where serializing object via parent type is not always applicable (eg `Message` in this example). When sub-type has type class property, it can be serialized directly without using polymorphic adapter (eg. `MessageWithType` in this example). Having this API would allow to have `type` property in class and allow it to play nicely with polymorphic adapter when deserializing as well.

> ✅ Signed "Square Inc. Individual Contributor License Agreement" for my GitHub username.


hossain-khan:
🎄 Happy Holidays, and happy to answer any questions related to this feature request. 🙏🏽

hossain-khan:
I don't think the `CI / Test Mode KSP` [failure](https://github.com/square/moshi/actions/runs/3752087291/jobs/6373866320) is related to my code changes

```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':moshi-kotlin-codegen'.
Caused by: org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException: Could not resolve all files for configuration ':moshi-kotlin-codegen:classpath'.
Caused by: org.gradle.api.resources.ResourceException: Could not get resource 'https://plugins.gradle.org/m2/gradle/plugin/com/github/jengelman/gradle/plugins/shadow/7.0.0/shadow-7.0.0.jar'.
Caused by: org.gradle.internal.resource.transport.http.HttpRequestException: Could not GET 'https://plugins-artifacts.gradle.org/gradle.plugin.com.github.jengelman.gradle.plugins/shadow/7.0.0/8e7b12a0a2c2d170884a860679a019bab60b63bc99ebfa0db0897d249d86f345/shadow-7.0.0.jar'.
Caused by: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
```

Could some please re-start the CI task when possible! Thanks

waqasmoazzam:
Hey folks, I am having the same issue at our organization where it is showing duplicate keys in the serialized JSON object. I cannot change the domain models. Having this config would help avoid the duplication issue.

@JakeWharton is this something that can be considered for future release?

hossain-khan:
I think a good way to summarize this PR requests is mentioned here https://github.com/square/moshi/issues/181#issuecomment-636039833 in the video 🎥 

> You can watch Jesse and I talk about this phenomenon here: https://youtu.be/PCxz2LEmuL4?t=1643. We don't build kitchen sink libraries.

Not sure how to solve it in extensible way.

PR #1597, opened 2023-01-03
updated read me documentation

prasannaec40:
Updated comment in the READ ME file for moshi.

Below line of code can be used if we opt in ExperimentalStdlibApi in Kotlin. Adding specific details to the READ ME file.

`val adapter = moshi.adapter<List<Card>>()`


ZacSweers (commented):
This will eventually not be experimental and the IDE will indicate this already when they try to use it.

PR #1601, opened 2023-01-22
[CI] Use built-in caching from `setup-java` action

valfirst:
https://github.com/actions/setup-java#caching-packages-dependencies:
> `setup-java` action has a built-in functionality for caching and restoring dependencies. It uses actions/cache under hood for caching dependencies but requires less configuration settings.


ZacSweers:
If we're going to change this, we should use the gradle build action and its caching instead.

JakeWharton:
We can land that separately, if it even matters. Builds on this project are already fast and few-and-far-between.

I tried to merge on mobile a few hours ago but I guess there's a new token scope for modifying actions or something?

PR #1603, opened 2023-01-28
Upgrade to Gradle 7.6

sconvent:
This will upgrade the Gradle version used in the Gradle Wrapper to 7.6

Done using `./gradlew wrapper --gradle-version 7.6 --distribution-type bin`


JakeWharton:
Thanks for sending the PR. Unfortunately this version of Gradle has a bug with JDK toolchains that prevents it from working. See #1608. We have to wait for Gradle 7.6.1 or 8.0 for the fix. Renovate will automatically update to that version when it's available, so I'm going to close your PR.

PR #1623, opened 2023-01-31
fix PolymorphicJsonAdapterFactory with FallbackJsonAdapter deserialize empty object error.

coofee:
well，when PolymorphicJsonAdapterFactory deserialize json array contains empty object like below,  the origin logic will throw `JsonDataException("Missing label for $labelKey")` instead of using fallbackJsonAdapter，this PR will use fallbackJsonAdapter deserialize empty json to object other than throw  `JsonDataException`。

```json
"[{\"type\":\"success\",\"value\":\"this is success message\"},{}]"
```

### DIFF
--- a/moshi-adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
+++ b/moshi-adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
@@ -16,6 +16,8 @@
 package com.squareup.moshi.adapters;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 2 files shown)


ZacSweers:
Can you please write a description explaining the PR changes?

coofee:
> its my fault，already update now，please review again。

moshi-adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java
```diff
@@ -16,19 +16,27 @@
 package com.squareup.moshi.adapters;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
```
ZacSweers, line 20: Please use Truth assertions

ZacSweers (changes requested):

Issue #1512, opened 2022-01-23
Allow PolymorphicJsonAdapterFactory's fallback adapter to handle case when expected `labelKey` is missing from JSON object.


hossain-khan:
> Start by telling us what problem you’re trying to solve. Often a solution already exists!

## Summary
I was trying to use the [PolymorphicJsonAdapterFactory](https://github.com/square/moshi/blob/master/moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt), and the usage is very straightforward, as seen in [test](https://github.com/square/moshi/blob/master/moshi-adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java):

```java
PolymorphicJsonAdapterFactory.of(Message.class, "type" /* labelKey */)
            .withSubtype(Success.class, "success")
            .withSubtype(Error.class, "error")
```

However, in non-ideal world, the backend server may not always send the expected `labelKey`. In that case, instead of failing, I wanted to leverage the `withFallbackJsonAdapter` [API](https://github.com/square/moshi/blob/master/moshi-adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.kt#L135) to return a different subtype. 

However, when the `labelKey` is missing, it throws `JsonDataException: Missing label for type`, which would result in failing the API request.

Provided a unit-test below that explains this further. I was wondering if allowing the provided `withFallbackJsonAdapter` to handle such case would be an acceptable use case?

### Test that reflects the ask
I have added a test in [PolymorphicJsonAdapterFactoryTest](https://github.com/square/moshi/blob/master/moshi-adapters/src/test/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactoryTest.java) to showcase what is expected 

```java
/**
 * When {@link PolymorphicJsonAdapterFactory#withFallbackJsonAdapter(JsonAdapter)} is used,
 * it should allow the adapter to handle missing `labelKey` from the source JSON object,
 * so that custom handling can be done using the provided fallback JSON adapter.
 *
 * In this unit-test, when type is missing, a special implementation {@link UnknownMessage} is provided,
 * so that the application can render something else based on business-logic/requirement.
 *
 * <p>---</p>
 *
 * Currently, for following JSON input (that has "type" labelKey missing in JSON object):
 * <code>{"faulty_state":"Unknown Message Type"}</code>
 *
[...]

PR #1650, opened 2023-03-26
Allow PolymorphicJsonAdapterFactory's fallback adapter to handle case…

tomridder:
[Allow PolymorphicJsonAdapterFactory's fallback adapter to handle case when label key doesn't appear in json string]
even if the lable key doesn't appear in json string,the better solution is still execute `PolymorphicJsonAdapter#fromJson`
#1512


tomridder:
hi.is there any problem @ZacSweers @JakeWharton

ZacSweers:
This is not what we intended this API for. It's for unrecognized keys, not completely missing keys. Also it's quite rude to @ mention maintainers the way you did there, please refrain from this in the future.

Issue #1321, opened 2021-03-20
PolymorphicJsonAdapterFactory only supports json string labels


maciej-kaznowski:
Hi, I have the following (simplified) json:

```
{
  "data": [
    {
      "was_comment": false
    },
    {
      "was_comment": true
    }
  ]
}
```

I wanted to de serialize these objects depending on the value of `was_comment`:

```
PolymorphicJsonAdapterFactory.of(Message::class.java, "was_comment")
                         .withSubtype(RedditCommentMessage::class.java, "true")
                         .withSubtype(RedditPrivateMessage::class.java, "false")
```

However the `com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory.PolymorphicJsonAdapter` only works for labels which are strings:
https://github.com/square/moshi/blob/797140c5cf0e376387b99fcaa8cc00988d1ffbb5/adapters/src/main/java/com/squareup/moshi/adapters/PolymorphicJsonAdapterFactory.java#L277

so it throws an exception:

```
com.squareup.moshi.JsonDataException: Expected a string but was BOOLEAN at path $.data.children[0].data.was_comment

	at com.squareup.moshi.JsonUtf8Reader.nextString(JsonUtf8Reader.java:674)
	at com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory$PolymorphicJsonAdapter.labelIndex(PolymorphicJsonAdapterFactory.java:285)
	at com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory$PolymorphicJsonAdapter.fromJson(PolymorphicJsonAdapterFactory.java:257)
```

So I believe it may be beneficial to add support for the other types:
- boolean
- double, int, long
- null

ZacSweers:
What would the API for that look like?

maciej-kaznowski:
Sorry for the late response.

I think the simplest approach would simply be to add methods for the other types:

Current:
`com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory#withSubtype(java.lang.Class<? extends T>, java.lang.String)`


New methods:
```
com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory#withSubtype(java.lang.Class<? extends T>, boolean)
com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory#withSubtype(java.lang.Class<? extends T>, double)
[...]

PR #1652, opened 2023-03-31
add the function of support Boolean Long Int Double labelType in PolymorphicJsonAdapterFactory with test case 

tomridder:
add the function of support Boolean Long Int Double labelType in PolymorphicJsonAdapterFactory with test case 
#1321 

all the testcase in moshi adapter module has been passed :)


ZacSweers:
This isn't something we want to do, sorry.

tomridder:
> This isn't something we want to do, sorry.

is there any thing that has been achieved in moshi and need developers to deal with :)

tomridder:
> This isn't something we want to do, sorry.

if issue 1321 is not something you guys want to do ,why would you guys gives a enhancement lable on issue 1321 :)

tomridder:
> This isn't something we want to do, sorry.

is it because of the scheme or is it because the issue doesn't matter:)

PR #1661, opened 2023-04-19
remove task comment

jelgun:
This PR removes a task comment (i.e. a comment referring to a work that could/should be done in the future or was already done). Instead, an issue was created.


JakeWharton:
We generally prefer to keep the source of truth for these in code. If you want to track them in issues I would recommend something like https://github.com/marketplace/actions/track-todo-action which I've used on a few projects.

PR #1665, opened 2023-04-23
DefaultConstructorMarker is now public

vvb2060:
https://github.com/square/moshi/pull/1545#issuecomment-1518943013


ZacSweers:
Same issues as the previous PR, please take no for an answer on this.

PR #1731, opened 2023-08-30
Insert `!!` to make some types in generated code non-nullable.

cpovirk:
Hi, Square folks.

The background here is that we're working to put nullness annotations on
more "JDK classes," such as `java.lang.Class`, at least in Google's
build system. That means that calls to methods like
`clazz.getMethod(...)` and `clazz.getDeclaredConstructor` must pass
non-nullable `Class` instances. If they don't, we see errors:

```
error: type mismatch: inferred type is Class<Int>? but Class<*> was expected
          String::class.java, String::class.java, Int::class.javaPrimitiveType, String::class.java,
                                                  ^
```

```
error: type mismatch: inferred type is Class<Int>? but Class<*> was expected
          String::class.java, FooResponse::class.java, Int::class.javaPrimitiveType,
                                                       ^
```

```
error: type mismatch: inferred type is Class<CapturedType(*)>? but Class<*> was expected
          Util.DEFAULT_CONSTRUCTOR_MARKER).also { this.constructorRef = it }
          ^
```

Expressions like `Int::class.javaPrimitiveType` are always non-nullable,
so we can use `!!` (or `checkNotNull` or whatever you might prefer)
there without risk.

(Note that it would be possible to avoiding using `!!` for
`INT_TYPE_BLOCK`: We could do so by passing `Integer.TYPE`. The more
general case of `TypeName.asTypeBlock` could theoretically be handled
similarly but would likely require maintaining a mapping from Kotlin
type name to Java primitive wrapper class.)

As for `DEFAULT_CONSTRUCTOR_MARKER`, I haven't looked at the generated
code enough to prove that it's non-nullable here. But if I'm reading the
bytecode right, the property value is being passed to
`getDeclaredConstructor`, so I'd expect NPE there if it were null.

Let me know what you think. I could check with our internal Moshi owners
about maintaining this as a Google-local patch if that's better, but I'm
hoping that this is a fairly harmless change, and our research so far
suggests that we _probably_ won't be back soon to ask about other
changes for this nullness work.

### DIFF
--- a/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
+++ b/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/AdapterGenerator.kt
@@ -58,9 +58,9 @@ public class AdapterGenerator(
 ) {
 
   private companion object {
-    private val INT_TYPE_BLOCK = CodeBlock.of("%T::class.javaPrimitiveType", INT)
+    private val INT_TYPE_BLOCK = CodeBlock.of("%T::class.javaPrimitiveType!!", INT)
     private val DEFAULT_CONSTRUCTOR_MARKER_TYPE_BLOCK = CodeBlock.of(
-      "%M",
+      "%M!!",
       MemberName(MOSHI_UTIL_PACKAGE, "DEFAULT_CONSTRUCTOR_MARKER"),
     )
     private val CN_MOSHI = Moshi::class.asClassName()
--- a/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
+++ b/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt
@@ -118,7 +118,7 @@ internal fun TypeName.asTypeBlock(): CodeBlock {
             // Remove nullable but keep the java object type
             CodeBlock.of("%T::class.javaObjectType", copy(nullable = false))
           } else {
-            CodeBlock.of("%T::class.javaPrimitiveType", this)
+            CodeBlock.of("%T::class.javaPrimitiveType!!", this)
           }
         }
         UNIT, Void::class.asTypeName(), NOTHING -> throw IllegalStateException("Parameter with void, Unit, or Nothing type is illegal")


cpovirk:
> As for `DEFAULT_CONSTRUCTOR_MARKER`, I haven't looked at the generated code enough to prove that it's non-nullable here. But if I'm reading the bytecode right, the property value is being passed to `getDeclaredConstructor`, so I'd expect NPE there if it were null.

Err, I think I was looking at the wrong thing there, sorry. But I do see a call to `checkNotNullExpressionValue` in the bytecode after the `putfield` for `constructorRef`, so there is probably still _some_ runtime null check that is already happening.

JakeWharton:
Since these will never actually be null in practice, I would prefer we use `checkNotNull`. 

Using `!!`, while short, results in the expression being tested being put into the constant pool for use as the exception message. `checkNotNull`, or something like `?: throw ISE()`, will not use a message.

cpovirk:
Thanks. I'm definitely up for making the switch, but I remember looking into the bytecode for `!!` a little while back, and I don't think it results in a constant-pool entry. A case that _does_ result in a constant-pool entry is when you have a platform-typed value that you, e.g., assign to a non-nullable property.

Here's some bytecode for `!!` vs. `checkNotNull`. (The `checkNotNull` bytecode is actually longer, but I wouldn't assume that that remains the case after optimization.)

```
$ cat NullChecks.kt
fun get(): String? = null

fun getAllBangBang(): List<String> =
  listOf(get()!!);

fun getAllCheckNotNull(): List<String> =
  listOf(checkNotNull(get()));

$ $HOME/kotlin-compiler-1.9.10/kotlinc/bin/kotlinc NullChecks.kt && javap -v NullChecksKt
...
  public static final java.util.List<java.lang.String> getAllBangBang();
...
         0: invokestatic  #13                 // Method get:()Ljava/lang/String;
         3: dup
         4: invokestatic  #19                 // Method kotlin/jvm/internal/Intrinsics.checkNotNull:(Ljava/lang/Object;)V
         7: invokestatic  #25                 // Method kotlin/collections/CollectionsKt.listOf:(Ljava/lang/Object;)Ljava/util/List;
        10: areturn
...
  public static final java.util.List<java.lang.String> getAllCheckNotNull();
...
         0: invokestatic  #13                 // Method get:()Ljava/lang/String;
         3: dup
         4: ifnonnull     23
         7: pop
         8: ldc           #28                 // String Required value was null.
        10: astore_0
        11: new           #30                 // class java/lang/IllegalStateException
        14: dup
        15: aload_0
        16: invokevirtual #33                 // Method java/lang/Object.toString:()Ljava/lang/String;
        19: invokespecial #37                 // Method java/lang/IllegalStateException."<init>":(Ljava/lang/String;)V
        22: athrow
        23: invokestatic  #25                 // Method kotlin/collections/CollectionsKt.listOf:(Ljava/lang/Object;)Ljava/util/List;
        26: areturn
...
```

Still, let me know if you'd prefer `checkNotNull` anyway (including if I got something wrong here).

JakeWharton:
Oh, cool. No needless string, no need to change!

PR #1848, opened 2024-06-03
Remove KAPT support

ZacSweers:
With this, only KSP is supported for code gen now.


JakeWharton (approved):

ZacSweers:
@JakeWharton can you remove the requirement on the KAPT build status? Doesn't look like I have settings perms to change branch requirements

JakeWharton:
Now removed

PR #1856, opened 2024-06-19
Don't replace nullable absent properties with null

ZacSweers:
This was always supposed to be a feature of moshi but not actually enforced in kotlin reflect or code gen. This fixes that in both with a regression test. Now nullable properties either need to explicitly declare a default value or explicitly have a null present in the JSON.

This uses a couple clever tricks with kotlin contracts to avoid the extra runtime intrinsic checks.

It's worth mentioning that while this was always how this was advertised to work, this is likely going to be a fairly invasive change to people that were implicitly relying on the old behavior.


swankjesse:
I don’t think this behavior is consistent with `.serializeNulls(false)` . . . JSON is generally pretty lenient between absent and null, and I think making Moshi strict is going to be more pain than it’s worth?

ZacSweers:
Agreed 👍

PR #1858, opened 2024-06-22
Generate bytecode directly to avoid reflection to invoke defaults constructors

ZacSweers:


ZacSweers:
Remains broken in KSP https://github.com/google/ksp/issues/95

Theschorpioen (approved):

PR #1875, opened 2024-07-31
Remove ksp/shadedUtil.kt

ting-yuan:
and use KSP's implementation.


ZacSweers:
We're reluctant to ship unshaded experimental KSP APIs since they could break again. Could these be promoted to stable instead?

ting-yuan:
A large part of those are actually meant for JVM and we haven't have time to (find the best way to) make them JVM only. The  earliest date I can think of to work on this is near the end of this year, though. Short term solutions can be updating shadedUtil.kt or making it a wrapper where the experimental annotations are contained.

JakeWharton:
The problem is the lack of ABI compatibility. We can't use it at all. But updating sounds like the right short-term solution.

Jean-Daniel:
Is there any plan to merge this change ? @JsonQualifier are still unusable with KSP 2 without it.

Brock-Gibson:
Any update on this, would love to be able to KSP2.

Tolriq:
Just fork and merge this, it works nicely.

PR #1883, opened 2024-08-18
Add a simple AST model and Adapter

desiderantes:
This adapter provides a similar functionality to Gson's `JsonElement`, but it's just another option and not a central part of the API.


JakeWharton:
I would recommend shipping this as its own library if you want to offer this. As you've seen in our README, its omission is intentional.

PR #1892, opened 2024-09-24
Fix #1891. Remove invalid and operation inside Byte Adapter

### LINKED ISSUE #1891: Byte Adapters encode wrong Value
The Standard Adapter for byte values encodes a wrong value for negative Byte values
[(https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.kt:L85)](https://github.com/square/moshi/blob/6f3031b8319f98b2dd87c3d5adef2d6e88a64f70/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.kt#L85)

The same Issue applies to the Java implementation.

Heres an example
https://gist.github.com/Andy3189/2110ce41e3159022cb3bdf646b0fa1e7

Andy3189:
The and operation creates invalid results for negative Byte values as mentioned in #1891

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.kt
+++ b/moshi/src/main/java/com/squareup/moshi/StandardJsonAdapters.kt
@@ -82,7 +82,7 @@ internal object StandardJsonAdapters : JsonAdapter.Factory {
     }
 
     override fun toJson(writer: JsonWriter, value: Byte?) {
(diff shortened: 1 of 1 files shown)


JakeWharton:
Test?

Andy3189:
There has been a misunderstanding of the number representation in json

PR #1913, opened 2024-12-05
Remove redundant preconditions from generated ProGuard rules

### LINKED ISSUE #1911: Moshi generated Proguard rules avoid -if rules for -keepnames rules
Moshi appears to be generating Proguard rules on the form `-if class T -keepnames class T`. The `-keepnames` directive is an alias for `-keep,allowshrinking`, which means that the `-if class T` precondition is redundant. The generated rules should ideally be changed to `-keepnames class T`.

Conditional `-if` rules are generally more expensive to evaluate than non `-if` rules, which leads to higher build speeds. In an app that had thousands of Moshi generated `-if class T -keepnames class T` rules, removing the `-if class T` part from these rules improved build speed by ~8 %.

PhilGlass:
Fixes #1911.

@christofferqa (who works on R8 at Google) was looking into Monzo's slow R8 builds. One thing he found was removing redundant `-if` rules generated by Moshi reduced the time we spend in R8 by about 8% (we have ~2.8k classes  annotated with `@JsonClass`).

### DIFF
--- a/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
+++ b/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/ProguardRules.kt
@@ -26,7 +26,8 @@ import com.squareup.kotlinpoet.ClassName
  *   constructor as well as the DefaultConstructorMarker type Kotlin adds to it.
  *
  * Each rule is intended to be as specific and targeted as possible to reduce footprint, and each is
- * conditioned on usage of the original target type.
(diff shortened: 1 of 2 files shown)


JakeWharton:
Thanks. We should probably backport this to 1.x, as Moshi is not in a releasable state as-is.

JakeWharton:
1.15.2

PhilGlass:
> Thanks. We should probably backport this to 1.x, as Moshi is not in a releasable state as-is.

Yeah I wasn't really sure what to PR against - I tested locally with a patch on top of `moshi_1.x`.

PR #1956, opened 2025-05-04
Replace deprecated `indentWithSpaces` with `leadingTabsToSpaces`

galiren:
Replaces the deprecated `indentWithSpace` with `leadingTabsToSpace` as discussed in [spotless#794](https://github.com/diffplug/spotless/issues/794).

Since `leadingTabsToSpace` has been available since spotless Gradle plugin `7.0.0`, and moshi currently uses `7.0.3`, it's safe to adopt the new API.


JakeWharton:
Thanks!

Issue #1995, opened 2025-09-17
2.0: Adapter generation fails with unresolved reference on annotation with enum default.


lassem:
This applies to 2.0.0-SNAPSHOT

Changing the `@Uppercase` annotation in the `GeneratedAdapterTest.kt:1199` to:
```kotlin
  @JsonQualifier
  annotation class Uppercase(val inFrench: Boolean, val onSundays: Boolean = false, val temperature: Temperature = Temperature.COLD) {
    enum class Temperature {
      WARM, COLD
    }
  }
```
will reproduce this issue.

`Unresolved reference 'Temperature'.`

I was trying to get rid of the KSP2 issue.

ZacSweers:
Merged into the `z/ksp2` branch

PR #2000, opened 2025-09-23
Work-around for missing fqn on default args to annotations

lassem:
Fix for https://github.com/square/moshi/issues/1995

If you have an annotation applied to your  POJOs and that annotation has a default parameter in its list, the new adapter generator would fail to produce a fully qualified name for the type.

This is my best-attempt, I'm not familiar with all the intricacies of resolving the type information here.
I'd be glad to improve on this if someone points me in the right direction.


JakeWharton:
I make @ZacSweers look at all the KSP stuff since I've never used it...

I'll let CI run, though.

ZacSweers:
@lassem mind allowing edits from maintainers in future PRs? The change I requested is easy enough for me to just do for ya but I can't!

PR #2002, opened 2025-09-30
Implement automatic generation of Moshi.AdapterRegistry to avoid looking up code gen entities by reflection class names

fankes:
I found that the adapter generated by code gen in Moshi is loaded by obtaining the `rawType` through `Class.forName` reflection. This makes it impossible to obfuscate class names after enabling R8 in an Android project. That's why this feature was created. It mainly serves Android R8 and allows class names to be obfuscated at will. Users can register the generated `Moshi.AdapterRegistry` with `Moshi.Builder` to create a mapping relationship. I believe this will have much higher performance than direct reflection.

To prevent problems, this feature is not enabled by default.

Configuring ksp:

```kotlin
ksp {
    // Enable obfuscated class name
    arg("moshi.proguardRulesDontKeepClassNames", "true")
    //  Enable generate default Moshi.AdapterRegistry
    arg("moshi.generateDefaultAdapterRegistry", "true")
}
```

All classes with `@JsonClass(generateAdapter = true)` added in your project will be scanned and `DefaultMoshiAdapterRegistry` will be generated.

For example, this is our entity class:

```kotlin
data class Foo(
    var someString: String = "123",
    var bar: Bar? = null
) {

    data class Bar(
        val name: String,
        val jobTitles: List<String>
    )
}
```

To prevent conflicts in generated package names, I took the longest package name, made a max 12-digit hash, and added `r` in front of it.

Like this: `com.squareup.moshi.generated.r2b362199.DefaultMoshiAdapterRegistry`

The generated class is almost like this:

```kotlin
// This file is auto generated by Moshi Kotlin CodeGen.
// **DO NOT EDIT THIS FILE MANUALLY**
package com.squareup.moshi.generated.r2b362199

import com.myproject.Foo
import com.myproject.FooJsonAdapter
import com.myproject.Foo_BarJsonAdapter
import com.squareup.moshi.JsonAdapter
import com.squareup.moshi.Moshi
import java.lang.Class
import java.lang.reflect.Type
import kotlin.collections.Map

public class DefaultMoshiAdapterRegistry : Moshi.AdapterRegistry {
  override val adapters: Map<Type, Class<out JsonAdapter<*>>>
    get() = mapOf(
      Foo::class.java to FooJsonAdapter::class.java,
      Foo.Bar::class.java to Foo_BarJsonAdapter::class.java
    )
}
```

To use the generated `DefaultMoshiAdapterRegistry`, you can register it with `Moshi.Builder` like this:

```kotlin
val moshi = Moshi.Builder()
    .setRegistry(DefaultMoshiAdapterRegistry())
    .build()
```

This will completely obfuscate the class name of the class annotated with `@JsonClass(generateAdapter = true)` and will not expose sensitive information.

Maybe this implementation still needs to be improved. I can put it here as a new pending feature for the time being. Let's see if you have any better suggestions or areas that need improvement. 🙏🏻


ZacSweers:
We don't want this for a number of reasons
- It performs eager, slow classloading
- It requires making the processor aggregating
- It adds a whole unnecessary dedicated `Moshi.AdapterRegistry` API that could be simply satisfied by a custom JsonAdapter.Factory.
- It can be implemented independently of moshi's first party code gen, which is what you should do

In the future I recommend starting an issue to ask for feedback before implementing

fankes:
You are right, there are indeed performance issues here, and it is reasonable for me to implement a solution myself and use the existing `JsonAdapter.Factory`. Thank you for your work!

PR #2004, opened 2025-10-01
Update moshi-kotlin to use kotlin-metadata instead

ZacSweers:
Mostly a straight port of https://github.com/ZacSweers/MoshiX/tree/main/moshi-metadata-reflect (benchmarks linked there too)

### DIFF
--- a/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
+++ b/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
@@ -268,6 +268,7 @@ class DualKotlinTest {
   @JsonClass(generateAdapter = true)
   data class InlineConsumer(val inline: ValueClass)
 
+  // TODO this produces {"inline":23} now
   @Test fun inlineClassConsumer() {
     val adapter = moshi.adapter<InlineConsumer>()
 
--- a/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
+++ b/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
@@ -243,6 +243,8 @@ class DualKotlinTest {
     assertThat(decoded.a).isEqualTo(null)
   }
 
+  // TODO CR ValueClass enables generateAdapter but it's not present in reflect-only.
+  //  We previously allowed reflect to do this
   @Test fun valueClass() {
     val adapter = moshi.adapter<ValueClass>()
 
--- a/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
+++ b/moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
@@ -288,9 +288,8 @@ class KotlinJsonAdapterTest {
       fail()
     } catch (expected: IllegalArgumentException) {
       assertThat(expected).hasMessageThat().isEqualTo(
-        "No default value for transient constructor parameter #0 " +
-          "a of fun `<init>`(kotlin.Int): " +
-          "com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest.RequiredTransientConstructorParameter",
+        "No default value for transient/ignored constructor parameter 'a' on type " +
+          "'com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest.RequiredTransientConstructorParameter'",
       )
     }
   }
@@ -304,9 +303,8 @@ class KotlinJsonAdapterTest {
       fail()
     } catch (expected: IllegalArgumentException) {
       assertThat(expected).hasMessageThat().isEqualTo(
-        "No default value for ignored constructor parameter #0 " +
-          "a of fun `<init>`(kotlin.Int): " +
-          "com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest.RequiredIgnoredConstructorParameter",
+        "No default value for transient/ignored constructor parameter 'a' on type " +
+          "'com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterTest.RequiredIgnoredConstructorParameter'",
       )
     }
   }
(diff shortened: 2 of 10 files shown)


moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
```diff
@@ -243,6 +243,8 @@ class DualKotlinTest {
     assertThat(decoded.a).isEqualTo(null)
   }
 
+  // TODO CR ValueClass enables generateAdapter but it's not present in reflect-only.
```
ZacSweers, line 246: I'm not actually sure why this didn't fail before tbh. We previously allowed reflect to read these if they weren't found so people could have shared source that selectively enabled code gen. Open to either approach, just need to make a call here

moshi-kotlin/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterFactory.kt
```diff
@@ -213,111 +225,102 @@ public class KotlinJsonAdapterFactory : JsonAdapter.Factory {
...
       "Cannot reflectively serialize sealed class ${rawType.name}. Please register an adapter."
     }
 
-    val constructor = rawTypeKotlin.primaryConstructor ?: return null
-    val parametersByName = constructor.parameters.associateBy { it.name }
-    constructor.isAccessible = true
+    val ktConstructor = KtConstructor.primary(rawType, kmClass) ?: return null
 
-    val bindingsByName = LinkedHashMap<String, KotlinJsonAdapter.Binding<Any, Any?>>()
+    // TODO CR this doesn't cover platform types
```
ZacSweers, line 250: I can add this, I just want to check if we really want to support this as code gen does not support this (explicitly) and there is a test against it. Commented on an example test of this in this PR too

moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterTest.kt
```diff
@@ -382,6 +380,7 @@ class KotlinJsonAdapterTest {
 
   internal class ExtendsPlatformClassWithPrivateField(var a: Int) : SimpleTimeZone(0, "C")
 
+  // TODO in code review - this now only produces {"a":3}. Do we want this?
```
ZacSweers, line 383: This is an example of a platform type getting its fields encoded before

moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
```diff
@@ -268,6 +270,7 @@ class DualKotlinTest {
   @JsonClass(generateAdapter = true)
   data class InlineConsumer(val inline: ValueClass)
 
+  // TODO this produces {"inline":23} now
```
ZacSweers, line 273: The new output matches what kotlinx-serialization does (inlines the value), but code gen produces the boxed type. Moshi's so far had no opinion in this space. Can look into it more but open to suggestions on what we should do

moshi-kotlin/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapterFactory.kt
```diff
@@ -141,51 +140,64 @@ internal class KotlinJsonAdapter<T>(
...
+
+      error("Could not get JVM field or invoke JVM getter for property '$name'")
+    }
 
     fun set(result: K, value: P) {
       if (value !== ABSENT_VALUE) {
-        (property as KMutableProperty1<K, P>).set(result, value)
+        property.jvmSetter?.let { setter ->
+          setter.invoke(result, value)
+          return
```
swankjesse: I don’t love `return` inside `let` because I cannot remember how much it returns

moshi-kotlin/src/main/java/com/squareup/moshi/kotlin/reflect/KtTypes.kt
```diff
@@ -0,0 +1,243 @@
...
+    @Suppress("UNCHECKED_CAST")
+    return jvm.newInstance(*arguments.toTypedArray()) as R
+  }
+
+  companion object {
+    fun primary(rawType: Class<*>, kmClass: KmClass): KtConstructor? {
+      val kmConstructor = kmClass.constructors.find { !it.isSecondary } ?: return null
+      val kmConstructorSignature = kmConstructor.signature?.toString() ?: return null
+      val constructorsBySignature =
+        rawType.declaredConstructors.associateBy { it.jvmMethodSignature }
```
swankjesse: We might be able to avoid building a whole map to just do a single lookup in it

swankjesse (approved):
This is super cool

I have a bunch of nitpicks but it’s all exactly that - nitpicks. I’d love to land this and then do a quick pass to see if we can further tune it for performance.

What’s really nice about this whole approach is it’s lower-level and there’s more optimization opportunities!

JakeWharton (commented):
Just submitting stuff that I had drafted when the PR was sent

moshi-kotlin/src/main/java/com/squareup/moshi/kotlin/reflect/JvmDescriptors.kt
```diff
@@ -0,0 +1,154 @@
...
+        PRIMITIVE_CLASS_TO_SYMBOL[this]?.toString()
+          ?: throw RuntimeException("Unrecognized primitive $this")
+      isArray -> "[${componentType.descriptor}"
+      else -> "L$name;".replace('.', '/')
+    }
+  }
+
+private class State(var index: Int)
+
+private fun readType(state: State, desc: String): Class<*> {
```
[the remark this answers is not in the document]
ZacSweers: [6a34c01](https://github.com/square/moshi/pull/2004/commits/6a34c01f185901b4810d54a63bd05b66322eaf21)

moshi-kotlin-tests/src/test/kotlin/com/squareup/moshi/kotlin/DualKotlinTest.kt
squarejesse: I think the best behavior is that adding the `inline` modifier doesn’t change the serialized form.

Probably we want that. If we really wanted to we could also offer a `JsonAdapter.Factory` that gives you the unwrapped behavior on all inline classes for compatibility?

ZacSweers: Done

ZacSweers:
The value class work was a bit more weedy than I was expecting but happy with it. Required a bit of reverse-engineering how kotlin handles these in a couple places. Happy to work on any comments in a followup!

ZacSweers:
Another thing we'll need to do is replicate the behavior in code gen

Issue #1874, opened 2024-07-31
KSP2 crash when using @JsonQualifier


Jean-Daniel:
When trying to generate Adapter for classes using a @JsonQualifier, KSP fails with the error:

```

> Task :app:kspDebugKotlin FAILED
e: [ksp] Error preparing ApiUserItem: jdk.proxy5/jdk.proxy5.$Proxy154.value(Unknown Source)
com.squareup.moshi.kotlin.codegen.ksp.MoshiApiUtilKt.generator(MoshiApiUtil.kt:77)
com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessor.adapterGenerator(JsonClassSymbolProcessorProvider.kt:122)
com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessor.process(JsonClassSymbolProcessorProvider.kt:92)
com.google.devtools.ksp.impl.KotlinSymbolProcessing.execute(KotlinSymbolProcessing.kt:538)
com.google.devtools.ksp.impl.KSPLoader$Companion.loadAndRunKSP(KSPLoader.kt:36)
com.google.devtools.ksp.impl.KSPLoader.loadAndRunKSP(KSPLoader.kt)
…
```

The class being compiled is simply:

```kotlin
@Retention(AnnotationRetention.RUNTIME)
@JsonQualifier
annotation class ISO8061

@JsonClass(generateAdapter = true)
open class ApiUserItem(
  @field:Json(name = "slug") val slug: String,
  @field:Json(name = "date_added") @field:ISO8061 val date: Long,
)
```

You can use this reduced test case to reproduce the bug.

```
git clone https://github.com/Jean-Daniel/moshiksp2
cd moshiksp2
./gradlew :app:assembleDebug
> Task :app:kspDebugKotlin FAILED
…
```

ZacSweers:
Upstream KSP2 bug, please file with them

ting-yuan:
This is caused by a [breaking change](https://github.com/google/ksp/blob/main/docs/ksp2api.md#evaluation-of-enum-entries-in-annotation-arguments) in KSP2. @ZacSweers could [this copy](https://github.com/square/moshi/blob/d17b6736740b49dd167dc0ca32c1dcef5ebb48f3/moshi-kotlin-codegen/src/main/java/com/squareup/moshi/kotlin/codegen/ksp/shadedUtil.kt) be deleted and let moshi use the implementation from KSP?

[...]

PR #2008, opened 2025-10-01
KSP2 prep

ZacSweers:


ygnessin:
Hi @ZacSweers, any plan to see this change through and support KSP2? Based on recent comments in https://github.com/square/moshi/issues/1874 it seems like many projects are blocked by Moshi, and AGP is now printing loud warnings about it.

Apologies for the bump, and thank you again for maintaining this library ❤️

PR #2015, opened 2025-10-02
Update releasing

### LINKED ISSUE #2011: Dokka 2 multi-module
This doesn't seem to work. We should also be deploying it like other projects.

ZacSweers:
Ported from Molecule and modified for Dokka v2

Resolves #2011

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 * Refuse `j$.*` types from Android library desugaring as platform types.
 * In-development snapshots are now published to the Central Portal Snapshots repository at https://central.sonatype.com/repository/maven-snapshots/.
+* Fully supports encoding/decoding of value classes in both moshi-kotlin and code gen.
+  * Note that Moshi does not propagate inlining to JSON by default. For example: `@JvmInline value class Color(val raw: Int)` is serialized to `{"raw": 12345}`.
+* New `@JsonClass.inline` property to allow inlining single-property JSON classes during encoding/decoding.
+  * This is particularly useful for value classes.
+  * For example, a class `@JvmInline value class UserId(val id: Int)` with `inline = true` will serialize as just `123` rather than `{"id": 123}`.
+* `PolymorphicJsonAdapterFactory` now invokes the fallback adapter when the label key is missing entirely from the JSON, not just when the label value is unrecognized.
 
 ### Upgrading to Moshi 2.x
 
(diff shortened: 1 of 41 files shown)


ZacSweers:
Not super familiar with the actions used for releasing here but the newer dokka multimodule config outputs it to `docs/2.x`

releasing.md
```diff
@@ -1,61 +1,50 @@
...
+# Releasing
 
 ### Prerequisite: Sonatype (Maven Central) Account
 
 Create an account on the [Sonatype issues site][sonatype_issues]. Ask an existing publisher to open
 an issue requesting publishing permissions for `com.squareup` projects.
 
-
 Cutting a Release
 -----------------
```
JakeWharton, line 9: We don't allow user accounts anymore
```suggestion
```

releasing.md
```diff
@@ -1,61 +1,50 @@
...
 
-8. Visit [Sonatype Nexus][sonatype_nexus] to promote (close then release) the releases. Or drop it
-   if there is a problem!
+   This will trigger a GitHub Action workflow which will create a GitHub release and upload the
+   release artifacts to Maven Central.
 
- [github_actions]: https://github.com/square/moshi/actions
- [sonatype_issues]: https://issues.sonatype.org/
- [sonatype_nexus]: https://s01.oss.sonatype.org/
+[sonatype_issues]: https://issues.sonatype.org/
```
JakeWharton, line 50:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+      - run: ./gradlew publish dokkaGenerate
+        env:
+          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.SONATYPE_CENTRAL_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.SONATYPE_CENTRAL_PASSWORD }}
+          ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.GPG_SECRET_KEY }}
+          ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.GPG_SECRET_PASSPHRASE }}
+
+      - name: Extract release notes
+        id: release_notes
+        uses: ffurrer2/extract-release-notes@v2
```
JakeWharton, line 28: We need to switch to keep-a-changelog format for this to work.

JakeWharton (approved):

releasing.md
ZacSweers: alas
CleanShot 2025-10-02 at 00 26 06@2x

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+          body: ${{ steps.release_notes.outputs.release_notes }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Deploy docs to website
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: site
+          FOLDER: docs/2.x/
```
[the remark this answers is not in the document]
ZacSweers: no the current dokka config outputs it there (copied from mkdocs though docs/ is excluded here so seems to be leftover as a target), but I can change it since we don't have an mkdocs site

.github/workflows/release.yml
JakeWharton: Yes. We use it mostly everywhere else.

JakeWharton:
Anything left here?

JakeWharton:
The only thing needed for keep-a-changelog is the version header. You can put whatever you want beneath it.

Example:
- https://github.com/cashapp/redwood/blob/0.19.0/CHANGELOG.md?plain=1#L3
- https://github.com/cashapp/redwood/releases/tag/0.19.0

ZacSweers:
Took a pass at the format

PR #2053, opened 2025-11-25
Propagate nullability to JsonAdapter

ZacSweers:
See changelog for primary bits

Some implementation details
- generated adapters still work via nullable types as it lets us give better error messages around which fields were unexpectedly null
- Was able to remove a bunch of `knownNotNull()` calls 🚀 
- PolymorphicJsonAdapter is asymmetric, it can decode nulls but doesn't allow encoding nullable instances. Matched existing behavior for now

### DIFF
--- a/examples/src/main/java/com/squareup/moshi/recipes/JsonString.kt
+++ b/examples/src/main/java/com/squareup/moshi/recipes/JsonString.kt
@@ -46,7 +46,7 @@ class JsonStringJsonAdapterFactory : JsonAdapter.Factory {
       reader.nextSource().use(BufferedSource::readUtf8)
 
     override fun toJson(writer: JsonWriter, value: String) {
-      writer.valueSink().use { sink -> sink.writeUtf8(checkNotNull(value)) }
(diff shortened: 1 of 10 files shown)


moshi/src/main/java/com/squareup/moshi/JsonAdapter.kt
```diff
@@ -175,19 +176,34 @@ public abstract class JsonAdapter<T> {
...
    */
   @CheckReturnValue
-  public fun nonNull(): JsonAdapter<T> {
-    return when (this) {
-      is NonNullJsonAdapter<*> -> this
+  public fun nonNull(): JsonAdapter<T & Any> {
+    val finalizedAdapter = when (this) {
+      is NonNullJsonAdapter<*> -> {
+        @Suppress("UNCHECKED_CAST")
+        this as JsonAdapter<T?>
```
swankjesse: `this as JsonAdapter<T & Any>` ?

swankjesse: Actually I’m unsure why we need this cast?

PR #2098, opened 2026-04-08
demo: implement breaking change with BooleanArray serialization

BL19:
Note that the tests should pass while the implementation is very broken.


BL19:
Missclick, part of coursework.

PR #2114, opened 2026-06-04
Fix: peekNumber rejects some numerals like 2^64 + "0"

ogolberg:
Issue: the "leading zero" logic in peekNumber, which tries to reject octal representations, does not account for long overflows. This results in valid numeric literals like "184467440737095516160" (2^64 + "0") being rejected in STRICT parsing mode.

The easiest way to reproduce this is to serialize and deserialize a BigInteger with the value above.

This is a minimal fix which adds an overflow check to the leading zero condition.

This bug was inherited from Gson, see https://github.com/google/gson/pull/3034. It affects 1.x as well.

### DIFF
--- a/moshi/src/main/java/com/squareup/moshi/-JsonUtf8Reader.kt
+++ b/moshi/src/main/java/com/squareup/moshi/-JsonUtf8Reader.kt
@@ -468,7 +468,7 @@ internal class `-JsonUtf8Reader` : JsonReader {
             }
 
             NUMBER_CHAR_DIGIT -> {
-              if (value == 0L) {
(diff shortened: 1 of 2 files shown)


JakeWharton:
Was this discovered in the wild? The problem is that I'm not sure there will ever be another release of Moshi...

At this point I think I'm going to move the `master` branch to `future` and make a new `trunk` which is based on the 1.x series where we can continue to release bug fixes.

ogolberg:
Thanks for the info! Yes, I ran into this in the wild. I have a workaround and can fork Moshi if necessary.

cwbith-cloud (approved):

PR #2115, opened 2026-06-10
Add google play sdk verification properties files

ZacSweers:
This is for https://developer.android.com/distribute/sdk-index, which Google's recently pinged me about again to set up.


JakeWharton (changes requested):
Just marking down what we spoke about separately, which is that we want to push back against this and likely aren't going to ship this here or in any artifact we control.

cwbith-cloud (approved):

cwbith-cloud (approved):

